PR #38, opened 2019-07-23
Refactor fetch logs to use real API, Added error handling

laviro:


laviro:
@ShimShtein could you have a look on the server side? thanks

Gemfile.lock
```diff
@@ -2,11 +2,13 @@ PATH
   remote: .
```
ShimShtein: Please remove Gemfile.lock from the pr

webpack/ForemanYupana/Components/ReportGenerate/ReportGenerate.js
```diff
@@ -54,9 +60,13 @@ ReportGenerate.propTypes = {
   processID: PropTypes.number,
   stopProcess: PropTypes.func,
   loading: PropTypes.bool,
-  logs: PropTypes.arrayOf(PropTypes.string),
+  logs: PropTypes.oneOfType([
+    PropTypes.arrayOf(PropTypes.string),
```
ShimShtein: When `logs` would receive multiple strings? It will get a single string from the server.

ShimShtein:
Please rebse the PR before merge

webpack/ForemanYupana/Components/ReportGenerate/ReportGenerate.js
laviro: I wonder whether we should get **all** of the output history as an array each polling,
or we should get **only the last one** and push it in to the redux history.
if the polling is slower than the server process, we might see something like:
```
- Writing host 1/10
- Writing host 6/10 
- Writing host 7/10
```
instead of a smooth process that shows 1-10 gradually.

ShimShtein: The server supplies full output each poll anyway. I don't think there is a justification for cursor/pagination here.

Watching this screen online is a rare case anyway.

laviro: gotcha, thanks

PR #40, opened 2019-07-25
Add queue endpoint

laviro:


app/controllers/foreman_yupana/uploads_controller.rb
```diff
@@ -8,5 +8,13 @@ def last
         output: output
       }, status: :ok
     end
+
+    def queue
+      directory = Dir["../foreman/#{ForemanYupana.base_folder}*"].select{ |f| File.file? f }.map{ |f| File.basename f }
```
ShimShtein: Files ready for uploads will land in [`uploads_folder`](https://github.com/laviro/foreman_yupana/blob/master/lib/foreman_yupana.rb#L9)

laviro: Thanks, updated

webpack/ForemanYupana/Components/Dashboard/DashboardActions.js
```diff
@@ -48,6 +49,27 @@ export const fetchLogs = () => async (dispatch, getState) => {
...
+      type: YUPANA_GET_QUEUE,
+      payload: {
+        queue,
+      },
+    });
+  } catch (error) {
+    dispatch({
+      type: YUPANA_GET_QUEUE,
+      payload: {
+        queue: error.message,
```
ShimShtein: Does it make sense to put the error in a different field? This way we can be a bit more type safe.

laviro: sure, 
by the way, do we want to inform the user if there is a network error or a problem with getting the process output? or should it be visible only in the terminal?

ShimShtein: Yes, there is a difference between unsuccessful upload process (which is sort of normal) as opposed to problem with getting data from our server (which is a problem).

laviro: Thanks, now when there is an error retrieving the queue - it will show:
Selection_171

laviro:
Closing this for now due to design changes.

PR #45, opened 2019-07-29
Add download_file action in uploads

laviro:


config/routes.rb
```diff
@@ -3,5 +3,6 @@
     get 'index', to: 'react#index'
     get 'reports/last', to: 'reports#last'
     get 'uploads/last', to: 'uploads#last'
+    get 'uploads/download/:name', to: 'uploads#download_file'
```
ShimShtein: I would change it to:

``` ruby
    get ':portal_user/uploads/file', to: 'uploads#download_file'

```

app/controllers/foreman_yupana/uploads_controller.rb
```diff
@@ -8,5 +8,36 @@ def last
...
+          'Expected an actual string for the file name'
+        )
+      end
+
+      if filename.include? '/'
+        return throw_flash_error(
+          'Illegal file name'
+        )
+      end
+      path = Rails.root.join(ForemanYupana.uploads_folder, filename)
```
ShimShtein:
```suggestion
      path = Rails.root.join(ForemanYupana.uploads_folder(params[:portal_user]), 'hosts_report.tar.gz')
```

laviro:
rebased and fixed with your suggestions @ShimShtein

PR #55, opened 2019-08-01
add foreman vendor

amirfefer:


laviro:
Thanks for your contribution @amirfefer,
currently we are trying to be compatible with older versions of Foreman.
We'll have a look at this PR after the release for those versions.

ShimShtein:
While we are at it, we should update [foreman core requirement](https://github.com/laviro/inventory_upload/blob/master/lib/foreman_yupana/engine.rb#L24) to the first version that works with vendor.

ehelms:
This would be awesome  to get in for 1.24 as it brings up the compatibility of this plugin and allows us to drop a lot of dependencies in packaging that are only being used by this plugin.

adamruzicka:
Since I don't know if @laviro would be able to rebase this, I took it over in https://github.com/laviro/foreman_inventory_upload/pull/99

laviro:
Thanks guys, back from the holidays :)
Don't worry @ehelms, the version that will support Foreman 1.24 will be using the vendor,
right now we are supporting older Foreman versions first and waiting for some feedback,
later we will create a new branch and add it.

ehelms:
Could you support older versions via a stable branch so that nightly can clean up some nodejs cruft?

ShimShtein:
@ehelms, we will branch once the version is stabilized enough. Right now there a a couple of features in the pipeline (a couple of added fields to the report and switching to per-org report) that should probably make it to the first release.

laviro:
upgrading to the newest versions in #113

PR #67, opened 2019-08-04
Change Yupana menu button icon and text

laviro:
Change it into something that makes more sense:

Selection_184

I tried to enter "Inventory Uploads" though it was too long..


laviro:
Changed to "RH Inventory"

PR #72, opened 2019-08-11
Rename foreman yupana to inventory upload

ShimShtein:


laviro:
Thanks @ShimShtein, I've tested this PR and found no regressions,
the plugin functions has it should,

could you just fix the linting errors?

I also found couple of file leftovers with the word 'yupana',
should they be removed too?:
- tx/config
- locale/en/foreman_yupana.po
- locale/foreman_yupana.pot
- locale/gemspec.rb
- locale/Makefile

Thanks

PR #80, opened 2019-08-14
Add Full screen button and fix redux undefined tab

laviro:


ShimShtein:
Screenshots:
main screen:
image

full screen:
image

laviro:
Yeah, from what I see the 2 terminal modals are overlapping each other..
I will put the modal in a separate file and connect it to redux so there will be only one modal
which will get updated by the current active tab logs.

laviro:
Fixed that @ShimShtein

ShimShtein:
Merged, thanks @laviro!

PR #83, opened 2019-08-18
add API and display scheduled report generating time

laviro:
Selection_194


app/controllers/foreman_inventory_upload/reports_controller.rb
```diff
@@ -1,11 +1,18 @@
...
+
 module ForemanInventoryUpload
   class ReportsController < ::ApplicationController
     def last
       label = ForemanInventoryUpload::Async::GenerateReportJob.output_label(params[:portal_user])
       output = ForemanInventoryUpload::Async::ProgressOutput.get(label)&.full_output
+      scheduled = ForemanTasks::Task.where(
+        :label => 'ForemanInventoryUpload::Async::GenerateAllReportsJob',
+        :state => 'scheduled'
+      )[0]&.start_at || nil
```
ShimShtein: Instead of `[0]`, you can use `first`

laviro:
Updated

app/controllers/foreman_inventory_upload/reports_controller.rb
```diff
@@ -1,11 +1,18 @@
+# frozen_string_literal: true
+
 module ForemanInventoryUpload
   class ReportsController < ::ApplicationController
     def last
       label = ForemanInventoryUpload::Async::GenerateReportJob.output_label(params[:portal_user])
       output = ForemanInventoryUpload::Async::ProgressOutput.get(label)&.full_output
+      scheduled = ForemanTasks::Task.where(
+        :label => 'ForemanInventoryUpload::Async::GenerateAllReportsJob',
```
ShimShtein: Please use [`humanized_name`](https://github.com/theforeman/foreman/blob/develop/app/jobs/application_job.rb#L2) instead of the string.

laviro: Thanks, used the `singleton_job_name` as we discussed offline

ShimShtein:
Merged, thanks @laviro!

PR #90, opened 2019-09-05
Add Satellite instance ID to facts

adamruzicka:
reports the instance id added in [this PR](https://github.com/theforeman/foreman/pull/7003) as a host's fact


ares:
It would be super awesome (and probably useful and required) to also add Satellite version. Can we get it from branding plugin somehow? The core PR has been merged btw.

ShimShtein:
Yes, branding exposes `ForemanThemeSatellite::SATELLITE_VERSION` and `ForemanThemeSatellite::SATELLITE_SHORT_VERSION`

lib/foreman_inventory_upload/generators/slice.rb
```diff
@@ -37,6 +37,7 @@ def report_host(host)
           @stream.simple_field('account', host.subscription_facet.pools.first.account_number.to_s)
           @stream.simple_field('subscription_manager_id', host.subscription_facet.uuid)
           @stream.simple_field('satellite_id', host.subscription_facet.uuid)
+          @stream.simple_field('satellite_instance_id', Foreman.instance_id)
```
ShimShtein: It will not be recognized here. Non-canonical fields should be added in `facts` node. See https://github.com/laviro/foreman_inventory_upload/pull/87 for more details.

ares:
@ShimShtein any more comments?

ShimShtein:
Looks good, once I get a first release out, I'll give it a proper testing round and get it in.

lib/foreman_inventory_upload/generators/slice.rb
```diff
@@ -97,13 +97,24 @@ def report_system_profile(host)
...
         end
-        @stream.array_field('installed_packages', :last) do
+        @stream.array_field('installed_packages') do
           first = true
           host.installed_packages.each do |package|
             @stream.raw("#{first ? '' : ', '}#{@stream.stringify_value(package.nvra)}")
             first = false
           end
         end
+        @stream.array_field('facts', :last) do
```
ShimShtein: facts should live under `host` node, as a sibling to [`system_profile`](https://github.com/laviro/foreman_inventory_upload/pull/87/files#diff-bc5727daebf3e0d3a95eec0e1b9f4975R55). Not inside the profile itself.

lib/foreman_inventory_upload/generators/slice.rb
```diff
@@ -97,13 +97,24 @@ def report_system_profile(host)
...
+          @stream.object do
+            @stream.simple_field('namespace', 'satellite')
+            @stream.object_field('facts', :last) do
+              if defined?(ForemanThemeSatellite)
+                @stream.simple_field('satellite_version', ForemanThemeSatellite::SATELLITE_VERSION)
+              end
+              @stream.simple_field('satellite_instance_id', Foreman.instance_id, :last)
```
ShimShtein:
```suggestion
              @stream.simple_field('satellite_instance_id', Foreman&.instance_id, :last)
```

I intend using this plugin in versions before 1.24, where `instance_id` is not there yet. Can we add defensive code here?

adamruzicka: Do we want to keep the key even if we don't know the value?

ShimShtein: `simple_field` knows to deal with that. It will not write the field at all.

ares:
could we also please add organization identifier? we will need to identify specific organization's receptor, hence the mapping is needed

ares:
I think it would be better if that's account id (I believe stored somewhere in Katello DB). We could use internal ID too, but if we ever need to do some mapping on cloud side between Satellite org and account, we'll need to change something on on-prem side.

lib/foreman_inventory_upload/generators/slice.rb
```diff
@@ -57,7 +57,12 @@ def report_host(host)
               @stream.simple_field('namespace', 'satellite')
               @stream.object_field('facts', :last) do
                 @stream.simple_field('virtual_host_name', host.subscription_facet.hypervisor_host&.name)
-                @stream.simple_field('virtual_host_uuid', host.subscription_facet.hypervisor_host&.subscription_facet&.uuid, :last)
+                @stream.simple_field('virtual_host_uuid', host.subscription_facet.hypervisor_host&.subscription_facet&.uuid)
+                @stream.simple_field('organization_id', host.organization_id)
+                if defined?(ForemanThemeSatellite)
+                  @stream.simple_field('satellite_version', ForemanThemeSatellite::SATELLITE_VERSION)
+                end
+                @stream.simple_field('satellite_instance_id', Foreman.respond_to?(:instance_id) ? Foreman.instance_id : nil, :last)
```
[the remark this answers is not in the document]
adamruzicka: Done

PR #93, opened 2019-09-10
make babel module resolver dependency optional

laviro:


ShimShtein:
Merged, thanks @laviro!

PR #94, opened 2019-09-10
fix List view initExpanded warning by upgrading patternfly-react

laviro:
the problem: the first list view item does not initially opens as expected with a warning seen.

while trying to align to Foreman 1.20.1 dependencies versions,
we used patternfly-react version 2.19.1
it seem that there was a fix to the `initExpanded` prop which is only available from version 2.29.12.
found no other regression while bumping its version, 
the first list view item initially opens as expected with no warning.


laviro:
already fixed when upgrading to vendor in #113

PR #97, opened 2019-10-10
Add ENV to set up test upload URL and set the default to prod

ShimShtein:


lib/foreman_inventory_upload.rb
```diff
@@ -38,7 +38,8 @@ def self.facts_archive_name
   end
 
   def self.upload_url
-    'https://ci.cloud.paas.psi.redhat.com/api/ingress/v1/upload'
+    # for testing set ENV to 'https://ci.cloud.paas.psi.redhat.com/api/ingress/v1/upload'
+    ENV['SATELLITE_INVENTORY_UPLOAD_URL'] || 'https://cloud.redhat.com/api/ingress/v1/upload'
```
adamruzicka: Wouldn't it be better to `@upload_url ||= ENV['something'] || 'url'`?

ShimShtein:
@adamruzicka, updated. Nice catch!

ShimShtein:
Merged, @adamruzicka, thanks for the review!

PR #99, opened 2019-10-18
Add Foreman vendor

adamruzicka:


laviro:
upgrading to the newest versions in #113

PR #106, opened 2019-11-27
Use foreman-rake in production instead of rake

ShimShtein:


laviro:
Thanks @ShimShtein, works as expected :+1:

PR #107, opened 2019-12-05
Fix done folder path

ShimShtein:


laviro:
As we discussed offline, this is the right behavior and after the report have been uploaded completely it will move to the `/done` folder, test errors are non related, merging

ShimShtein:
It will go to `done` once the script succeeds to upload the file.

PR #108, opened 2019-12-05
Fix default prod endpoint

ShimShtein:


laviro:
rubocop failure is unrelated

PR #109, opened 2019-12-05
Use plugin:rubocop rake task instead of running rubocop locally

ShimShtein:


ShimShtein:
@laviro , as you can see, this fixes rubocop run issues :)
Now all we need is to be aligned to foreman core.

PR #112, opened 2020-01-12
Use katello http proxy setting

ShimShtein:
Cherrypick from master (750f0ad0174aaa6d58062c4c7cf9e4df4bfde283)


ShimShtein:
@laviro, can you take a quick look before I merge it?

PR #118, opened 2020-02-05
convert memsize from kb to bytes

laviro:
still work in progress


ShimShtein:
Please add a test for it.

ShimShtein:
Merged, thanks @laviro!

ShimShtein:
Now we need to CP it to both foreman 24 and foreman 20 branches.

PR #120, opened 2020-02-06
import isEqual directly from lodash

laviro:
on the packaging PR, koji was complaining it can't find the `lodash/isEqual` module,
changing the import fixed it.


ShimShtein:
Merged, thanks @laviro!

PR #129, opened 2020-02-16
Insights rules import

ShimShtein:
Provides the ability to store and show rules in foreman screens.


ShimShtein:
@laviro, ready for review

lib/foreman_rh_cloud/engine.rb
```diff
@@ -37,8 +37,13 @@ class Engine < ::Rails::Engine
         role 'ForemanRhCloud', [:view_foreman_rh_cloud]
 
         # Adding a sub menu after hosts menu
-        sub_menu :top_menu, :foreman_rh_cloud, :caption => N_('RH Cloud'), :icon => 'fa fa-cloud-upload' do
-          menu :top_menu, :level1, :caption => N_('Inventory Upload'), :url_hash => { controller: :'foreman_inventory_upload/react', :action => :index}
+        sub_menu :top_menu, :rh_cloud, :caption => N_('RH Cloud Services'), :icon => 'fa fa-cloud' do
```
laviro, line 40: nit: name is a bit long

Selection_311

how about Cloud Services?
Selection_312

lib/foreman_rh_cloud/engine.rb
```diff
@@ -37,8 +37,13 @@ class Engine < ::Rails::Engine
...
+          menu :top_menu, :inventory_upload, :caption => N_('Manage inventory upload'), :url_hash => { controller: :'foreman_inventory_upload/react', :action => :index}
+          menu :top_menu, :insights_hits_import, :caption => N_('Manage insights hits import'), :url_hash => { controller: :'insights_cloud/react', :action => :index}
```
laviro, line 42: This is also getting a bit long, can we use dividers, e.g:
```rb
divider :top_menu, :parent => :rh_cloud, :caption => N_('Inventory upload')
divider :top_menu, :parent => :rh_cloud, :caption => N_('Insights')
```

webpack/InsightsCloudSync/InsightsCloudSync.js
```diff
@@ -0,0 +1,18 @@
+import React from 'react';
+import { IntlProvider } from 'react-intl';
+import { translate as __ } from 'foremanReact/common/I18n';
+import { Button } from 'patternfly-react';
+import { syncInsights } from './InsightsCloudSyncActions';
+
+const InsightsCloudSync = () => (
+  <IntlProvider locale={navigator.language}>
```
laviro, line 8: maybe we should to combine `webpack/InsightsCloudSync` and `webpack/ForemanInventoryUpload` into a one
`webpack/ForemanRhCloud` folder, so we won't have duplication of intl and other registries

webpack/InsightsCloudSync/InsightsCloudSync.js
```diff
@@ -0,0 +1,18 @@
...
+    <div className="insights-cloud-sync">
+      <h1>{__('Red Hat Insights Sync')}</h1>
+      <Button bsStyle="primary" onClick={syncInsights()}>
+        {__('Sync now')}
+      </Button>
```
laviro, line 13: I think we should move it into its own component folder,
and `cconnect` that component to `redux` so we could pass the action properly

lib/foreman_rh_cloud/engine.rb
ShimShtein: Sounds good, but a bit unclear what cloud services do we mean. I suggest "Cloud services", but with RH logo instead of the cloud. Can we arrange it?

laviro: @rohoover @MariSvirik do you have any ideas?
I know we are using font-awesome old free version where the RH icon doesn't exist,
is there a RH icon we can use?

webpack/InsightsCloudSync/InsightsCloudSyncReducers.js
```diff
@@ -0,0 +1,3 @@
+const reducers = {};
```
laviro, line 1: since we are not updating the redux store with some data,
I would be OK with avoiding the use of redux on this early stage.. unless we are planning to add more components to the page

webpack/InsightsCloudSync/InsightsCloudSync.js
ShimShtein: I suppose we can create sort of layout component for both pages. Right now it's only the `IntlProvider` component, so I suggest to leave this refactoring to a different PR

webpack/InsightsCloudSync/InsightsCloudSyncReducers.js
ShimShtein: I suppose we will need to add more components later. I think leaving this boiler plate code makes sense.

webpack/InsightsCloudSync/InsightsCloudSync.js
laviro: can you move it into another component?
e.g: `webpack/InsightsCloudSync/SyncButton`

webpack/InsightsCloudSync/InsightsCloudSyncSelectors.js
```diff
@@ -0,0 +1,4 @@
+// export const selectForemanInventoryUpload = state => state.ForemanInventoryUpload || {};
+// Added 'state' to use redux in storybook only, should be removed.
+export const selectInsightsCloudSync = state =>
+  state.InsightsCloudSync || state || {};
```
laviro, line 4: can you move the selector to https://github.com/theforeman/foreman_rh_cloud/blob/master/webpack/ForemanRhCloudSelectors.js#L2

webpack/InsightsCloudSync/InsightsCloudSyncActions.js
```diff
@@ -0,0 +1,10 @@
+import API from 'foremanReact/API';
+import { insightsCloudUrl } from './InsightsCloudSyncHelpers';
+
+export const syncInsights = () => dispatch => {
+  API.post(insightsCloudUrl('tasks'));
```
laviro, line 5: I think that right now because there's no reducer that needs to act upon this action
this code can be moved to the button component to an event handler:
```js
...
const handleClick = () => {
  API.post(insightsCloudUrl('tasks'))
}

...
return <Button onClick={handleClick} />
```

ShimShtein:
@laviro, can you give it another round?

ShimShtein:

webpack/InsightsCloudSync/InsightsCloudSync.js
```diff
@@ -0,0 +1,24 @@
+import React from 'react';
+import { IntlProvider } from 'react-intl';
+import { translate as __ } from 'foremanReact/common/I18n';
+import { Button } from 'patternfly-react';
+import { syncInsights } from './InsightsCloudSyncActions';
```
laviro, line 5: you will receive the actions through component props, after the component will be connected the redux.
can you connect it in the `index.js` file?
as a reference you can copy the [AccountList/index.js](https://github.com/theforeman/foreman_rh_cloud/blob/master/webpack/ForemanInventoryUpload/Components/AccountList/index.js)

webpack/InsightsCloudSync/InsightsCloudSyncHelpers.js
```diff
@@ -0,0 +1,3 @@
+import { foremanUrl } from '../ForemanRhCloudHelpers';
+
+export const insightsCloudUrl = path => foremanUrl(`/insights_cloud/${path}`);
```
laviro, line 3: can you add a test for that?

webpack/InsightsCloudSync/InsightsCloudSyncActions.js
```diff
@@ -0,0 +1,11 @@
...
+import { insightsCloudUrl } from './InsightsCloudSyncHelpers';
+import { INSIGHTS_CLOUD_SYNC_TASK } from './InsightsCloudSyncConstants';
+
+export const syncInsights = () => {
+  API.post(insightsCloudUrl('tasks'));
+  return {
+    type: INSIGHTS_CLOUD_SYNC_TASK,
+    payload: {},
+  };
+};
```
laviro, line 11: can you add a test for the action?

webpack/InsightsCloudSync/InsightsCloudSyncActions.js
```diff
@@ -0,0 +1,20 @@
...
+export const syncInsights = () => async dispatch => {
+  try {
+    await API.post(insightsCloudUrl('tasks'));
+    dispatch({
+      type: INSIGHTS_CLOUD_SYNC_TASK,
+      payload: {},
+    });
+  } catch (error) {
+    dispatch({
+      type: INSIGHTS_CLOUD_SYNC_TASK,
```
laviro, line 14: we should distinguish between the `"succes"/"error"` responses
can you rename them to something like: `INSIGHTS_SYNC_SUCCESS`, `INSIGHTS_SYNC_FAILURE`?

app/overrides/hosts_list.rb
```diff
@@ -0,0 +1,13 @@
...
+  name: 'insights_hits_header',
+  insert_before: 'thead tr th.hidden-xs:first-of-type',
+  text: '<th class="hidden-xs" width="10%"><%= _("Insights")%></th>'
+)
+
+Deface::Override.new(
+  virtual_path: 'hosts/_list',
+  name: 'insights_hits_cells',
+  insert_before: 'tbody tr td.hidden-xs:first-of-type',
+  text: '<%= hits_counts_cell(host) %>'
```
laviro, line 12: are we able to pass a react component here?

lib/foreman_rh_cloud/engine.rb
```diff
@@ -38,7 +38,12 @@ class Engine < ::Rails::Engine
 
         # Adding a sub menu after hosts menu
         sub_menu :top_menu, :foreman_rh_cloud, :caption => N_('RH Cloud'), :icon => 'fa fa-cloud-upload' do
-          menu :top_menu, :level1, :caption => N_('Inventory Upload'), :url_hash => { controller: :'foreman_rh_cloud/react', :action => :inventory_upload}
+          menu :top_menu, :inventory_upload, :caption => N_('Inventory Upload'), :url_hash => { controller: :'foreman_rh_cloud/react', :action => :inventory_upload }
+          menu :top_menu, :insights_hits_import, :caption => N_('Manage insights hits import'), :url_hash => { controller: :'foreman_rh_cloud/react', :action => :insights_cloud }
```
laviro, line 42: what do you think about renaming to "Insights sync"?

webpack/ForemanRhCloudSelectors.js
```diff
@@ -1,3 +1,5 @@
 export const selectForemanRhCloud = state => state.ForemanRhCloud;
 export const selectForemanInventoryUpload = state =>
   selectForemanRhCloud(state).inventoryUpload;
+export const selectInsightsCloudSync = state =>
+  selectForemanRhCloud(state).InsightsCloudSync;
```
laviro, line 5: can you add a test for it in the `ForemanRhCloudSelectors.test.js`?

app/overrides/hosts_list.rb
ShimShtein: Theoretically yes, but I don't see a reason to do that. The table is completely non-react, so I don't see much benefit from making this specific cell a react component.

lib/foreman_rh_cloud/engine.rb
ShimShtein: I don't want to use ambiguous word "sync" - this page should also contain all settings related to remediations e.t.c.

What about "Insights data"?

laviro: > What about "Insights data"?

works for me, maybe even just "Insights" could fit than

laviro:
@ShimShtein can you rebase?
travis issue should be gone now after https://github.com/theforeman/foreman-tasks/pull/551 got merged

ShimShtein:
@laviro, I think it means we need also pin the requirement in katello, since bundler doesn't read `requires_foreman` directive.

I see Katello are defining only minimal version of foreman-tasks, so we need to pin the version in our own deployment.

app/helpers/foreman_inventory_upload_host_helper.rb
```diff
@@ -0,0 +1,11 @@
+module ForemanInventoryUploadHostHelper
+  def hits_counts
+    @hits_counts ||= InsightsHit.where(host_id: @hosts).group(:host_id).count
+  end
+
+  def hits_counts_cell(host)
+    host_hits = hits_counts[host.id]
+    content = "#{host_hits} Rule hits" if host_hits
```
laviro, line 8: Can you rename `hits` to `recommendations` ?

ShimShtein:
@laviro , rebased and changed the text as we decided.

webpack/InsightsCloudSync/InsightsCloudSync.js
```diff
@@ -0,0 +1,26 @@
+import React from 'react';
+import { IntlProvider } from 'react-intl';
+import PropTypes from 'prop-types';
+import { translate as __ } from 'foremanReact/common/I18n';
+import { Button, noop } from 'patternfly-react';
+
+const InsightsCloudSync = ({ syncInsights }) => (
+  <IntlProvider locale={navigator.language}>
```
laviro, line 8: can you set here the document title
`document.title = INSIGHTS_SYNC_PAGE_TITLE;`

webpack/__tests__/ForemanRhCloudSelectors.test.js
```diff
@@ -11,6 +12,7 @@ const fixtures = {
   'should return ForemanRhCloud': () => selectForemanRhCloud(state),
   'should return ForemanInventoryUpload': () =>
     selectForemanInventoryUpload(state),
+  'should return InsightsCloudSync': () => selectInsightsCloudSync(state),
```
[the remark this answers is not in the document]
ShimShtein: It's already there: https://github.com/theforeman/foreman_rh_cloud/pull/129/files#diff-779be5162c4d6bf6d8b385a1253f625aR4

webpack/ForemanRhCloudTestHelpers.js
```diff
@@ -1,5 +1,6 @@
 export const inventoryStateWrapper = innerState => ({
   ForemanRhCloud: {
     inventoryUpload: { ...innerState },
+    InsightsCloudSync: { ...innerState },
```
laviro, line 4: oh didn't see it, IMO we should create a separate state wrapper or rename the current `inventoryStateWrapper` to `RhCloudStateWrapper`

ShimShtein: Makes sense, fixed.

config/routes.rb
```diff
@@ -8,7 +8,12 @@
     post 'auto_upload', to: 'uploads#auto_upload'
   end
 
+  namespace :insights_cloud do
+    resources :tasks
```
adamruzicka, line 12: This generates seven routes, but unless I'm missing something `InsightsCloud::TasksController` implements only `#create`

app/models/setting/rh_cloud.rb
```diff
@@ -5,6 +5,7 @@ def self.default_settings
     [
       set('allow_auto_inventory_upload', N_('Allow automatic upload of the host inventory to the Red Hat cloud'), true),
       set('obfuscate_inventory_hostnames', N_('Obfuscate host names sent to Red Hat cloud'), false),
+      set('rh_cloud_token', N_('Authentication token to RH cloud services. Used to authenticate requests to Red Hat cloud APIs'), 'DEFAULT', N_('RH Cloud token'), nil, encrypted: true),
```
adamruzicka, line 8: There's `RH cloud`, `Red Hat cloud` and `RH Cloud` on this line. This should probably be unified

app/models/insights_hit.rb
```diff
@@ -0,0 +1,2 @@
+class InsightsHit < ApplicationRecord
```
adamruzicka, line 1: Shouldn't this define the hit-to-facet association? Just for the sake of completeness

db/migrate/20191216062008_create_insights_facets.foreman_inventory_upload.rb
```diff
@@ -0,0 +1,7 @@
+class CreateInsightsFacets < ActiveRecord::Migration[5.2]
+  def change
+    create_table :insights_facets do |t|
+      t.references :host, foreign_key: true
```
adamruzicka, line 4: Shouldn't it also have a `null: false`?

db/migrate/20191215104806_create_insights_hits.foreman_inventory_upload.rb
```diff
@@ -0,0 +1,14 @@
+class CreateInsightsHits < ActiveRecord::Migration[5.2]
+  def change
+    create_table :insights_hits do |t|
+      t.references :host, foreign_key: true
```
adamruzicka, line 4: Shouldn't this reference the insights facet instead of the host itself? In any way, I'd add `null: false`

ShimShtein:
@adamruzicka, fixed your comments. Let's wait for test to go green, and do another round.

db/migrate/20191215104806_create_insights_hits.foreman_inventory_upload.rb
ShimShtein: Since I need to match between host names in foreman to hostnames that come from Insights, it was easier not to query extra ID.

I have created a connection through `host_id` column on hit and on facet. Since the facet is 1:1 to host, it would work as expected.

lib/insights_cloud/async/insights_full_sync.rb
```diff
@@ -0,0 +1,90 @@
...
+
+      def host_id(host_name)
+        @host_ids[host_name]
+      end
+
+      def replace_hits_data(hits)
+        InsightsHit.transaction do
+          InsightsHit.delete_all
+          InsightsHit.create(hits.map { |hits_hash| to_model_hash(hits_hash) })
+          InsightsFacet.create(@host_ids.values.map { |id| {host_id: id} })
```
adamruzicka, line 72: What happens here if the facet already exists?

app/models/setting/rh_cloud.rb
```diff
@@ -5,6 +5,7 @@ def self.default_settings
     [
       set('allow_auto_inventory_upload', N_('Allow automatic upload of the host inventory to the Red Hat cloud'), true),
       set('obfuscate_inventory_hostnames', N_('Obfuscate host names sent to Red Hat cloud'), false),
+      set('rh_cloud_token', N_('Authentication token to Red Hat cloud services. Used to authenticate requests to cloud APIs'), 'DEFAULT', N_('Red Hat Cloud token'), nil, encrypted: true),
```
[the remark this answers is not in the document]
ShimShtein: `Red Hat cloud **services**` vs `Red Hat Cloud` is it still weird for you?

ShimShtein:
@adamruzicka, fixed existing facets filter

app/models/setting/rh_cloud.rb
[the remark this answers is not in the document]
adamruzicka: I just wanted to check if this was intentional. To be honest I'm not sure what the rules for capitalization are.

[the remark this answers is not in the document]
ShimShtein: As long as it doesn't feel weird to you, we can have it. We can always change it later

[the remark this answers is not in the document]
adamruzicka: Alright, let's keep it for now

PR #138, opened 2020-02-23
WIP - add insights extension to hosts index

laviro:
Selection_269


laviro:
in this POC we proved that we can extend the tables in foreman,
and import into them the insights front-end components, all of those from this plugin.
closing it for now

PR #139, opened 2020-02-26
fix host interface undefined mtu error

laviro:


ShimShtein:
Merged, thanks @laviro!

PR #140, opened 2020-03-04
Fix older katello version without default_global_content_proxy method

ShimShtein:


lib/foreman_inventory_upload/async/upload_report_job.rb
```diff
@@ -37,7 +37,11 @@ def env
       end
 
       def http_proxy_string
-        @http_proxy_string ||= HttpProxy.default_global_content_proxy&.full_url
+        @http_proxy_string ||= begin
+          if Setting[:content_default_http_proxy]
+            HttpProxy.unscoped.find_by(name: Setting[:content_default_http_proxy])
+          end
+        end&.full_url
```
adamruzicka: I'd probably move the `&.full_url` inside the `if`, but that's just a personal preference

adamruzicka:
Thank you @ShimShtein !

PR #145, opened 2020-03-15
Update rubocop to use foreman develop settings

laviro:


.rubocop.yml
```diff
@@ -1,2 +1,2 @@
 inherit_from:
-  - https://raw.githubusercontent.com/theforeman/foreman/4aa0bb6a24b45347ff9b673055499763eec72489/.rubocop.yml
+  - https://raw.githubusercontent.com/theforeman/foreman/develop/.rubocop.yml
```
ShimShtein: This will cause rubocop to fail each time core changes its definition.

test/factories/inventory_upload_factories.rb
```diff
@@ -23,50 +23,50 @@
...
+      composite { true }
     end
   end
 end
 
 FactoryBot.define do
   factory :katello_k_t_environment, :aliases => [:katello_environment], :class => Katello::KTEnvironment do
-    sequence(:name) { |n| "Environment#{n}" }
-    sequence(:label) { |n| "environment#{n}" }
+    sequence(:name) { |n| "Environment#{n} " }
```
ShimShtein: This doesn't make sense, why the whitespace inside the string?

laviro: oops, good catch !

.rubocop.yml
laviro: IMO each time it will be a very small change,
instead of updating to the latest commit with lots of changes to our files.

test/factories/inventory_upload_factories.rb
```diff
@@ -49,24 +49,24 @@
 
 FactoryBot.define do
   factory :katello_subscription_facets, :aliases => [:subscription_facet], :class => ::Katello::Host::SubscriptionFacet do
-    sequence(:uuid) { |n| "uuid-#{n}-#{rand(500)}" }
-    facts { {'memory.memtotal' => "12 GB"} }
+    sequence(:uuid) { |n| "uuid-#{n}-#{rand(500)} " }
```
ShimShtein: This one added space inside a string

ShimShtein:
Merged. Thanks @laviro!

PR #146, opened 2020-03-16
Add Auto Upload Switcher

laviro:
https___localhost_3000_foreman_inventory_upload_index

What's happening in this PR:
- Added Auto Upload Switcher
- Added API post request that changes the `allow_auto_inventory_upload` setting
- `isAutoUpload` is being updated live with the account statuses
- the UI scheduled run is hidden when auto upload is 'OFF'
- updated tests
- Added PageHeader component


ShimShtein:
Please rebase, the setting is already in

laviro:
updated, and tried to split it into several commits
so it would be also easier to understand and to cherry pick

laviro:
rebased, ready for another review

app/controllers/foreman_inventory_upload/uploads_controller.rb
```diff
@@ -27,5 +27,13 @@ def throw_flash_error(message)
         :error_msg => message
       )
     end
+
+    def toggle_auto_upload
```
ShimShtein: I think it would be better to stick with REST notion:

``` ruby
def auto_upload(value)
  Setting[:allow_auto_inventory_upload] = value
end
```

This way we are not dependent on UI implementation

app/controllers/foreman_inventory_upload/accounts_controller.rb
```diff
@@ -25,6 +25,7 @@ def index
       ]
 
       render json: {
+        isAutoUpload: Setting[:allow_auto_inventory_upload],
```
ShimShtein:
```suggestion
        autoUploadEnabled: Setting[:allow_auto_inventory_upload],
```

Sounds a bit better

app/controllers/foreman_inventory_upload/uploads_controller.rb
```diff
@@ -27,5 +27,13 @@ def throw_flash_error(message)
...
+      toggled_auto_upload = !Setting[:allow_auto_inventory_upload]
+      Setting[:allow_auto_inventory_upload] = toggled_auto_upload
+      render json: {
+        isAutoUpload: toggled_auto_upload,
```
ShimShtein:
```suggestion
        autoUploadEnabled: toggled_auto_upload,
```

config/routes.rb
```diff
@@ -6,5 +6,6 @@
     get ':organization_id/uploads/last', to: 'uploads#last', constraints: { organization_id: %r{[^\/]+} }
     get ':organization_id/uploads/file', to: 'uploads#download_file', constraints: { organization_id: %r{[^\/]+} }
     get 'accounts', to: 'accounts#index'
+    post 'toggle_auto_upload', to: 'uploads#toggle_auto_upload'
```
ShimShtein:
```suggestion
    post 'auto_upload', to: 'uploads#auto_upload'
```
And the value will be passed as a prameter

laviro:
Thanks @ShimShtein, updated

webpack/ForemanInventoryUpload/Components/AccountList/AccountListSelectors.js
```diff
@@ -6,3 +6,5 @@ export const selectAccounts = state => selectAccountsList(state).accounts;
 export const selectPollingProcessID = state =>
   selectAccountsList(state).pollingProcessID;
 export const selectError = state => selectAccountsList(state).error;
+export const selectAutoUploadEnabled = state =>
+  selectAccountsList(state).autoUploadEnabled || false;
```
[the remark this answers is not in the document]
laviro: it's only to ensure that when we receive `undefined` it will be `false`, for example when the state doesn't exist yet.
but maybe you are right, maybe we should leave it as it is, 
and in case of `undefined`, `proptypes` will already take care of it with its default value which will be `true`

ShimShtein:
Merged, thanks @laviro!

PR #149, opened 2020-03-24
BZ 1816147 Select pool with non-empty account

ShimShtein:


lib/foreman_inventory_upload/generators/slice.rb
```diff
@@ -36,7 +36,7 @@ def report_host(host)
         @stream.object do
           @stream.simple_field('display_name', host.name)
           @stream.simple_field('fqdn', host.fqdn)
-          @stream.simple_field('account', host.subscription_facet.pools.first.account_number.to_s)
+          @stream.simple_field('account', host.subscription_facet.pools.where.not(account_number: nil).first&.account_number&.to_s)
```
adamruzicka: If there is no pool with an account number, do we want a `nil` or an empty string or it doesn't matter?

ShimShtein: I don't think it matters. Specifically in this case, the whole `account` fields would be omitted from host.

laviro:
Looks good to me, waiting on tests to go green before merging.

test/unit/slice_generator_test.rb
```diff
@@ -174,4 +174,24 @@ def fact_names
...
+    first_pool = @host.subscription_facet.pools.first
+    second_pool = FactoryBot.create(:katello_pool, account_number: nil, cp_id: 2)
+    @host.subscription_facet.pools = []
+    @host.subscription_facet.save!
+    @host.subscription_facet.pools << first_pool
+    @host.subscription_facet.pools << second_pool
+
+    batch = Host.where(id: @host.id).in_batches.first
+    generator = ForemanInventoryUpload::Generators::Slice.new(batch, [], 'slice_123')
+
```
laviro: nit, rubocop is angry about this: `Layout/TrailingWhitespace: Trailing whitespace detected.`

PR #150, opened 2020-03-25
Fix hosts_count to reduce hosts with no subscription

laviro:


lib/foreman_inventory_upload/generators/archived_report.rb
```diff
@@ -16,9 +16,8 @@ def render(organization:)
               first = true
               host_batches.each do |hosts_batch|
                 slice_id = Foreman.uuid
-                hosts_count = hosts_batch.count
+                hosts_count = generate_slice(tmpdir, slice_id, hosts_batch)
                 @logger.info "Adding slice #{slice_id} with #{hosts_count} hosts"
```
ShimShtein, line 20: Can you move this line before the `generate_slice` call, but without the `hosts_count`?

laviro: Would this work?
```rb
@logger.info "Adding slice #{slice_id}"
hosts_count = generate_slice(tmpdir, slice_id, hosts_batch)
@logger.info "slice #{slice_id} was created with #{hosts_count} hosts"
```

ShimShtein: :+1:

ShimShtein:
Merged, thanks @laviro!

PR #151, opened 2020-03-29
BZ 1817521 Make org id tag string

ShimShtein:


.travis.yml
```diff
@@ -17,7 +17,6 @@ install:
   - node --version
   - npm version
   - nvm version
-  - bundle install --jobs=3 --retry=3
```
laviro: the issue seem to be `rubocop-rails`, removing that [dependency line](https://github.com/theforeman/foreman_rh_cloud/blob/master/foreman_rh_cloud.gemspec#L25) fixed it for me

ShimShtein: Interesting... Doesn't make any sense though.

laviro:
Thanks @ShimShtein !

PR #161, opened 2020-04-05
Fix download file path

laviro:
when file is moved to `uploads/done/` the download file action raise an error that the file doesn't exists in `uploads/`.

in this PR the `download_file` action will look also if the file was moved to the `done/` folder.

I also added `uploads_file_path`, `done_folder`, `done_file_path` properties 
to `lib/foreman_inventory_upload.rb`

### DIFF
--- a/app/controllers/foreman_inventory_upload/uploads_controller.rb
+++ b/app/controllers/foreman_inventory_upload/uploads_controller.rb
@@ -11,14 +11,10 @@ module ForemanInventoryUpload
 
     def download_file
       filename = ForemanInventoryUpload.facts_archive_name(params[:organization_id])
-      path = Rails.root.join(ForemanInventoryUpload.uploads_folder, filename)
-      unless File.exist? path
-        return throw_flash_error(
-          "Path doesn't exist: #{path}"
-        )
-      end
+      files = Dir["{#{ForemanInventoryUpload.uploads_file_path(filename)},#{ForemanInventoryUpload.done_file_path(filename)}}"]
 
-      send_file path, disposition: 'attachment', filename: filename
+      return send_file files.first, disposition: 'attachment', filename: filename unless files.empty?
+      throw_flash_error "File doesn't exist"
     end
 
     def throw_flash_error(message)
--- a/lib/foreman_inventory_upload.rb
+++ b/lib/foreman_inventory_upload.rb
@@ -18,6 +18,18 @@ module ForemanInventoryUpload
     )
   end
 
+  def self.uploads_file_path(filename)
+    File.join(ForemanInventoryUpload.uploads_folder, filename)
+  end
+
+  def self.done_folder
+    File.join(ForemanInventoryUpload.uploads_folder, 'done/')
+  end
+
+  def self.done_file_path(filename)
+    File.join(ForemanInventoryUpload.done_folder, filename)
+  end
+
   def self.generated_reports_folder
     @generated_reports_folder ||= ensure_folder(
       File.join(


app/controllers/foreman_inventory_upload/uploads_controller.rb
```diff
@@ -11,14 +11,18 @@ def last
...
-      send_file path, disposition: 'attachment', filename: filename
+      unless File.exist? uploads_file_path
+        unless File.exist? done_file_path
+          return throw_flash_error(
+            "File doesn't exist in: #{uploads_file_path} -or- #{done_file_path}"
+          )
+        end
+        return send_file done_file_path, disposition: 'attachment', filename: filename
+      end
+      send_file uploads_file_path, disposition: 'attachment', filename: filename
```
adamruzicka, line 25: How about putting the paths into an array, trying to `#find` first one which exists on the filesystem and if there is none found, returning the flash error with message generated from the array?

ShimShtein: While we are at it, we can use `Dir.glob` do search for a file.

laviro: what do you think about this?
@ShimShtein where should we use the `Dir.glob`? Thanks

```rb
def download_file
      filename = ForemanInventoryUpload.facts_archive_name(params[:organization_id])
      file_paths = [
        ForemanInventoryUpload.uploads_file_path(filename),
        ForemanInventoryUpload.done_file_path(filename)
      ]
 
      final_path = file_paths.find do |path|
        path unless path.empty?
      end
      return send final_path, disposition: 'attachment', filename: filename} unless final_path.nil?
      throw_flash_error(
        "File doesn't exist in: #{file_paths.join(',')}"
      )
    end
```

adamruzicka: Instead of the array and find, you could do something like `Dir["#{ForemanInventoryUpload.uploads_folder}/**/#{filename}]` and take a first from it. Given how things currently work, the `Dir[...]` should give you back an array with 0 or 1 elements.

rahulbajaj0509: @laviro we can have the `final_path` in one line, since the block contains only one line something like: `final_path = file_paths.find { |path| path unless path.empty? }`

What do you think?

I have not tested this code but this code will only work if `ForemanInventoryUpload.uploads_file_path(filename) and ForemanInventoryUpload.done_file_path(filename)` gives a empty string (in the case of a failure), but if it gives nil as a value then this code might fail.

laviro: Thanks everyone, changed into this:
```rb
def download_file
  filename = ForemanInventoryUpload.facts_archive_name(params[:organization_id])
  files = Dir["#{ForemanInventoryUpload.uploads_folder}**/#{filename}"]

  return send_file files[0], disposition: 'attachment', filename: filename unless files.empty?
  throw_flash_error "File doesn't exist"
end
```

app/controllers/foreman_inventory_upload/uploads_controller.rb
```diff
@@ -11,14 +11,10 @@ def last
 
     def download_file
       filename = ForemanInventoryUpload.facts_archive_name(params[:organization_id])
-      path = Rails.root.join(ForemanInventoryUpload.uploads_folder, filename)
-      unless File.exist? path
-        return throw_flash_error(
-          "Path doesn't exist: #{path}"
-        )
-      end
+      files = Dir["#{ForemanInventoryUpload.uploads_folder}**/#{filename}"]
```
ShimShtein, line 14: We can be a bit more strict:
```suggestion
      files = Dir["{#{ForemanInventoryUpload.uploads_file_path(filename)},#{ForemanInventoryUpload.done_file_path(filename)}}"]
```

This will look only in done and uploads folder.

laviro: Thanks! updated

ShimShtein:
Merged, thanks @laviro!

PR #170, opened 2020-04-14
When running in satellite, organization_id variable needs to be set

jhutar:
Adding what I just found out when trying to figure out how to run it without UI


ShimShtein:
Awesome! Thanks @jhutar!. Merged.

PR #171, opened 2020-04-16
BZ 1824183 - add infrastructure_type field to system_profile

laviro:
System profile's "infrastructure_type" field should be reported as "virtual" or "physical"


ShimShtein:
Merged, thanks @laviro!

PR #175, opened 2020-04-19
[CP 20] add system_profile infrastructure_type (#171)

ShimShtein:


laviro:
Thanks @ShimShtein,
tested with Foreman 1.20 and 1.22 and it works as expected :)

PR #187, opened 2020-04-30
BZ 1817481 - add inventory page title

laviro:
Selection_313


laviro:
@ShimShtein if we are going to use react-router in #188
the title will be already set in the client side

PR #188, opened 2020-04-30
Refactor webpack folder structure

laviro:
Started as adding react router PR..
though after I found that Foreman doesn't support it yet,
I decided to keep just the webpack folder refactoring, so it would be easier to add more pages.


webpack/ForemanRhCloud/Components/PageHeader/PageHeader.js
```diff
@@ -0,0 +1,13 @@
+import React from 'react';
+import { Grid } from 'patternfly-react';
+import AutoUploadSwitcher from '../AutoUploadSwitcher';
+
+const PageHeader = () => (
+  <Grid.Row>
+    <Grid.Col xs={4} xsOffset={8}>
+      <AutoUploadSwitcher />
```
ShimShtein, line 8: I think this is specific to inventory upload.

webpack/ForemanRhCloud/ForemanRhCloudSelectors.js
```diff
@@ -0,0 +1,4 @@
+// export const selectForemanRhCloud = state => state.ForemanRhCloud || {};
+// Added 'state' to use redux in storybook only, should be removed.
```
ShimShtein, line 2: When should it be removed?

webpack/ForemanRhCloud/Pages/ForemanInventoryUpload.js
```diff
@@ -0,0 +1,17 @@
...
+import PageLayout from 'foremanReact/routes/common/PageLayout/PageLayout';
+import { translate as __ } from 'foremanReact/common/I18n';
+import AccountList from '../Components/AccountList';
+import PageHeader from '../Components/PageHeader';
+
+const ForemanInventoryUpload = () => (
+  <PageLayout header={__('Red Hat Inventory Upload')} searchable={false}>
+    <Grid fluid className="inventory-upload">
+      <PageHeader />
+      <AccountList />
```
ShimShtein, line 12: This one should be either refactored to be generic, or called `InventoryAccountsList` (since it's inventory specific)

webpack/ForemanRhCloud/Routes/ForemanRhCloudRoutes.js
```diff
@@ -0,0 +1,12 @@
+import React from 'react';
+import ForemanInventoryUpload from '../Pages/ForemanInventoryUpload';
+
+const ForemanRhCloudRoutes = {
+  indexInventoryUpload: {
+    path: '/foreman_inventory_upload/index',
```
ShimShtein, line 6: Maybe it should be `/foreman_rh_cloud/inventory_upload`?

lib/foreman_rh_cloud/engine.rb
```diff
@@ -38,7 +38,7 @@ class Engine < ::Rails::Engine
 
         # Adding a sub menu after hosts menu
         sub_menu :top_menu, :foreman_rh_cloud, :caption => N_('RH Cloud'), :icon => 'fa fa-cloud-upload' do
-          menu :top_menu, :level1, :caption => N_('Inventory Upload'), :url_hash => { controller: :'foreman_inventory_upload/react', :action => :index}
+          menu :top_menu, :level1, :caption => N_('Inventory Upload'), :url_hash => { controller: :'foreman_rh_cloud/react', :action => :inventory_upload}
```
ShimShtein, line 41: Does this mean I will need to add an action to `react_controller.rb` for each menu item?

webpack/ForemanInventoryUpload/ForemanInventoryConstants.js
```diff
@@ -0,0 +1,4 @@
+import { translate as __ } from 'foremanReact/common/I18n';
+
+export const INVENTORY_PAGE_TITLE = __('Red Hat Inventory Uploads');
+export const INVENTORY_BASE_URL = 'INVENTORY_BASE_URL';
```
ShimShtein, line 4: I think this will render all URLs to start with `INVENTORY_BASE_URL/`. I don't think it's desired.

lib/foreman_rh_cloud/engine.rb
laviro: basically what I did is in that action, I set the `@component_name` to render
if we can pass that as a `param` it can also work, can we?

webpack/ForemanInventoryUpload/ForemanInventoryConstants.js
laviro: ahah, good catch.. forgot to change it :rofl:

lib/foreman_rh_cloud/engine.rb
ShimShtein: Yes, we can pass it as param.
@sharvit, is it a good practice?

laviro: @sharvit some context,
We wanted to add react router but than we realized there is no registry yet for plugins routes, and we need to support foreman v1.20 too..

What we need is to handle several pages, should we mount each page component and use a sigle react controller?

sharvit: IMO it is too generic, having the `@component_name` and have it rendered by a generic erb layout.

I believe that a better approach would be to use a single controller for react (ReactController#index) but always render the same react-router layout component that would be aware of the route and would render the right component accordingly.

ReactController#index -> render react-router layout -> render the right page based on the location.

sharvit: Once https://github.com/theforeman/foreman/pull/6936 would merge, you would be able to delete some code and reuse the core logic.

laviro: Thanks @sharvit !
That was the first thing that I did, though after realizing it's not going to be a real router
since every page will do a full page reload with the nav layout.

as we need to support also foreman 1.20 and above, I didn't find much value adding it as it will do the same as mounting each component.

another reason not to add to those version is that we need to make node packaging for `react-router-dom` for each version.

I do agree when https://github.com/theforeman/foreman/pull/6936 is merged we should use it and move only the last version to router which is already part of `@theforeman` packages which doesn't need packaging

app/controllers/foreman_rh_cloud/react_controller.rb
```diff
@@ -0,0 +1,7 @@
+module ForemanRhCloud
+  class ReactController < ::ApplicationController
+    def inventory_upload
+      render 'inventory_upload', :layout => false
```
ShimShtein, line 4: I think you don't have to explicitly call `render` here. `inventory_upload.html.erb` will be called implicitly

app/views/foreman_rh_cloud/react/_application.html.erb
```diff
@@ -10,7 +10,6 @@
   <%= notifications %>
   <div id="organization-id" data-id="<%= Organization.current.id if Organization.current %>" ></div>
   <div id="user-id" data-id="<%= User.current.id if User.current %>" ></div>
-  <div id="ForemanInventoryUploadReactRoot"></div>
+  <div id="ForemanRhCloudReactRoot"></div>
```
ShimShtein, line 13: Since it's the base layout for our plugin, we can use convention here too. You can rename this file into `app/views/layouts/application.html.erb` if you want every page to have it, or `app/views/layouts/foreman_rh_cloud/react.html.erb` if you want it for this controller only.

app/views/foreman_rh_cloud/react/inventory_upload.html.erb
```diff
@@ -0,0 +1,3 @@
+
+<%= render "application" %>
```
ShimShtein, line 2: Why do you need it here? If this line is absolutely necessary, you can move it to layout I suppose.

webpack/ForemanInventoryUpload/ForemanInventoryConstants.js
```diff
@@ -0,0 +1,4 @@
+import { translate as __ } from 'foremanReact/common/I18n';
+
+export const INVENTORY_PAGE_TITLE = __('Red Hat Inventory Uploads');
+export const INVENTORY_BASE_URL = '/foreman_inventory_upload/';
```
ShimShtein, line 4: We can't assume foreman lives in root url (`https://example.com`). You can probably use foreman's base function.

webpack/ForemanInventoryUpload/ForemanInventoryHelpers.js
```diff
@@ -0,0 +1,9 @@
+import { INVENTORY_BASE_URL } from './ForemanInventoryConstants';
+
+export const inventoryUrl = path => `${INVENTORY_BASE_URL}${path}`;
+
+export const inventoryStateWrapper = innerState => ({
```
ShimShtein, line 5: Should we have this function here? It's related only to tests.

app/controllers/foreman_rh_cloud/react_controller.rb
laviro: I tried, though we need to specify in some way the `:layout => false` for not using core page layout.. it didn't work for me

webpack/ForemanInventoryUpload/ForemanInventoryHelpers.js
laviro: should we move it to a folder named `testHelpers`?
or maybe we can benefit from it being in the general helpers (?)

laviro:
updated! thanks @ShimShtein for all of the offline help

ShimShtein:
Merged, thanks @laviro!

PR #189, opened 2020-05-03
[CP_1_20] Add React Router

laviro:


laviro:
closing for now as this needs [the Layout component](https://github.com/theforeman/foreman/blob/f7e8785ba53f793ca1ab847bdeb1fa07b77da1d7/webpack/assets/javascripts/react_app/components/Layout/Layout.js#L14) which was added only in Foreman 1.23

PR #193, opened 2020-05-05
Add inventory org filter

laviro:
https___localhost_3000_foreman_inventory_upload_index (1)


laviro:
Thanks @ShimShtein, added test to increase coverage :)

> If an org is selected in Foreman, I guess it would be nice to have it set as a filter.

I guess it's a bit tricky as one org can have access to see other orgs and locations (?)
I think this is something that should be decided on the server side, which orgs to filter for the specific user

ShimShtein:
The list of all available orgs will be filtered on server side, that's right. I was referring to setting the filter to "current org". So by default the screen will show only the current org, but clearing the filter will show also other orgs.

laviro:
Thanks @ShimShtein, that's a great idea, updated
now it is waiting on #188

ShimShtein:
@laviro, can you please rebase?

laviro:
rebased :)

ShimShtein:
Merged, thanks @laviro! Now for the cherrypicks

PR #194, opened 2020-05-05
Publishing js-test coverage from travis

laviro:


.travis.yml
```diff
@@ -19,7 +19,7 @@ install:
   - node --version
   - npm version
   - nvm version
-  - # bundle install --jobs=3 --retry=3
+  -  # bundle install --jobs=3 --retry=3
```
ShimShtein, line 22: I suppose we can remove the line altogether

.travis.yml
```diff
@@ -41,7 +41,7 @@ script:
   - bundle exec rake "${FOREMAN_PLUGIN_NAME}:rubocop"
   - cd ../${FOREMAN_PLUGIN_NAME}
   - npm run test;
-  - # npm run coveralls;
+  - npm run publish-coverage;
```
ShimShtein, line 44: Does this require a specific package in npm?

laviro: this is a script in our package.json: https://github.com/theforeman/foreman_rh_cloud/blob/master/package.json#L11
that comes from `foreman-js` that uploads the test coverage to https://coveralls.io/github/theforeman/foreman_rh_cloud

it should add the coverage line on each PR we have, though I guess it will be added only after this will get merge

ShimShtein:
ACK, do we want it also on other versions/branches, or will it make a mess in coveralls?

laviro:
coveralls recognize the branch the test were running on, so I don't see any issue :)
sure, I will open CPs

PR #200, opened 2020-05-07
Run JS tests only on relevant PRs

laviro:


script/travis_run_js_tests.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+set -ev
+if [[ $( git diff --name-only HEAD~1..HEAD webpack/ .travis.yml babel.config.js .eslintrc package.json | wc -l ) -ne 0 ]]; then
```
ShimShtein, line 3: This does not work for multi-commit PRs

laviro: > Why do we want to incorporate this feature? Are we hitting some limit on Travis currently?

Just to make tests faster, and avoid the extra coverage report when nothing really changed in those folders.

> This does not work for multi-commit PRs

we can compare with upstream/master, right?

ShimShtein:
Merged, thanks @laviro!

PR #204, opened 2020-05-12
Make inventoryBaseUrl dynamic

ShimShtein:


webpack/ForemanInventoryUpload/ForemanInventoryConstants.js
```diff
@@ -2,4 +2,5 @@ import { translate as __ } from 'foremanReact/common/I18n';
 import { foremanUrl } from '../ForemanRhCloudHelpers';
 
 export const INVENTORY_PAGE_TITLE = __('Red Hat Inventory Uploads');
-export const INVENTORY_BASE_URL = foremanUrl('/foreman_inventory_upload/');
+export const INVENTORY_BASE_URL = () =>
+  foremanUrl('/foreman_inventory_upload/');
```
laviro, line 6: @sharvit @mariaaga should we use functions in constants?
or this should be moved to the helper method of `inventoryUrl`?

I don't see an issue with the two approaches, though what should be the convention here? thanks

MariaAga: I think it would make sense more as a helper function like `getInventoryURL` because it's not actually a constant value and you need to "get it" each time.

laviro: agree, makes sense

ShimShtein:
@laviro, done

PR #206, opened 2020-05-17
[CP_24] Run JS tests only on relevant PRs

laviro:


ShimShtein:
Awesome, now for 1.20 branch, and we're done with this one

PR #207, opened 2020-05-17
[CP_24] Add InventoryFilter

laviro:


ShimShtein:
Minor bug I have found:
image
I think we need a separate PR in master for that. Notice the filter isn't cleared when moving to "Any organization". It's only the text that does not change.

laviro:
Thanks @ShimShtein, I created a fix for that on master first: see #210
when it's merged I will add it in this PR

laviro:
added the fix as the last commit, ready for another round @ShimShtein

ShimShtein:
Now we need to cherry pick it into 20

PR #208, opened 2020-05-18
[CP 20] Run JS tests only on relevant PRs

laviro:


script/travis_run_js_tests.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+set -ev
+
+if [[ $( git diff --name-only origin/master..HEAD webpack/ .travis.yml .babelrc .eslintrc package.json | wc -l ) -ne 0 ]]; then
```
ShimShtein, line 4: It should not be against the master branch

laviro: good catch, fixed!

ShimShtein:
Merged. Thanks @laviro!

PR #210, opened 2020-05-19
Fix filter intial org value update

laviro:


webpack/ForemanInventoryUpload/Components/InventoryFilter/InventoryFilter.js
```diff
@@ -5,44 +5,35 @@ import { translate as __ } from 'foremanReact/common/I18n';
 import ClearButton from './Components/ClearButton';
 import './inventoryFilter.scss';
 
-class InventoryFilter extends React.Component {
-  componentDidMount() {
-    const { handleFilterChange, initialValue } = this.props;
-    handleFilterChange(initialValue);
-  }
```
laviro, line 12: This is not needed anymore as we will update the `filterTerm` through the reducer
after the layout initialize and trigger the `LAYOUT_CHANGE_ORG` - which always happen before the rest of the page is loaded. luckily for us, it is being supported since foreman 1.20 :fireworks:

PR #222, opened 2020-06-02
Add host obfuscation switcher

laviro:
Added obfuscation filter, in a Setting section
with a page description to its side:

Selection_332


laviro:
> The settings menu is good, but it hides the current status of settings (which are part of the reason the user opens this page)

Since the requirement have changed a bit and the setting should be visible,
I would suggest re-using the Setting section as seen in the PR description,
especially if this section will grow bigger.

> Red Hat Inventory Uploads :repeat:  :closed_lock_with_key:

I think showing the switchers will be easier to understand than the icons.

ShimShtein:
I don't expect there will be more switches/settings. Maybe some sort of a horizontal layout would take less space and leave the switches visible.

laviro:
Updated the page description with the text we discussed offline:
```text
'Red Hat Insights is a set of cloud services which provide unified subscription reporting, predictive analysis and remediation of issues through this Satellite instance.'

'In order to utilize these servies, you can set the auto upload in the settings to "ON".'
        
'You can also trigger the upload manually by openning the relevant organization card, and clicking on the "Restart" button'
```

ShimShtein:
Looks good! Merged. Thanks @laviro!

PR #225, opened 2020-06-11
Notify user about inventory upload if a manifest exists and upload is turned off

ShimShtein:


app/controllers/foreman_inventory_upload/uploads_controller.rb
```diff
@@ -31,6 +31,12 @@ def auto_upload
       }
     end
 
+    def show_auto_upload
+      render json: {
+        autoUploadEnabled: Setting[:allow_auto_inventory_upload],
+      }
+    end
```
laviro, line 38: I wonder if we can reuse it, e.g:
```rb
    def auto_upload # was show_auto_upload
￼      render json: {
￼        autoUploadEnabled: Setting[:allow_auto_inventory_upload],
￼      }
￼    end
    
   def set_auto_upload # was auto_upload
      Setting[:allow_auto_inventory_upload] = auto_upload_params
￼      auto_upload
￼    end 
```

webpack/subscriptions_extension_index.js
```diff
@@ -0,0 +1,8 @@
+import { subscriptionCountListener } from './ForemanInventoryUpload/SubscriptionsPageExtension/SubscriptionsPageExtensionActions';
+
+if (window.location.pathname === '/subscriptions') {
+  window.tfm.store.observeStore(
```
laviro, line 4: since this is being called each time the subscription page loads,
I think it worth to call the component here as  `subscriptionCountListener()`
and connect the component to redux.

can we also check here if the `auto_upload` setting is turned `off`? otherwise we don't need to load the function/component

webpack/ForemanInventoryUpload/SubscriptionsPageExtension/SubscriptionsPageExtensionActions.js
```diff
@@ -0,0 +1,44 @@
...
+
+export const subscriptionCountListener = (items, unsubscribe) => {
+  if (items === undefined || items === 0) {
+    return;
+  }
+
+  fetchInventoryAutoUploadSetting(showToast, unsubscribe);
+};
+
+const showToast = (autoUploadEnabled, unsubscribe) => {
```
laviro, line 14: Can you move `showToast` to the scope of `subscriptionCountListener`?
then you don't need to pass `unsubscribe` as a parameter because it's already exist in the scope.

webpack/ForemanInventoryUpload/SubscriptionsPageExtension/SubscriptionsPageExtensionActions.js
```diff
@@ -0,0 +1,44 @@
...
+  if (autoUploadEnabled === undefined || autoUploadEnabled) {
+    return;
+  }
+
+  window.tfm.toastNotifications.notify({
```
laviro, line 19:
```suggestion
  autoUploadEnabled && window.tfm.toastNotifications.notify(...)
```

webpack/ForemanInventoryUpload/SubscriptionsPageExtension/SubscriptionsPageExtensionActions.js
```diff
@@ -0,0 +1,44 @@
...
+    },
+  });
+
+  unsubscribe();
+};
+
+const fetchInventoryAutoUploadSetting = async (
+  processApiResponse,
+  unsubscribe
+) => {
```
laviro, line 34: you can pass here one `callback` param which will be called in the `try` section.
it will let the consumer pass whatever they want.

example of usage:
```js
`fetchInventoryAutoUploadSetting(showToast)` 
//which already have `unsubscribe` if you init it inside `subscriptionCountListener`
```
this method can be changed a bit into:
```js
const fetchInventoryAutoUploadSetting = async callback => {
  try {
    const {
      data: { autoUploadEnabled },
    } = await API.get(inventoryUrl('auto_upload'));
    callback(autoUploadEnabled);
  } catch(error) {
    console.error(error)
  }
};
```

ShimShtein: I want to call the callback even if the API errored. This way I make sure unsubscribe is called, and I don't leave a dangling event listener.

ShimShtein:
@laviro, I think I addressed all your concerns.

PR #233, opened 2020-06-25
[CP 24] Add host obfuscation switcher (#222)

laviro:


ShimShtein:
Merged, thanks @laviro

ShimShtein:
Now we need 1.20 too

PR #249, opened 2020-07-01
Remove ip_addresses and macs when empty

ShimShtein:


lib/foreman_inventory_upload/generators/slice.rb
```diff
@@ -161,6 +153,30 @@ def report_satellite_facts(host)
...
+        macs = host.interfaces.map do |nic|
+          @stream.stringify_value(nic.mac) if nic.mac
+        end.compact
+
+        return if macs.empty?
+
+        @stream.array_field('mac_addresses') do
+          @stream.raw(macs.join(', '))
+        end
+      end
```
adamruzicka, line 178: How about something like this?
```ruby
def report_many_if_present(coll, key)
  results = coll.map do |element|
    value = yield element
    @stream.stringify_value(value) if value
  end.compact

  return if results.empty?

  @stream.array_field(key) do
    @stream.raw(results.join(', '))
  end
end
```

and then calling it like

```ruby
report_many_if_present(host.interfaces, 'ip_addresses')  { |nic| nic.ip }
report_many_if_present(host.interfaces, 'mac_addresses') { |nic| nic.mac }
```

laviro:
@ShimShtein test error is related

PR #254, opened 2020-07-07
Fixes #TBFL - Work around SELinux issues when uploading reports

adamruzicka:
Executing the script directly causes SELinux issues because the dynflow workers running in `foreman_rails_t` are not allowed to execute things of type `foreman_lib_t`. By wrapping the execution in a shell we work around this we effectively work around the original issue because `foreman_rails_t` is apparently allowed to execute `bin_t`.

### DIFF
--- a/lib/foreman_inventory_upload/async/upload_report_job.rb
+++ b/lib/foreman_inventory_upload/async/upload_report_job.rb
@@ -21,7 +21,7 @@ module ForemanInventoryUpload
       end
 
       def command
-        ['/usr/bin/sh', File.join(File.dirname(@filename), ForemanInventoryUpload.upload_script_file)]
+        ['/bin/bash', File.join(File.dirname(@filename), ForemanInventoryUpload.upload_script_file)]
       end
 
       def env
--- a/lib/foreman_inventory_upload/async/upload_report_job.rb
+++ b/lib/foreman_inventory_upload/async/upload_report_job.rb
@@ -21,7 +21,7 @@ module ForemanInventoryUpload
       end
 
       def command
-        File.join(File.dirname(@filename), ForemanInventoryUpload.upload_script_file)
+        ['/usr/bin/sh', File.join(File.dirname(@filename), ForemanInventoryUpload.upload_script_file)]
       end
 
       def env
--- a/lib/foreman_inventory_upload/async/shell_process.rb
+++ b/lib/foreman_inventory_upload/async/shell_process.rb
@@ -9,7 +9,7 @@ module ForemanInventoryUpload
         klass_name = self.class.name
         logger.debug("Starting #{klass_name} with label #{instance_label}")
         progress_output = ProgressOutput.register(instance_label)
-        Open3.popen2e(hash_to_s(env), command) do |_stdin, stdout_stderr, wait_thread|
+        Open3.popen2e(hash_to_s(env), *preprocess_command(command)) do |_stdin, stdout_stderr, wait_thread|
           progress_output.status = "Running in pid #{wait_thread.pid}"
 
           stdout_stderr.each do |out_line|
@@ -32,6 +32,12 @@ module ForemanInventoryUpload
 
       def logger
         Foreman::Logging.logger('background')
+      end
+
+      private
+
+      def preprocess_command(command)
+        command.kind_of?(Array) ? command : [command]
       end
     end
   end
--- a/lib/foreman_inventory_upload/scripts/uploader.sh.erb
+++ b/lib/foreman_inventory_upload/scripts/uploader.sh.erb
@@ -1,4 +1,4 @@
-#! /bin/sh
+#! /bin/bash
 
 DEST=<%= @upload_url %>
 RH_USERNAME=<%= @rh_username %>


evgeni:
Can't we just label that file with `foreman_rails_exec_t`? cc @lzap

evgeni:
Or have an own context for that in `foreman-selinux`

ShimShtein:
@evgeni, you will have to make changes to stable foremans from 1.22. Doesn't sound like a lot of fun.

evgeni:
2.0+, the others are unsupported ;)

But would it be the right thing?

lzap:
Hey, generally we s hould avoid spawning shells if possible. Foreman Rails is currently allowed to spawn, but this is temporary (foreman_hooks + SCL wrapper). What @evgeni said, the better fix is to make sure files this plugin needs to spawn are properly labelled. I can do a PR in foreman-selinux if you tell me the directory which holds these scripts.

ShimShtein:
@lzap, in production setup it will be `/var/lib/foreman/red_hat_inventory`. The code is here: https://github.com/theforeman/foreman_rh_cloud/blob/master/lib/foreman_inventory_upload.rb#L7

lzap:
https://github.com/theforeman/foreman-selinux/pull/108

lib/foreman_inventory_upload/async/upload_report_job.rb
```diff
@@ -21,7 +21,7 @@ def perform(filename, organization_id)
       end
 
       def command
-        File.join(File.dirname(@filename), ForemanInventoryUpload.upload_script_file)
+        ['/usr/bin/sh', File.join(File.dirname(@filename), ForemanInventoryUpload.upload_script_file)]
```
ekohl, line 24: Not sure if it's relevant, but on Debian:
```console
$ ls /usr/bin/sh
ls: cannot access '/usr/bin/sh': No such file or directory
```

Traditionally this was always `/bin/sh`.

evgeni: This plugin depends on Katello, so not too relevant. But /bin/sh is indeed clearer

ekohl:
If the file is temporary, shouldn't it be written to `Rails.root.join('tmp')` instead? `/var/lib/foreman` is backed up and should persist. By default in production it's `/var/run/foreman/tmp` (through symlinks) but works in dev mode. Typically `/var/run` is a symlink to `/run`.

One thing to consider is having a static script with parameters. As a reference for disconnected setups you can generate a script that calls the static script with the right parameters.

Can you link the source of the actual upload script? That could give some suggestion on whether it's actually feasible.

ShimShtein:
@ekohl, the report is persistent - we want to make sure it's there even after restart. We want the user to be able to download the report after a while.

ShimShtein:
> Can you link the source of the actual upload script? That could give some suggestion on whether it's actually feasible.

https://github.com/theforeman/foreman_rh_cloud/blob/master/lib/foreman_inventory_upload/scripts/uploader.sh.erb

@ekohl, Sorry, missed this part from your original response

ekohl:
> @ekohl, the report is persistent - we want to make sure it's there even after restart. We want the user to be able to download the report after a while.

For what it's worth - this is not something we support in Foreman because we don't mandate shared storage and we know users do have load balanced setups. This is why we typically store things in the database. If we do want to support this, we should probably look into [ActiveStorage](https://edgeguides.rubyonrails.org/active_storage_overview.html).

These can also be useful when you want to avoid problems like uploading all files in a single directory. If you want to know what I'm talking about, try creating millions of files in a single directory. At some point the directory index can be multiple megabytes. At that point tools like `ls` will grind down to a halt. http://www.be-n.com/spw/you-can-list-a-million-files-in-a-directory-but-not-with-ls.html is one example, but I've seen the exact same thing in production at $previous_company by naive upload/caching code. This is why you see this:
```console
# ls /run/foreman/cache/
017  0F0  17C  1FB  40E  656  7E9  82D  857  890  8B8  8FC  965  9CD  A59  A88  AAC  B2A  B33  B5C  BA3  BF7  C68  D1F  DE6  E6D  EDD  F3B  F5F
029  176  1ED  32C  568  719  802  84C  870  8A9  8BB  913  9AA  A32  A71  AAB  B21  B31  B34  BA2  BBD  C07  CE0  D80  E31  EAC  F05  F4B
```

This is to avoid placing all cache files in a single directory.

ShimShtein:
I wanted all pieces in place, then we can start complicating the setup. 

In this case the amount of files is limited (it's a file per org). 

As for the fact I am using files - I wanted to create a code that will run both for connected and disconnected setups - hence the creation of the script and feeding a payload to it. Theoretically for connected use cases I can use direct `Net::HTTP` connection with a temp file as a payload, but then I will have a completely different code path for disconnected use cases. I want to make as little assumptions as possible about things that exist on a machine running the script in disconnected environment.

Anyway, returning to our current problem: is this PR a valid workaround? I want to have a working version for the users, and we can fix it properly as a follow up in the next version.

ekohl:
> In this case the amount of files is limited (it's a file per org).

This suggests every report is a new file.
https://github.com/theforeman/foreman_rh_cloud/blob/452678670badf5f48db21bd96f41db20f725e4eb/lib/foreman_inventory_upload.rb#L12-L23
Tools like ActiveStorage provide the utilities to also clean these files when the database entry for it is removed. I'm sensitive about this because I've seen production setups grind to a halt due to these things. You don't notice them until the setup has been running for weeks or months so you want to design this right. Ideally using existing libraries which are battle tested.

> As for the fact I am using files - I wanted to create a code that will run both for connected and disconnected setups - hence the creation of the script and feeding a payload to it. Theoretically for connected use cases I can use direct `Net::HTTP` connection with a temp file as a payload, but then I will have a completely different code path for disconnected use cases. I want to make as little assumptions as possible about things that exist on a machine running the script in disconnected environment.

Long term I'd suggest to have a static `/usr/libexec/foreman/rh-cloud-upload.sh` and pass params via `$1` and `$2`. For convenience you can generate an `upload.sh` from the following template.

```erb
#!/bin/sh
/usr/libexec/foreman/rh-cloud-upload.sh "<%= @upload_url %>" "<%= @rh_username %>"
```

> Anyway, returning to our current problem: is this PR a valid workaround? I want to have a working version for the users, and we can fix it properly as a follow up in the next version.

Probably a good workaround for the short term. I think I'd prefer `/bin/sh` because that's classically the been the location. I'm not sure if `${BASH_SOURCE:-0}` works in POSIX shell. If it doesn't, you should call `/bin/bash` explicitly (and change the shebang in the template).

ShimShtein:
Thanks a lot for the review! I will create a card to refactor this part to something more resilient.

@adamruzicka, can you please change the shebang accordingly, so we will be able to merge it?

adamruzicka:
I changed the shebang, can we get this in?

ShimShtein:
Merged, thanks to everyone!
Tested it on a prod-like machine and everything works.

ekohl:
What's `#TBFL` btw?

adamruzicka:
to be filled later. looks that somehow slipped by

PR #262, opened 2020-07-12
Filter out hits for non-existing hosts

ShimShtein:


laviro:
Thanks @ShimShtein ! merged

PR #276, opened 2020-07-19
Add Inventory sync status to hosts

laviro:
By syncing, you'll be able to know which hosts were uploaded successfully to the Red Hat cloud as your inventory.

## When you don't have a token:
Red Hat Inventory (1)

## When you have token but no subscribed host:
Red Hat Inventory

## When you have token and at least one subscribed host:
Red Hat Inventory (2)

PR #277, opened 2020-07-26
Add scheduled sync for insights hits 

ShimShtein:


webpack/InsightsCloudSync/Components/InsightsSettings/InsightsSettingsReducer.js
```diff
@@ -0,0 +1,40 @@
...
+    case INSIGHTS_SYNC_SETTING_SET:
+      return state.merge({
+        ...state,
+        insightsSyncEnabled,
+      });
+    case INSIGHTS_SYNC_SETTING_SET_ERROR:
+      return state.merge({
+        ...state,
+        insightsSyncEnabled,
+      });
```
adamruzicka, line 36: All the bodies are the same, how about
```javascript
  switch (action.type) {
    case INSIGHTS_SYNC_SETTINGS_GET_SUCCESS, INSIGHTS_SYNC_SETTINGS_GET_FAILURE, INSIGHTS_SYNC_SETTING_SET, INSIGHTS_SYNC_SETTING_SET_ERROR:
      return state.merge({
        ...state,
        insightsSyncEnabled,
      });
    default:
      return state;
  }
```

Also should the bodies be the same? When an error happens when talking to the server, shouldn't the error be taken out of the payload and put into the state?

webpack/InsightsCloudSync/Components/InsightsSyncSwitcher/InsightsSyncSwitcher.js
```diff
@@ -0,0 +1,68 @@
...
+//       )}
+//     />
+//     <Switch
+//       size="mini"
+//       value={insightsSyncEnabled}
+//       onChange={() => handleToggle(insightsSyncEnabled)}
+//     />
+//   </div>
+// );
+
```
adamruzicka, line 58: Does this need to be around?

webpack/InsightsCloudSync/index.js
```diff
@@ -10,8 +10,5 @@ const mapStateToProps = state => ({});
 // map action dispatchers to props
 const mapDispatchToProps = dispatch => bindActionCreators(actions, dispatch);
 
-// export reducers
-export const reducers = {};
-
```
adamruzicka, line 15: why?

webpack/InsightsCloudSync/Components/InsightsSyncSwitcher/index.js
```diff
@@ -0,0 +1,22 @@
...
+// });
+//
+// // map action dispatchers to props
+// const mapDispatchToProps = dispatch => bindActionCreators(actions, dispatch);
+//
+// // export connected component
+// export default connect(mapStateToProps, mapDispatchToProps)(InsightsSyncSwitcher);
+//
+// export const reducers = { accountsList: reducer };
+export default InsightsSyncSwitcher;
```
adamruzicka, line 22: Are all the comments dead code or will you need them later?

ShimShtein:
Thanks @adamruzicka! Cleaned up the code and thanks for catching the error in the reducer. Changed accordingly.

webpack/InsightsCloudSync/Components/InsightsSettings/InsightsSettings.js
```diff
@@ -0,0 +1,33 @@
+import React, { Component } from 'react';
+import PropTypes from 'prop-types';
+import { translate as __ } from 'foremanReact/common/I18n';
+import InsightsSyncSwitcher from '../InsightsSyncSwitcher';
+import './insightsSettings.scss';
+
+class InsightsSettings extends Component {
+  componentWillMount() {
+    const { getInsightsSyncSettings } = this.props;
+    getInsightsSyncSettings();
```
laviro, line 10: should we pass all of the settings we need as `props` [when mounting the component](https://github.com/theforeman/foreman_rh_cloud/blob/master/app/views/foreman_rh_cloud/react/insights_cloud.html.erb#L5 )

webpack/InsightsCloudSync/Components/InsightsSettings/InsightsSettingsActions.js
```diff
@@ -0,0 +1,60 @@
...
+      },
+    });
+  }
+};
+
+export const setInsightsSyncEnabled = currentInsightsSyncEnabled => async dispatch => {
+  try {
+    const {
+      data: { insightsSyncEnabled },
+    } = await API.post(insightsCloudUrl('settings'), {
```
laviro, line 39: should the link update a specific setting, e.g:

```suggestion
    } = await API.post(insightsCloudUrl('insightSync'), {
```

webpack/InsightsCloudSync/Components/InsightsSettings/InsightsSettingsReducer.js
```diff
@@ -0,0 +1,40 @@
...
+});
+
+export default (state = initialState, action) => {
+  const { payload: { settings: insightsSyncEnabled, error } = {} } = action;
+
+  switch (action.type) {
+    case INSIGHTS_SYNC_SETTINGS_GET_SUCCESS:
+      return state.merge({
+        ...state,
+        insightsSyncEnabled,
```
laviro, line 20: for future additions, we can destruct the settings object here:

```suggestion
       return state.merge({
￼        ...state,
￼        ...settings,
       })
```

webpack/InsightsCloudSync/Components/InsightsSyncSwitcher/InsightsSyncSwitcher.js
```diff
@@ -0,0 +1,52 @@
...
+    // This binding is necessary to make `this` work in the callback
+    this.handleToggle = this.handleToggle.bind(this);
+  }
+
+  handleToggle() {
+    const { insightsSyncEnabled, setInsightsSyncEnabled } = this.props;
+
+    const toggledInsightsSyncEnabled = !insightsSyncEnabled;
+    setInsightsSyncEnabled(toggledInsightsSyncEnabled);
+  }
```
laviro, line 20: arrow function should do the binding work for you:

```suggestion
handleToggle = () => {
    const { insightsSyncEnabled, setInsightsSyncEnabled } = this.props;

    const toggledInsightsSyncEnabled = !insightsSyncEnabled;
    setInsightsSyncEnabled(toggledInsightsSyncEnabled);
  }
```

webpack/InsightsCloudSync/Components/InsightsSyncSwitcher/index.js
```diff
@@ -0,0 +1,3 @@
+import InsightsSyncSwitcher from './InsightsSyncSwitcher';
+
+export default InsightsSyncSwitcher;
```
laviro, line 3:
```suggestion
export { default } from './InsightsSyncSwitcher';
```

webpack/InsightsCloudSync/Components/InsightsSettings/InsightsSettings.js
ShimShtein: This component is not mounted directly. It is mounted from the `InsightsCloudSync` page-component.

webpack/InsightsCloudSync/Components/InsightsSettings/InsightsSettingsActions.js
ShimShtein: The settings component should be responsible for all settings at once - I don't want to create separate calls for each setting.

webpack/InsightsCloudSync/Components/InsightsSettings/InsightsSettingsReducer.js
laviro: agree, i think we can do something like:

```js
    case INSIGHTS_SYNC_SETTINGS_GET_SUCCESS:
    case INSIGHTS_SYNC_SETTING_SET:
      return state.merge({
        ...state,
        insightsSyncEnabled,
      });
    case INSIGHTS_SYNC_SETTINGS_GET_FAILURE:
    case INSIGHTS_SYNC_SETTING_SET_ERROR:
      return state.merge({
        ...state,
        error,
      });
```

adamruzicka: Didn't know case in js falls through, TIL :)

ShimShtein:
@laviro, I think it should be ready for another round of review.

app/controllers/insights_cloud/tasks_controller.rb
```diff
@@ -1,8 +1,28 @@
...
+    private
+
+    def render_setting(node_name, setting)
+      render json: {
+        node_name => Setting[setting],
+      }
+    end
+
+    def settings_params
+      ActiveModel::Type::Boolean.new.cast(params.require(:insightsSyncEnabled))
```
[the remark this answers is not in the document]
laviro: do we still need this section after you added the settings controller?
didn't see any routing to those actions.

webpack/InsightsCloudSync/index.js
```diff
@@ -10,8 +10,5 @@ const mapStateToProps = state => ({});
 // map action dispatchers to props
 const mapDispatchToProps = dispatch => bindActionCreators(actions, dispatch);
 
-// export reducers
-export const reducers = {};
```
laviro, line 14: I guess we can actually keep this, currently it doesn't affect anything,
but in case we will want to add a reducer to it, we can just fill this object.

app/models/setting/rh_cloud.rb
```diff
@@ -4,6 +4,7 @@ def self.default_settings
     return unless super
     [
       set('allow_auto_inventory_upload', N_('Allow automatic upload of the host inventory to the Red Hat cloud'), false),
+      set('allow_insights_sync', N_('Allow recommendations synchronization from Red Hat cloud'), false),
```
laviro, line 7: how about `allow_auto_insights_sync` as the one above this line?

webpack/InsightsCloudSync/Components/InsightsSettings/InsightsSettingsActions.js
```diff
@@ -0,0 +1,60 @@
...
+      },
+    });
+  }
+};
+
+export const setInsightsSyncEnabled = currentInsightsSyncEnabled => async dispatch => {
+  try {
+    const {
+      data: { insightsSyncEnabled },
+    } = await API.patch(insightsCloudUrl('settings'), {
```
laviro, line 39: Interesting, I looked on the settings page, and saw they are using `PUT` there for setting update,
after reading a bit more about REST here: https://blog.fullstacktraining.com/restful-api-design-post-vs-put-vs-patch/

it feels that `PUT` might be a better option here, since it's replacing the setting's value and clearing what was before
though it's not a deal breaker for me :)

webpack/InsightsCloudSync/InsightsCloudSync.js
```diff
@@ -4,12 +4,14 @@ import PropTypes from 'prop-types';
...
 import { Button, Icon } from 'patternfly-react';
 import { INSIGHTS_SYNC_PAGE_TITLE } from './InsightsCloudSyncConstants';
+import InsightsSettings from './Components/InsightsSettings';
 
 const InsightsCloudSync = ({ data: { settingsUrl }, syncInsights }) => {
   document.title = INSIGHTS_SYNC_PAGE_TITLE;
   return (
     <IntlProvider locale={navigator.language}>
       <div className="insights-cloud-sync">
+        <InsightsSettings />
```
laviro, line 14: Can you move it to float right? (same as in the inventory page)
Selection_345

webpack/ForemanInventoryUpload/Components/InventoryFilter/__tests__/__snapshots__/integration.test.js.snap
```diff
@@ -14,6 +14,11 @@ Object {
   ],
   "state": Object {
     "ForemanRhCloud": Object {
+      "InsightsCloudSync": Object {
+        "settings": Object {
+          "insightsSyncEnabled": false,
+        },
+      },
```
laviro, line 21: interesting, why was it updated?
it's under `InventoryFilter` directory.. not sure how it got here :)

webpack/InsightsCloudSync/Components/InsightsSettings/insightsSettings.scss
```diff
@@ -0,0 +1,15 @@
+.inventory_settings {
```
laviro, line 1:
```suggestion
.insights_settings {
```

ShimShtein:
@laviro, ready!

webpack/InsightsCloudSync/Components/InsightsSettings/InsightsSettingsReducer.js
```diff
@@ -0,0 +1,28 @@
...
+});
+
+export default (state = initialState, action) => {
+  const { payload: { settings } = {} } = action;
+
+  switch (action.type) {
+    case INSIGHTS_SYNC_SETTINGS_GET_SUCCESS:
+    case INSIGHTS_SYNC_SETTINGS_GET_FAILURE:
+    case INSIGHTS_SYNC_SETTING_SET:
+    case INSIGHTS_SYNC_SETTING_SET_ERROR:
```
laviro, line 20: as the action passes the [error message in payload](https://github.com/theforeman/foreman_rh_cloud/pull/277/files#diff-ad0bf04f6b972ee2294b94466357f7a8R28) we should handle the error a bit differently.

webpack/InsightsCloudSync/Components/InsightsSettings/InsightsSettingsConstants.js
```diff
@@ -0,0 +1,10 @@
+export const INSIGHTS_SYNC_SETTING_SET = 'INSIGHTS_SYNC_SETTING_SET';
+
+export const INSIGHTS_SYNC_SETTING_SET_ERROR =
```
laviro, line 3:
```suggestion
export const INSIGHTS_SYNC_SETTING_SET_FAILURE =
```

webpack/ForemanInventoryUpload/Components/InventoryFilter/__tests__/__snapshots__/integration.test.js.snap
ShimShtein: From the naming, I assume it actually loads all the components tree. So it loaded the whole InsightsSyncComponent too, since it appears under ForemanRhCloud.

webpack/InsightsCloudSync/Components/InsightsSettings/InsightsSettingsActions.js
ShimShtein: Since the API call is at the `settings` level but the call updates only one setting, I think the proper verb would be patch.
Although Rails doesn't distinguish between the two.

webpack/InsightsCloudSync/Components/InsightsSettings/InsightsSettingsReducer.js
ShimShtein: Error json looks like this:
``` json
"settings": {
  "error": "error message",
},
```
I think it should be destructed correctly by this statement.

laviro: Nice ! Didn't spot the action payload structure :)

webpack/InsightsCloudSync/index.js
ShimShtein: Added the reducer to the export

webpack/InsightsCloudSync/InsightsCloudSync.js
ShimShtein: Done

webpack/InsightsCloudSync/Components/InsightsSettings/insightsSettings.scss
ShimShtein: Done

ShimShtein:
@adamruzicka, can give it a round of review PLZ?

lib/insights_cloud/async/insights_scheduled_sync.rb
```diff
@@ -0,0 +1,25 @@
+require 'rest-client'
```
adamruzicka, line 1: Why?

lib/insights_cloud/async/insights_scheduled_sync.rb
```diff
@@ -0,0 +1,25 @@
...
+          logger.debug(
+            'The scheduled process is disabled due to the "allow_auto_insights_sync"
+            setting being set to false.'
+          )
+          return
+        end
+
+        InsightsFullSync.perform_later()
+      ensure
+        self.class.set(:wait => 24.hours).perform_later
```
adamruzicka, line 17: So if the setting is disabled, this job will keep on firing daily even though it won't really do anything?

lib/insights_cloud/async/insights_scheduled_sync.rb
```diff
@@ -0,0 +1,25 @@
+require 'rest-client'
+
+module InsightsCloud
+  module Async
+    class InsightsScheduledSync < ::ApplicationJob
```
adamruzicka, line 5: Where is the first execution of this scheduled from?

lib/insights_cloud/async/insights_scheduled_sync.rb
ShimShtein: Yes, the job is responsible to reschedule itself. This way I don't need to monitor settings and enable the schedule once the setting changes.

lib/insights_cloud/async/insights_scheduled_sync.rb
ShimShtein: Nice catch, fixed: https://github.com/theforeman/foreman_rh_cloud/pull/277/files#diff-8baf49eae5c74fe89761eae48984cfb3R74

adamruzicka:
The tests are red, but I wouldn't say the failures are related.

ShimShtein:
Retesting...

ShimShtein:
Now it's only coveralls.

adamruzicka:
I'm fine with this, sadly the part of not having a dev env with katello still applies

webpack/InsightsCloudSync/Components/InsightsSettings/__tests__/InsightsSettingsReducer.test.js
```diff
@@ -0,0 +1,56 @@
...
+    action: {
+      type: INSIGHTS_SYNC_SETTINGS_GET_FAILURE,
+      payload: {
+        settings: {
+          error: 'test error',
+        },
+      },
+    },
+  },
+  'should handle INVENTORY_ACCOUNT_STATUS_POLLING_START': {
```
laviro, line 33:
```suggestion
  'should handle INSIGHTS_SYNC_SETTING_SET': {
```

webpack/InsightsCloudSync/Components/InsightsSettings/__tests__/InsightsSettingsSelectors.test.js
```diff
@@ -0,0 +1,21 @@
+import { testSelectorsSnapshotWithFixtures } from '@theforeman/test';
+import {
+  selectSettings,
+  selectInsightsSyncEnabled,
+} from '../InsightsSettingsSelectors';
+import { rhCloudStateWrapper } from '../../../../ForemanRhCloudTestHelpers';
+
+const state = rhCloudStateWrapper({
```
laviro, line 8: travis fails because state helper have changed a bit in #276
can you use `insightsStateWrapper` instead?
```suggestion
const state = insightsStateWrapper({
```

laviro:
Thanks @ShimShtein !! great work :)

PR #278, opened 2020-07-27
Add compatibility proxy as a fallback to proxy selection

ShimShtein:


lib/foreman_inventory_upload/async/upload_report_job.rb
```diff
@@ -37,7 +37,9 @@ def env
       end
 
       def http_proxy_string
-        @http_proxy_string ||= HttpProxy.default_global_content_proxy&.full_url
+        @http_proxy_string ||=
+          HttpProxy.default_global_content_proxy&.full_url ||
```
adamruzicka, line 41: If this is trying to address the http proxy situation on 6.6 (katello 3.12ish), then `HttpProxy.default_global_content_proxy` is not defined and calling it would raise a `NoMethodError`, no?

ShimShtein: The method is implemented a bit differently in foreman 22 compatible version: https://github.com/theforeman/foreman_rh_cloud/blob/foreman_1_22/lib/foreman_inventory_upload/async/upload_report_job.rb#L41 . We will change the method accordingly once we cherry-pick this PR to 22 branch. The idea is to leave the code intact: first check `global_content_proxy`, if it's not there - fall back to `cdn_proxy` setting.

adamruzicka: Make sense, thanks for the explanation

ShimShtein:
@adamruzicka, added also global `http_proxy` setting to the fallback list.

lib/foreman_inventory_upload/async/upload_report_job.rb
```diff
@@ -49,6 +52,31 @@ def rh_credentials
...
+      def cdn_proxy
+        cdn_settings = SETTINGS[:katello][:cdn_proxy] || {}
+
+        return nil unless cdn_settings[:host]
+
+        proxy_uri = URI('')
+
+        original_uri = URI.parse(cdn_settings[:host])
+
+        proxy_uri.scheme = original_uri.scheme || 'http'
```
mirekdlugosz, line 65: @ShimShtein I know this has been merged already, but according to https://github.com/Katello/katello/blob/master/app/lib/katello/util/proxy_uri.rb scheme might be "proxy" or "proxys", which I believe curl won't understand.

See also proxy-related code in older insights plugin: https://github.com/redhataccess/foreman-plugin/commit/48b7583dc7f0372f254e71da067fa146b9c7bb19

ShimShtein: According to [Katello](https://github.com/Katello/katello/blob/675d54134c23e358b500ed5d9f6e603fc033b5ab/app/lib/katello/util/http_proxy.rb) `proxy` and `proxys` is used only internally. I don't think users know about those schemes at all. Since I don't work with HttpClient here, the issue should not affect me here.

PR #303, opened 2020-08-05
Use GH actions for JS tests

laviro:


laviro:
to make the action appear we need first to merge it, currently you can view it in https://github.com/laviro/foreman_rh_cloud/actions
though there is one PR need to make foreman presence optional in foreman-js,
currently the linter fails because of that.

ShimShtein:
In your repo's action, I see `npm lint`requires foreman to be there for running. Can you add another checkout step for it?

laviro:
> In your repo's action, I see npm lintrequires foreman to be there for running. Can you add another checkout step for it?

yeah it's an issue in foreman-js that soon will be fixed!

ShimShtein:
Since we want this action to work for all three branches, I suppose we need to be backwards compatible.

laviro:
yes, we can checkout the right foreman branch in https://github.com/theforeman/foreman_rh_cloud/pull/303/files#diff-ad865c90af2a9176f92fb5167d1c4869R33

```yml
 strategy:
      fail-fast: false
      matrix:
        foreman-core-branch: [develop]
```

ShimShtein:
@laviro, let's split it to two different PRs. In the meantime we can leave ruby testing in Travis.

laviro:
Updated, looks good on https://github.com/laviro/foreman_rh_cloud/actions

PR #305, opened 2020-08-06
fix report system_profile os fields

laviro:
Generated OS values:
```json
"os_release": "7.8",
"os_kernel_version": "3.10.0",
```
Those values before:
```json
"os_release": "Red Hat Enterprise Linux Server 7.8 (Maipo)",
"os_kernel_version": "3.10.0-1127.el7.x86_64",
```

background:

Some application expect those values to be "clean",
see the example of other reporting systems, the middle one was reported by `foreman_rh_cloud`:

image

### DIFF
--- a/lib/foreman_inventory_upload/generators/slice.rb
+++ b/lib/foreman_inventory_upload/generators/slice.rb
@@ -122,15 +122,8 @@ module ForemanInventoryUpload
             end.join(', '))
           end
         end
-        @stream.simple_field(
-          'os_release',
-          os_release_value(
-            name: fact_value(host, 'distribution::name'),
-            version: fact_value(host, 'distribution::version'),
-            codename: fact_value(host, 'distribution::id')
-          )
-        )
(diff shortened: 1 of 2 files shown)


ShimShtein:
Let's see, maybe we should still report the full version at least as a fact/tag. I am still waiting for answers on that one.

laviro:
@ShimShtein should we merge this for now and change if we will get another reject?

beav:
Is anything else needed for this PR?

ShimShtein:
@beav, Since we have a problem to distinguish between RHEL and non-RHEL systems, I want to transform those fields on the cloud (yupana) level. This way Yupana will be able to tag hosts with OS type and then transform the fields to the minimalistic form.

ShimShtein:
Just for the clarity: there is no other way for yupana to know which operating system is used for the host, unless we report it in `os_release`.

beav:
ah i see. thanks!

ShimShtein:
As we decided to transform those values in Yupana, we can close this PR in the meantime.

PR #310, opened 2020-08-13
Extend host details page with insights

laviro:
sat-6-6-qa-rhel7 apex example com


laviro:
@ShimShtein it's ready for review :)

app/views/hosts/_insights_tab.html.erb
```diff
@@ -0,0 +1,15 @@
...
+
+<% content_for(:stylesheets) do %>
+    <%= webpacked_plugins_css_for :'foreman_rh_cloud' %>
+<% end %>
+
+<div id="host_details_insights_react_container"></div>
+<%= mount_react_component(
+  'InsightsHostDetailsTab',
+  '#host_details_insights_react_container',
+  { hostname: host.name }.to_json,
```
ShimShtein, line 13: You can pass `host.id` here, and then the route will be a bit more simple.

webpack/InsightsHostDetailsTab/InsightsTab.scss
```diff
@@ -0,0 +1,86 @@
...
+      }
+    }
+
+    &::-webkit-scrollbar {
+      width: 12px;
+      height: 12px;
+    }
+
+    &::-webkit-scrollbar-thumb {
+      background: #0e0e0e6e;
```
ShimShtein, line 80: Do we have to change scrollbar background here? It will require theming downstream.

webpack/InsightsHostDetailsTab/InsightsTabActions.js
```diff
@@ -0,0 +1,30 @@
...
+  INSIGHTS_HITS_REQUEST,
+  INSIGHTS_HITS_SUCCESS,
+  INSIGHTS_HITS_FAILURE,
+} from './InsightsTabConstants';
+
+export const fetchHits = hostname => async dispatch => {
+  try {
+    dispatch({
+      type: INSIGHTS_HITS_REQUEST,
+      payload: {},
```
ShimShtein, line 13: I am not sure why do we need to emit the request, but if we already do it, I suppose it should have `hostname` (or `hostId`) parameter as the payload.

ShimShtein (changes requested):

webpack/InsightsHostDetailsTab/InsightsTabActions.js
laviro: in a follow up PR it will be used to handle better pending state etc..
for now I just created its structure

webpack/InsightsHostDetailsTab/InsightsTab.scss
laviro: it's only the insights tab list scrollbar,
IMO it looks much better than the default one :rofl:

app/views/hosts/_insights_tab.html.erb
laviro: how would the route be simpler? something like:
`get 'hits/:id', to: 'hits#show', constraints: { id: %r{[^\/]+} }`
instead of:
`get 'hits/:host_name', to: 'hits#show', constraints: { host_name: %r{[^\/]+} }`

ShimShtein (changes requested):

laviro:
Thanks @ShimShtein for the review, updated !

app/controllers/insights_cloud/hits_controller.rb
```diff
@@ -0,0 +1,17 @@
+module InsightsCloud
+  class HitsController < ::ApplicationController
+    def index
+      host = Host.where(id: hits_params).first
```
ShimShtein, line 4:
```suggestion
      host = Host.find(host_id_param).first
```

`find` will throw a meaningful exception here

app/controllers/insights_cloud/hits_controller.rb
```diff
@@ -0,0 +1,17 @@
...
+
+      render json: {
+        hits: host.insights.hits,
+      }, status: :ok
+    end
+
+    private
+
+    def hits_params
```
ShimShtein, line 13:
```suggestion
    def host_id_param
```
Just to be a bit more clear with the value.

ShimShtein (changes requested):

ShimShtein (approved):

ShimShtein:
Merged, thanks @laviro!

PR #311, opened 2020-08-13
Filter hosts with only RedHat os

ShimShtein:


lib/foreman_rh_cloud/engine.rb
```diff
@@ -69,6 +69,13 @@ class Engine < ::Rails::Engine
       end
 
       ::Katello::UINotifications::Subscriptions::ManifestImportSuccess.include ForemanInventoryUpload::Notifications::ManifestImportSuccessNotificationOverride if defined?(Katello)
+
+      Host::Base.has_many(
+        :inventory_upload_facts,
```
laviro, line 74: not sure it was actually running on my end,
I got the error: 
`Association named 'inventory_upload_facts' was not found on Host::Managed; perhaps you misspelled it?`
see more: https://pastebin.com/u4hDSGXs

laviro (approved):
Nice models `include` magic !
works as expected :+1:

PR #312, opened 2020-08-18
stop upload if Satellite setting is disconnected

laviro:


lib/foreman_inventory_upload/async/shell_process.rb
```diff
@@ -8,24 +8,30 @@ class ShellProcess < ::ApplicationJob
       def perform(instance_label)
         klass_name = self.class.name
         logger.debug("Starting #{klass_name} with label #{instance_label}")
-        progress_output = ProgressOutput.register(instance_label)
-        Open3.popen2e(hash_to_s(env), *preprocess_command(command)) do |_stdin, stdout_stderr, wait_thread|
-          progress_output.status = "Running in pid #{wait_thread.pid}"
+        progress_output_for(instance_label) do |progress_output|
+          Open3.popen2e(hash_to_s(env), *preprocess_command(command)) do |stdin, stdout_stderr, wait_thread|
```
ShimShtein, line 12:
```suggestion
          Open3.popen2e(hash_to_s(env), *preprocess_command(command)) do |_stdin, stdout_stderr, wait_thread|
```
we do not use `stdin` parameter.

ShimShtein (changes requested):

lib/foreman_inventory_upload/async/shell_process.rb
laviro: interesting why `rubocop` didn't catch it either

laviro:
Thanks @ShimShtein, updated

lib/foreman_inventory_upload/async/upload_report_job.rb
```diff
@@ -8,6 +8,14 @@ def self.output_label(label)
       end
 
       def perform(filename, organization_id)
+        label = UploadReportJob.output_label(organization_id)
+        if Setting[:content_disconnected]
+          progress_output_for(label) do |progress_output|
+            progress_output.write_line('Upload was stopped since this disconnected setting is enabled')
```
ShimShtein, line 14:
```suggestion
            progress_output.write_line('Upload was stopped since disconnected mode setting is enabled for content on this instance.')
```

ShimShtein (changes requested):

lib/foreman_inventory_upload/async/upload_report_job.rb
laviro: Fixed, thanks it sounds better :+1:

PR #314, opened 2020-08-19
Add plugin version to report metadata

ShimShtein:


laviro (approved):
Tested and it works well !
whoever will look on this metadata should understand the way we do versioning, e.g:
- version < `1.0.0` supports foreman `1.22`
- `1.0.0` < version < `2.0.0` supports foreman `1.24`
- `2.0.0` < version supports foreman above `1.24`

PR #325, opened 2020-08-31
Add branch_info endpoint

xprazak2:
Moving an endpoint from redhat-access plugin.
branch_info provides insights-client
with information about a consumer from
Candlepin and Katello.

### DIFF
--- a/app/models/setting/rh_cloud.rb
+++ b/app/models/setting/rh_cloud.rb
@@ -7,6 +7,7 @@ class Setting::RhCloud < Setting
       set('allow_auto_insights_sync', N_('Allow recommendations synchronization from Red Hat cloud'), false),
       set('obfuscate_inventory_hostnames', N_('Obfuscate host names sent to Red Hat cloud'), false),
       set('rh_cloud_token', N_('Authentication token to Red Hat cloud services. Used to authenticate requests to cloud APIs'), 'DEFAULT', N_('Red Hat Cloud token'), nil, encrypted: true),
+      set('include_parameter_tags', N_('Should import include parameter tags from Foreman?'), false),
     ]
   end
 
--- a/app/services/foreman_rh_cloud/branch_info.rb
+++ b/app/services/foreman_rh_cloud/branch_info.rb
@@ -1,0 +1,59 @@
+module ForemanRhCloud
+  class BranchInfo
+    def generate(uuid, host, branch_id, request_hostname)
+      {
+        :remote_leaf => uuid,
+        :remote_branch => branch_id,
+        :display_name => host.organization.name,
+        :hostname => request_hostname,
+        :product => {
+          :type => core_app_name,
+          :major_version => core_app_version.major,
+          :minor_version => core_app_version.minor,
+        },
+        :organization_id => host.organization.id,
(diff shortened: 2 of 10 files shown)


app/services/foreman_rh_cloud/branch_info.rb
```diff
@@ -0,0 +1,56 @@
...
+          :minor_version => core_app_version.minor,
+        },
+        :organization_id => host.organization.id,
+        :satellite_instance_id => Foreman.instance_id,
+        :labels => host_labels(host)
+      }
+    end
+
+    def core_app_name
+      return 'Satellite' if defined? ForemanThemeSatellite
```
xprazak2, line 21: Should this be handled by theme plugin instead?

app/models/setting/rh_cloud.rb
```diff
@@ -7,6 +7,7 @@ def self.default_settings
       set('allow_auto_insights_sync', N_('Allow recommendations synchronization from Red Hat cloud'), false),
       set('obfuscate_inventory_hostnames', N_('Obfuscate host names sent to Red Hat cloud'), false),
       set('rh_cloud_token', N_('Authentication token to Red Hat cloud services. Used to authenticate requests to cloud APIs'), 'DEFAULT', N_('Red Hat Cloud token'), nil, encrypted: true),
+      set('include_parameter_tags', N_('Should import include parameter tags from Foreman?'), false),
```
xprazak2, line 10: Adding this as a boolean setting, not that the [original](https://github.com/redhataccess/foreman-plugin/blob/master/redhat-access/app/controllers/redhat_access/api/machine_telemetry_api_controller.rb#L175) could be `true`, `false`, or array of values to filter by, which is not compatible with our setting types at all. Do we need that granularity? Not sure how many people were using it since there was not even an example for that.

app/services/foreman_rh_cloud/branch_info.rb
ShimShtein: Is it a customer facing string? The theme will brand all translated (and customer facing strings should be translated) strings. If it's saved in the DB or reported somewhere else, by default the theme will not handle this change.
If you still want to brand this string for downstream, we will need to add override code in the theme.

app/services/foreman_rh_cloud/branch_info.rb
```diff
@@ -0,0 +1,59 @@
...
+      'Foreman'
+    end
+
+    def core_app_version
+      return Foreman::Version.new(ForemanThemeSatellite::SATELLITE_VERSION) if defined? ForemanThemeSatellite
```
ShimShtein, line 26: I suppose this one should be done in the theme too.

ShimShtein (changes requested):
@iphands, @rexwhite, Do we still need this, assuming we have `foreman_rh_cloud` reporting to the cloud?

xprazak2:
I am not entirely sure if Insights display those strings anywhere. In any case, I no longer handle them here and I added overrides for the theme plugin.

ShimShtein:
@xprazak2, since #412 is already in, can this PR be closed?

xprazak2:
Yes, closing.

PR #326, opened 2020-09-02
Fix typo in hosts sync toast

mirekdlugosz:
subscriprion -> subscription

Rephrased this line of message a little to be more formal. Fact that we are talking about number of hosts is obvious from context - some number will appear right after this message, in the same line.


laviro (approved):
Thanks @mirekdlugosz !

PR #327, opened 2020-09-02
Run spell-checker over JavaScript files for each PR

mirekdlugosz:
In the past, it did happen that typo in user-facing string slipped in and made it into released version. I thought it might be useful to prevent this from happening again and include spell-checker in development pipeline.

This PR adds [eslint-plugin-spellcheck](https://github.com/aotaduy/eslint-plugin-spellcheck) module, which does exactly that. I added new entries to eslint configuration and new `npm run` command, as well as included this command in github actions workflow.

New `npm run` is required, because `tfm-lint` ignores existing eslintrc file and provides no way to change this behavior during runtime.

Second commit fixes one spurious spelling mistake found by running `npm run lint:spelling`. ~~Spell checker still complains about "subscriprion" - this is addressed in #326 . That's why github action "test_js" fail now. I can close the other PR and include both typo fixes in one commit here.~~ Since #326 is merged and this is rebased, github pipeline passes without error.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -1,7 +1,36 @@
 {
-  "plugins": ["@theforeman/foreman"],
+  "plugins": [
+    "@theforeman/foreman",
+    "spellcheck"
+  ],
   "extends": [
     "plugin:@theforeman/foreman/core",
     "plugin:@theforeman/foreman/plugins"
-  ]
+  ],
+  "rules": {
+    "spellcheck/spell-checker": [
+      "error",
+      {
+        "identifiers": false,
+        "ignoreRequire": true,
+        "skipWords": [
+          "ansible",
+          "donut",
+          "jed",
+          "katello",
+          "knowledgebase",
+          "noopener",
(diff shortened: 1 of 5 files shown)


laviro (commented):
Awesome work @mirekdlugosz ! can you rebase?
cc @amirfefer @sharvit what are your thoughts about adding this also to `foreman-js eslint` ?

mirekdlugosz:
@laviro rebased and force-pushed. Github actions "test_js" now passes :)

laviro (approved):

sharvit:
> Awesome work @mirekdlugosz ! can you rebase?
> cc @amirfefer @sharvit what are your thoughts about adding this also to `foreman-js eslint` ?

Interesting...

IMO it depends on how well it works and if we are going to get `false errors` or `false positives`.
Also, how many errors it is going to find in foreman and in plugins, and how bad is it going to be to fix them.

I would love to see a POC 👍

PR #328, opened 2020-09-03
Add rake tasks for data synchronization from cloud

mirekdlugosz:
We already have rake tasks for uploading report to the cloud, but now plugin also fetches data from the cloud to Satellite. These actions could be triggered only from UI or by scheduled task. Add rake tasks for them, so they can be run on demand from shell.

I expect this to be useful during my testing and in QE automation.

I tested this by uploading new file into Satellite server. `foreman-rake` reports new tasks:

```
$ foreman-rake -AT | grep inventory_upload
rake foreman_inventory_upload:report:generate                                        # Generate inventory report to be sent to Red Hat cloud
rake foreman_inventory_upload:sync:insights                                          # Synchronize Insights hosts hits
rake foreman_inventory_upload:sync:inventory                                         # Synchronize Insights inventory
```

They can be run by `foreman-rake` and in `foreman-tail` I see mentions that tasks were scheduled and completed.


lib/tasks/sync.rake
```diff
@@ -0,0 +1,23 @@
...
+        organizations = Organization.all
+      end
+
+      organizations.each do |organization|
+        InventorySync::Async::InventoryFullSync.perform_now(organization)
+        puts "Synchronized inventory for organization '#{organization.name}'"
+      end
+    end
+  end
+end
```
laviro, line 23: great addition!
I wonder if we should place the tasks under `insights.rake` and `inventory.rake` in case those will grow in the future, cc @ShimShtein

lib/tasks/sync.rake
```diff
@@ -0,0 +1,23 @@
+namespace :foreman_inventory_upload do
```
laviro, line 1:
```suggestion
namespace :foreman_rh_cloud do
```

mirekdlugosz: Changed.
I used `foreman_inventory_upload` to keep all tasks close together on `foreman-rake -T`, so they are easier to discover. After this change, they are still close enough, and if user greps `foreman-rake -T` for "cloud", they will find all of them.

Thanks!

ShimShtein (changes requested):
I like the idea of being able to run those tasks from command line. It's nice addition. 

One thing though: making it a rake task will make the tasks being available only to Foreman admins. Is this the intention, or do you expect regular users being able to run those tasks from command line too?

lib/tasks/sync.rake
```diff
@@ -0,0 +1,23 @@
+namespace :foreman_rh_cloud do
+  namespace :sync do
+    desc "Synchronize Insights hosts hits"
+    task insights: :environment do
```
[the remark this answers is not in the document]
mirekdlugosz: Done! Did I get that right, as of da7ff78f98c61b5ea2e462a4f1189b3036eed452 file names and tasks names are what you expected?

ShimShtein (changes requested):
The plugin uses the task to generate the report, so you need to change the invocation in the code too. It's invoked here: https://github.com/theforeman/foreman_rh_cloud/blob/e06b521add9fdab069973406826343be9810dd38/lib/foreman_inventory_upload/async/generate_report_job.rb#L26.
Can you please change it too?

mirekdlugosz:
@ShimShtein ah, right. Sorry. Fixed in 0f3435257db6c75e0b24b88eb5780f02514cc474

ShimShtein:
Merged, thanks @mirekdlugosz!

PR #329, opened 2020-09-03
[CP_24] - Add eslint spell-checker

laviro:


package.json
```diff
@@ -5,6 +5,7 @@
   "main": "index.js",
   "scripts": {
     "lint": "./node_modules/.bin/eslint -c .eslintrc webpack/",
+    "lint:spelling": "eslint ./",
```
mirekdlugosz, line 8: Since in this branch we still use `eslint`, `npm run lint:spelling` is effectively the same as `npm run lint` (except it might use global `eslint` command, I guess).

This also means `.github/workflows/js_tests.yml` should be back to `run: npm run lint`, there's single command doing both things.

webpack/stories/decorators/withCardsDecorator.js
```diff
@@ -4,7 +4,7 @@ export const withCardsDecorator = storyFn => (
   <div
     style={{
       width: '100%',
-      height: '100vh',
+      height: '100%',
```
mirekdlugosz, line 7: I encountered same problem in my backport. Not sure why spellcheck doesn't like this string, but I solved it by adding "vh" to list of ignored words.

mirekdlugosz (commented):
@laviro I worked on backporting this change myself. I was about to open PR, when I saw yours :D

package.json
laviro: Awesome thanks !

webpack/stories/decorators/withCardsDecorator.js
laviro: yeah, I was about to add it, although on a second thought it means the same in this case :)

laviro:
@mirekdlugosz Updated and JS tests are green :)

mirekdlugosz (approved):
@laviro looks great! Thanks for taking care of backporting this.

Do you want to also backport/cherry-pick into foreman-1.22 branch?

ShimShtein:
Merged, thanks @laviro and @mirekdlugosz! Now for foreman_22 version :)

PR #336, opened 2020-09-14
Run the generation task as admin

ShimShtein:
Hostgroup tag was missing form the report, since the report was not running under a specific user. Now the report will run under administrator permissions.

### DIFF
--- a/test/unit/slice_generator_test.rb
+++ b/test/unit/slice_generator_test.rb
@@ -147,6 +147,10 @@ class ReportGeneratorTest < ActiveSupport::TestCase
   end
 
   test 'generates a report with satellite facts' do
+    hostgroup = FactoryBot.create(:hostgroup)
+    @host.hostgroup = hostgroup
+    @host.save!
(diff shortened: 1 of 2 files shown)


test/unit/slice_generator_test.rb
```diff
@@ -166,6 +170,8 @@ def fact_names
     assert_tag(@host.content_view.name, actual_host, 'content_view')
     assert_tag(@host.location.name, actual_host, 'location')
     assert_tag(@host.organization.name, actual_host, 'organization')
+    assert_tag(@host.hostgroup.name, actual_host, 'hostgroup')
+    puts "checked the hostgroup too with hostgroup: #{@host.hostgroup.name}"
```
adamruzicka, line 174: This feels like a debugging leftover

adamruzicka (commented):
Left one nit, haven't tested but the general approach makes sense to me

laviro (approved):

PR #337, opened 2020-09-15
Reuse 'http_proxy_string' in RestClient requests

laviro:


jlsherrill:
Since you're executing a bash script with an env variable, i'm not 100% sure if this applies, but using a proxy url (i.e.  https://foo:bar@server.com:9090)  will typically work fine as long as the username or password has no special characters in it.  If it does, it generally does not work.  This is really tricky because typically we test with simple passwords and it works fine there.  

When using a library to set the proxy info, its better to explicitly set the username/password to bypass these issues (assuming the library supports it, we have a pretty nasty workaround to make it with Restclient iirc).  However, you're using an env variable with a bash script, something i've not tried, but i would highly suggest testing this to confirm if it works.  This link suggests there may be additional processing that needs to be done:  https://www.cyberciti.biz/faq/unix-linux-export-variable-http_proxy-with-special-characters/

ShimShtein (changes requested):

lib/foreman_rh_cloud.rb
```diff
@@ -14,4 +16,75 @@ def self.authentication_url
...
+      uri = URI("#{scheme}://#{proxy_host}:#{proxy_port}")
+      if proxy_config.user
+        uri.user = CGI.escape(proxy_config.user)
+        uri.password = CGI.escape(proxy_config.password)
+      end
+
+      uri.to_s
+    end
+  rescue URI::Error => e
+    logger.warn("cdn_proxy parsing failed: #{e}")
```
ShimShtein, line 82: Should it be `global_foreman_proxy parsing failed`?

lib/foreman_rh_cloud.rb
```diff
@@ -14,4 +16,75 @@ def self.authentication_url
   def self.verify_ssl_method
     @verify_ssl_method ||= ENV['SATELLITE_RH_CLOUD_URL'] ? OpenSSL::SSL::VERIFY_NONE : OpenSSL::SSL::VERIFY_PEER
   end
+
+  def self.http_proxy_string
+    @http_proxy_string ||= ForemanRhCloud.default_global_content_proxy ||
+    ForemanRhCloud.cdn_proxy ||
+    ForemanRhCloud.global_foreman_proxy
+  end
```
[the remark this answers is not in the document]
laviro: Nice ! Although cdn_proxy is a expecting a bit different proxy object, ehile the others expect string

ShimShtein (changes requested):

laviro:
Still in progress, test is still failing due to setup I believe..

lib/foreman_rh_cloud.rb
```diff
@@ -14,4 +16,61 @@ def self.authentication_url
...
+
+  def self.http_proxy_string
+    ForemanRhCloud.get_proxy
+  end
+
+  def self.transformed_http_proxy_string
+    ForemanRhCloud.transform_scheme(ForemanRhCloud.get_proxy)
+  end
+
+  def self.get_proxy
```
ShimShtein, line 28:
```suggestion
  def self.cloud_proxy
```
or maybe
```suggestion
  def self.proxy_setting
```

lib/foreman_rh_cloud.rb
```diff
@@ -14,4 +16,61 @@ def self.authentication_url
...
+      transformed_uri.scheme = 'proxy'
+    when "https"
+      transformed_uri.scheme = 'proxys'
+    end
+
+    transformed_uri.to_s
+  end
+
+  def self.logger
+    Foreman::Logging.logger('background')
```
ShimShtein, line 74: You are in `lib` context, I don't think it's correct to use background logger specifically. I suppose it would be better to pass the logger into each method with a sane default:

``` ruby
  def self.cdn_proxy(is_proxy_scheme = true, logger: Foreman::Logging.logger('rh_cloud'))  
  #...
  rescue URI::Error => e
    logger.warn("cdn_proxy parsing failed: #{e}")
    nil
  end
```

test/unit/rh_cloud_http_proxy_test.rb
```diff
@@ -0,0 +1,48 @@
+require 'test_plugin_helper'
+
+class RhCloudHttpProxyTest < ActiveSupport::TestCase
+  setup do
+  end
+
+  test 'Proxy selected in the correct order' do
+    global_content_proxy_mock = 'http://global:content@localhost:8888'
+    katello_cdn_proxy_mock = 'http://katello:cdn@localhost:8888'
+    global_foreman_proxy_mock = 'http://global:foreman@localhost:8888'
```
ShimShtein, line 10: Move to setup block
```suggestion
    @global_content_proxy_mock = 'http://global:content@localhost:8888'
    @katello_cdn_proxy_mock = 'http://katello:cdn@localhost:8888'
    @global_foreman_proxy_mock = 'http://global:foreman@localhost:8888'
```

test/unit/rh_cloud_http_proxy_test.rb
```diff
@@ -0,0 +1,48 @@
...
+
+    HttpProxy.default_global_content_proxy.full_url = nil
+
+    # now cdn_proxy should be selected.
+    assert_equal katello_cdn_proxy_mock, ForemanRhCloud.get_proxy
+
+    SETTINGS[:katello][:cdn] = nil
+
+    # finally global_foreman_proxy should be selected.
+    assert_equal global_foreman_proxy_mock, ForemanRhCloud.get_proxy
```
ShimShtein, line 32: Move this to a separate test method.

You want three tests:

``` ruby
test 'selects global content proxy'
  setup_global_content_proxy
  setup_global_foreman_proxy
  setup_cdn_proxy do
    assert_equal @global_content_proxy_mock, ForemanRhCloud.get_proxy
  end
end

test 'selects cdn proxy'
  setup_global_foreman_proxy
  setup_cdn_proxy do
    assert_equal @katello_cdn_proxy_mock, ForemanRhCloud.get_proxy
  end
end

test 'selects cdn proxy'
  setup_global_foreman_proxy

  assert_equal @global_foreman_proxy_mock, ForemanRhCloud.get_proxy
end

def setup_global_content_proxy
  http_proxy = FactoryBot.create(:http_proxy, url: @global_content_proxy_mock
  HttpProxy.stubs(:default_global_content_proxy).returns(http_proxy)
end

def setup_global_foreman_proxy
    FactoryBot.create(:setting, :name => 'http_proxy', :value => global_foreman_proxy_mock)
end

def setup_cdn_proxy
   old_cdn_setting = SETTINGS[:katello][:cdn]
   SETTINGS[:katello][:cdn] = katello_cdn_proxy_mock
   yield
ensure
  SETTINGS[:katello][:cdn] = old_cdn_setting
end
```

test/unit/rh_cloud_http_proxy_test.rb
```diff
@@ -0,0 +1,48 @@
+require 'test_plugin_helper'
+
+class RhCloudHttpProxyTest < ActiveSupport::TestCase
+  setup do
+  end
+
+  test 'Proxy selected in the correct order' do
+    global_content_proxy_mock = 'http://global:content@localhost:8888'
+    katello_cdn_proxy_mock = 'http://katello:cdn@localhost:8888'
```
ShimShtein, line 9: cdn setting is a hash, so the value must be a hash too:

```suggestion
    katello_cdn_proxy_mock = {
      host: 'localhost',
      port: '8888',
      user: 'katello',
      password: 'c@n',
      scheme: 'http'
    }
```

ShimShtein (changes requested):

test/unit/rh_cloud_http_proxy_test.rb
laviro: Wow thanks!! it really helped :)

lib/foreman_rh_cloud.rb
laviro: got the error: `RuntimeError: Trying to use logger rh_cloud which has not been configured.`

ShimShtein: Well then, we can use `app`. I think it looks for one of the options from [here](https://github.com/theforeman/foreman/blob/develop/config/settings.yaml.example#L78).

lib/foreman_rh_cloud.rb
```diff
@@ -14,4 +16,57 @@ def self.authentication_url
...
+    ForemanRhCloud.proxy_setting
+  end
+
+  def self.transformed_http_proxy_string
+    ForemanRhCloud.transform_scheme(ForemanRhCloud.proxy_setting)
+  end
+
+  def self.proxy_setting
+    HttpProxy.default_global_content_proxy&.full_url ||
+    ForemanRhCloud.cdn_proxy ||
```
ShimShtein, line 30: You have to pass the logger all the way to "public" methods, if you want it to be overridden in the task:

```suggestion
  def self.http_proxy_string(logger: Foreman::Logging.logger('background'))
    ForemanRhCloud.proxy_setting(logger: logger)
  end

  def self.transformed_http_proxy_string(logger: Foreman::Logging.logger('background'))
    ForemanRhCloud.transform_scheme(ForemanRhCloud.proxy_setting(logger: logger))
  end

  def self.proxy_setting(logger: Foreman::Logging.logger('background'))
    HttpProxy.default_global_content_proxy&.full_url ||
    ForemanRhCloud.cdn_proxy(logger: logger) ||
```

lib/foreman_inventory_upload/async/upload_report_job.rb
```diff
@@ -47,20 +47,15 @@ def env
           'FILES' => @filename,
           'CER_PATH' => @cer_path
         )
+
+        http_proxy_string = ForemanRhCloud.http_proxy_string
```
ShimShtein, line 51:
```suggestion
        http_proxy_string = ForemanRhCloud.http_proxy_string(logger: logger)
```

lib/insights_cloud/async/insights_full_sync.rb
```diff
@@ -27,6 +27,7 @@ def query_insights_hits
           method: :get,
           url: InsightsCloud.hits_export_url,
           verify_ssl: ForemanRhCloud.verify_ssl_method,
+          proxy: ForemanRhCloud.transformed_http_proxy_string,
```
ShimShtein, line 30:
```suggestion
          proxy: ForemanRhCloud.transformed_http_proxy_string(logger: logger),
```

lib/insights_cloud/async/insights_full_sync.rb
```diff
@@ -40,6 +41,7 @@ def query_refresh_token
           method: :post,
           url: ForemanRhCloud.authentication_url,
           verify_ssl: ForemanRhCloud.verify_ssl_method,
+          proxy: ForemanRhCloud.transformed_http_proxy_string,
```
ShimShtein, line 44:
```suggestion
          proxy: ForemanRhCloud.transformed_http_proxy_string(logger: logger),
```

lib/inventory_sync/async/inventory_full_sync.rb
```diff
@@ -65,6 +65,7 @@ def query_inventory(page = 1)
           method: :get,
           url: ForemanInventoryUpload.inventory_export_url,
           verify_ssl: ForemanRhCloud.verify_ssl_method,
+          proxy: ForemanRhCloud.transformed_http_proxy_string,
```
ShimShtein, line 68:
```suggestion
          proxy: ForemanRhCloud.transformed_http_proxy_string(logger: logger),
```

lib/inventory_sync/async/inventory_full_sync.rb
```diff
@@ -82,6 +83,7 @@ def query_refresh_token
           method: :post,
           url: ForemanRhCloud.authentication_url,
           verify_ssl: ForemanRhCloud.verify_ssl_method,
+          proxy: ForemanRhCloud.transformed_http_proxy_string,
```
ShimShtein, line 86:
```suggestion
          proxy: ForemanRhCloud.transformed_http_proxy_string(logger: logger),
```

test/unit/rh_cloud_http_proxy_test.rb
```diff
@@ -0,0 +1,64 @@
...
+    setup_global_foreman_proxy
+    setup_cdn_proxy do
+      assert_equal @global_content_proxy_mock, ForemanRhCloud.proxy_setting
+    end
+  end
+
+  test 'selects cdn proxy' do
+    setup_global_foreman_proxy
+    setup_cdn_proxy do
+      assert_equal @katello_cdn_proxy_mock, ForemanRhCloud.proxy_setting
```
ShimShtein, line 27: You have to construct the string here. `@katello_cdn_proxy_mock` is a hash object.

ShimShtein (changes requested):
Almost there. Let's change the default logger to `app`and fix the test and we're done.

laviro:
Thanks @ShimShtein ! the select cdn test is still failing with:
```
--- expected
+++ actual
@@ -1 +1 @@
-"http://katello:cdn@localhost:8888"
+"http://global:foreman@localhost:8888"
```

while debugging it in `ForemanRhCloud.cdn_proxy` I get on first test run the correct proxy setting, while on the second run (the failing test) I see it's `nil`

while debugging it in `test > setup_cdn_proxy` I see it is set correctly in both runs..

ShimShtein:
Merged, thanks @laviro!

PR #338, opened 2020-09-15
Disable inventory sync button while it's loading

laviro:
To disable double API calls to fetch host inventory from the cloud.


ShimShtein:
Merged, thanks @laviro!

PR #339, opened 2020-09-15
Disable report generate button while it's loading

laviro:


ShimShtein:
Merged, thanks @laviro!

PR #351, opened 2020-09-29
Add toast with error message for generate report action

ShimShtein:
Adds a toast with error message:

image

### DIFF
--- a/webpack/ForemanInventoryUpload/Components/AccountList/AccountListActions.js
+++ b/webpack/ForemanInventoryUpload/Components/AccountList/AccountListActions.js
@@ -1,4 +1,5 @@
 import API from 'foremanReact/API';
+import { addToast } from 'foremanReact/redux/actions/toasts';
 import { inventoryUrl } from '../../ForemanInventoryHelpers';
 import {
   INVENTORY_ACCOUNT_STATUS_POLLING,
@@ -46,7 +47,7 @@ export const stopAccountStatusPolling = pollingProcessID => dispatch => {
   });
 };
 
-export const restartProcess = (accountID, activeTab) => dispatch => {
+export const restartProcess = (accountID, activeTab) => async dispatch => {
   let processController = null;
   let processStatusName = null;
 
@@ -58,12 +59,22 @@ export const restartProcess = (accountID, activeTab) => dispatch => {
     processStatusName = 'generate_report_status';
   }
 
-  API.post(inventoryUrl(`${accountID}/${processController}`));
-  dispatch({
-    type: INVENTORY_PROCESS_RESTART,
-    payload: {
-      accountID,
-      processStatusName,
-    },
-  });
+  try {
+    await API.post(inventoryUrl(`${accountID}/${processController}`));
+    dispatch({
+      type: INVENTORY_PROCESS_RESTART,
+      payload: {
+        accountID,
+        processStatusName,
+      },
+    });
+  } catch (error) {
+    dispatch(
+      addToast({
+        sticky: true,
+        type: 'error',
+        message: error.message,
+      })
+    );
+  }
 };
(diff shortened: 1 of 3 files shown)


webpack/ForemanInventoryUpload/Components/AccountList/AccountListActions.js
```diff
@@ -58,7 +59,15 @@ export const restartProcess = (accountID, activeTab) => dispatch => {
...
-  API.post(inventoryUrl(`${accountID}/${processController}`));
+  API.post(inventoryUrl(`${accountID}/${processController}`)).catch(error => {
+    dispatch(
+      addToast({
+        sticky: true,
+        type: 'error',
+        message: error.message,
+      })
+    );
+  });
```
laviro, line 70: I think it is more readable inside a `try -> catch` block, 
and it will also prevent the dispatch of `INVENTORY_PROCESS_RESTART` if not needed, e,g:

```suggestion
   try {
    await API.post(inventoryUrl(`${accountID}/${processController}`))
    dispatch({
      type: INVENTORY_PROCESS_RESTART,
      payload: {
        accountID,
        processStatusName,
      },
    });
  } catch(error){
    dispatch(
      addToast({
        sticky: true,
        type: 'error',
        message: error.message,
      })
    );
  }; 
};
```

ShimShtein:
@laviro, I think I am missing an assert here, but at least the tests are passing.

webpack/ForemanInventoryUpload/Components/AccountList/AccountListActions.js
ShimShtein: Since the method is not async, I cannot use `await`. If the dispatch of the restart is needed only on success, I can change the code to 
```suggestion
  API.post(inventoryUrl(`${accountID}/${processController}`)).then(
  () => {
    dispatch({
      type: INVENTORY_PROCESS_RESTART,
      payload: {
        accountID,
        processStatusName,
      },
    });
  }, 
  error => {
    dispatch(
      addToast({
        sticky: true,
        type: 'error',
        message: error.message,
      })
    );
  });
```

laviro: > Since the method is not async, I cannot use `await`

you can add `async` just before passing the `dispatch` here: https://github.com/theforeman/foreman_rh_cloud/pull/351/files#diff-cc6147ee1ab873c90de32fe3b134f1aeR50

e.g:
`export const restartProcess = (accountID, activeTab) => async dispatch => {`

and then it should work

webpack/ForemanInventoryUpload/Components/AccountList/__tests__/AccountListActions.test.js
```diff
@@ -24,4 +24,21 @@ const fixtures = {
...
+  it('should invoke toast notification upon failure', () => {
+    const apiError = new Error('test error');
+
+    API.post.mockImplementation(
+      () =>
+        new Promise((resolve, reject) => {
+          reject(apiError);
+        })
+    );
+  });
```
laviro, line 43: Actually I found a good example in tasks to test also the action failure: https://github.com/theforeman/foreman-tasks/blob/master/webpack/ForemanTasks/Components/TasksTable/__tests__/TasksBulkActions.test.js#L25

you can add another fixture that will check for failure, something like:

```js
const fixtures = {
  ...,
  'should invoke toast notification upon failure': () => {
    API.post.mockImplementation(() =>
      Promise.reject(new Error('Test Error'))
    );
    return restartProcess(accountID, activeTab)
  }
}
```

laviro: There is also [mockfnmockImplementationOnce](https://jestjs.io/docs/en/mock-function-api#mockfnmockimplementationoncefn) which might be better if we also want to test the `success` case

ShimShtein:
@laviro I think it's ready for review.

laviro (commented):
Should we put put the server side part in a `try, catch` block and render the correct json response correctly?

laviro (approved):

PR #350, opened 2020-09-29
Add 'exclude_installed_packages' setting

laviro:

PR #352, opened 2020-09-30
Add UI switcher for exclude packages setting

laviro:
~~blocked by #350~~


webpack/ForemanInventoryUpload/Components/ExcludePackagesSwitcher/ExcludePackagesSwitcher.js
```diff
@@ -0,0 +1,30 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { Switch, FieldLevelHelp } from 'patternfly-react';
+import { translate as __ } from 'foremanReact/common/I18n';
+import './excludePackagesSwitcher.scss';
+
+const ExcludePackagesSwitcher = ({ excludePackages, handleToggle }) => (
```
ShimShtein, line 7: I think it makes sense to refactor all switchers to a common component and reuse the same component with different params in `InventorySettings`.

webpack/ForemanInventoryUpload/Components/ExcludePackagesSwitcher/ExcludePackagesSwitcherActions.js
```diff
@@ -0,0 +1,30 @@
...
+    });
+    dispatch({
+      type: EXCLUDE_PACKAGES_TOGGLE,
+      payload: {
+        excludePackages,
+      },
+    });
+  } catch (error) {
+    dispatch({
+      type: EXCLUDE_PACKAGES_TOGGLE_ERROR,
```
ShimShtein, line 24: Currently there is no code that catches this action. We need one of three choices here:

1. Add reducer code to handle this action, and add a component to handle the state change
2. Publish a toast message instead of a new action
3. Add a generic reducer and a generic component to handle its changes, so all errors could be refactored to emit the same action.

ShimShtein (changes requested):
Since server side is now merged, this PR needs a rebase.

webpack/ForemanInventoryUpload/Components/ExcludePackagesSwitcher/ExcludePackagesSwitcher.js
laviro: Sounds good, how about getting this feature in first
and than do in a follow-up the refactor for the 3 buttons?

webpack/ForemanInventoryUpload/Components/ExcludePackagesSwitcher/ExcludePackagesSwitcherActions.js
laviro: I think a toast would be the best fit, and it could become a generic component :+1:

webpack/ForemanInventoryUpload/Components/ExcludePackagesSwitcher/__tests__/ExcludePackagesSwitcher.test.js
```diff
@@ -0,0 +1,14 @@
+import { testComponentSnapshotsWithFixtures } from '@theforeman/test';
+import { noop } from 'patternfly-react';
+
+import ExcludePackagesSwitcher from '../ExcludePackagesSwitcher';
+
+const fixtures = {
+  'render with props': { excludePackages: true, handleToggle: noop },
+  /** fixtures, props for the component */
```
ShimShtein, line 8: Leftovers?

ShimShtein (changes requested):

webpack/ForemanInventoryUpload/Components/ExcludePackagesSwitcher/__tests__/ExcludePackagesSwitcher.test.js
laviro: Removed, thanks :+1:

laviro:
Added integration test

ShimShtein (approved):

ShimShtein:
Merged, thanks @laviro!

PR #353, opened 2020-10-04
[CP_24] - Disable buttons while loading

laviro:


ShimShtein:
Merged, thanks @laviro!

PR #355, opened 2020-10-04
[CP 22] Add toast with error message for generate report action

ShimShtein:


laviro (approved):

laviro:
Tested and it works as expected,
user with view permissions only can't restart the generating process
and they will see a toast with the error.

PR #358, opened 2020-10-05
WIP - new host details tab extention

laviro:
**This is still a WIP**

### DIFF
--- a/webpack/global_index.js
+++ b/webpack/global_index.js
@@ -1,0 +1,33 @@
+/* eslint-disable import/extensions */
+import React from 'react';
+import { addGlobalFill } from 'foremanReact/components/common/Fill/GlobalFill';
+import { registerReducer } from 'foremanReact/common/MountingService';
+import reducers from './ForemanRhCloudReducers';
+import InsightsHostDetailsTab from './InsightsHostDetailsTab';
+import { subscriptionCountListener } from './ForemanInventoryUpload/SubscriptionsPageExtension/SubscriptionsPageExtensionActions';
+
+// Register reducers
+Object.entries(reducers).forEach(([key, reducer]) =>
+  registerReducer(key, reducer)
+);
+
+const [prefix, mainPath, subPath] = window.location.pathname.split('/');
+switch (mainPath) {
+  case 'subscriptions':
+    window.tfm.store.observeStore(
+      'katello.subscriptions.itemCount',
+      subscriptionCountListener
+    );
+    break;
+  case 'experimental':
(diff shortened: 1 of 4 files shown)


laviro:
still a WIP, but made css to work :fireworks: 

@ShimShtein @sharvit @amirfefer

Screenshot (14)

webpack/global_index.js
```diff
@@ -0,0 +1,49 @@
...
+  document
+    .getElementsByTagName('head')[0]
+    .insertAdjacentHTML('beforeend', linkElement);
+
+  const [prefix, mainPath, subPath] = window.location.pathname.split('/');
+  switch (mainPath) {
+    case 'subscriptions':
+      window.tfm.store.observeStore(
+        'katello.subscriptions.itemCount',
+        subscriptionCountListener
```
amirfefer, line 29: what is the purpose of this?

webpack/global_index.js
```diff
@@ -0,0 +1,49 @@
...
+      );
+      break;
+    case 'experimental':
```
amirfefer, line 32: is this switch case triggering once or on every page transition? 
Scenario:
1. Go to a react based page (i.e Audits) - This switch case catch  nothing
2. Navigate to a hosts details page - This switch case isn't re-triggered
3.  Insights tab isn't registered

If this switch case triggers on every page transition we need to verify that we don't register fills twice.

amirfefer (changes requested):
Thanks @laviro :+1: 
The CSS importing should work as expected so you can remove the workaround. 
I left a few inline comments.

app/controllers/foreman_rh_cloud/react_controller.rb
```diff
@@ -7,5 +7,12 @@ def inventory_upload
 
     def insights_cloud
     end
+
+    def plugin_css_path
+      link_element = helpers.css_tags_for([:foreman_rh_cloud]).first
+      render json: {
+        linkElement: link_element,
+      }, status: :ok
+    end
```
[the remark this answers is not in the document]
laviro: awesome! great job @amirfefer

webpack/global_index.js
laviro: it was brought from the older global_js file

webpack/global_index.js
laviro: need a bit more investigation, though IMO when it comes to React pages it will get registered just once.

adamruzicka:
Needs a rebase, also general ping :)

laviro:
opened a new PR in #653

PR #360, opened 2020-10-06
Extract cloud authentication

xprazak2:
Extract auth logic to a concern,
so it can be easily used in multiple places.


ShimShtein (approved):

ShimShtein:
Awesome! Thanks @xprazak2, care doing cherry picks to `foreman_1_24` and `foreman_1_22` branches?
Usually we want to achieve feature parity between foreman versions, so the plugin will feel the same for all versions.

xprazak2:
Cherry picks:

https://github.com/theforeman/foreman_rh_cloud/pull/367
https://github.com/theforeman/foreman_rh_cloud/pull/368

PR #361, opened 2020-10-06
Add insights sorting to host index page

laviro:


laviro:
Ready for review @ShimShtein

ShimShtein (changes requested):
Two additional issues:

- [ ] Hosts that didn't have insights facet objects do not cache `inisghts_hits` correctly

- [ ] When we sort `desc`, `nil` values appear on top, instead of the bottom.

ShimShtein:
@adamruzicka, if I read correctly `scoped_search` [builder](https://github.com/wvanbergen/scoped_search/blob/master/lib/scoped_search/query_builder.rb#L587) of order by statements, there is no way we can influence `null first`/`null last` definition?

adamruzicka:
@ShimShtein I read it the same way :/

app/overrides/hosts_list.rb
```diff
@@ -2,7 +2,7 @@
   virtual_path: 'hosts/_list',
   name: 'insights_hits_header',
   insert_before: 'thead tr th.hidden-xs:first-of-type',
-  text: '<th class="hidden-xs" width="10%"><%= _("Insights")%></th>'
+  text: '<th class="hidden-xs" width="10%"><%= sort :"insights_recommendations_count", :as => _("Insights")%></th>'
```
ShimShtein, line 5:
```suggestion
  text: '<th class="hidden-xs" width="10%"><%= sort :insights_recommendations_count, :as => _("Insights")%></th>'
```

No need for string here.

ShimShtein (changes requested):
Small nitpick and we're done.

ShimShtein:
Merged, thanks @laviro!

PR #364, opened 2020-10-11
Set cloudToken prop to not required

laviro:
There is this warning in the console:

Selection_349

And also since we can handle when string is empty or null, the prop doesn't have to be required.


ShimShtein:
Merged, thanks @laviro! Now for cherry picks...

PR #365, opened 2020-10-11
omit fields where fact is nil

laviro:


lib/foreman_inventory_upload/generators/json_stream.rb
```diff
@@ -26,8 +26,14 @@ def raw(string)
         @out << string
       end
 
-      def simple_field(name, value, last = false)
-        @out << "\"#{name}\": #{stringify_value(value)}#{last ? '' : ','}" unless value.nil?
+      def simple_field(name, value, last = false, &block)
+        return if value.nil?
+        return if value.kind_of?(String) && value.empty?
+        return if value.kind_of?(Array) && value.compact.empty?
```
ShimShtein, line 32:
```suggestion
        return if value.empty?
        return if value.kind_of?(Array) && value.compact.empty?
```

`empty?` works on simple values.

ShimShtein (changes requested):

lib/foreman_inventory_upload/generators/json_stream.rb
laviro: yes but for numbers it can return an error as there is no `empty?` for `Integer`
in case we will want to pass in an Integer (?)

ShimShtein: You are right, sorry:
``` ruby
return if value.try(:empty?)
```
will do the trick for types that don't have empty, like Integer.

laviro: Thanks, updated !

ShimShtein (changes requested):
Can we have some tests for these cases?
We can either add direct tests for `json_stream` or we can add more test cases in `slice_gnerator_test` for nil and non-nil values.

laviro:
Added tests :)

san7ket:
No validations issues from Yupana or HBI. HBI sets the fields to "None" for some reason though doesn't omit them. maybe these are default/required fields.

ShimShtein:
Merged, thanks @laviro and @san7ket!

PR #375, opened 2020-10-13
Add endpoint to proxy requests

xprazak2:
Adds an endpoint that forwards requests
from insights-client to cloud.


app/controllers/redhat_access/machine_telemetries_controller.rb
```diff
@@ -0,0 +1,71 @@
...
+
+    def assign_header(res, cloud_res, header, transform)
+      header_content = cloud_res.headers[header]
+      return unless header_content
+      new_header = transform ? header.to_s.gsub('_', '-') : header.to_s
+      res.headers[new_header] = header_content
+    end
+
+    def ensure_telemetry_enabled_for_consumer
+      render_message 'Telemetry is not enabled for your organization', :status => 403 unless telemetry_enabled_for_consumer?
```
xprazak2, line 42: This check is always true. I did not find any other place where `TelemetryConfiguration` is used. Unless there is a related feature that we want to have in the future, this looks like a good candidate to drop.

app/services/foreman_rh_cloud/cloud_request_forwarder.rb
```diff
@@ -0,0 +1,75 @@
...
+  class CloudRequestForwarder
+    include ::ForemanRhCloud::CloudAuth
+
+    def forward_request(original_request, controller_name, branch_id)
+      forward_params = prepare_forward_params(original_request, branch_id)
+      Rails.logger.debug{"Request parameters for telemetry request: #{forward_params}"}
+
+      forward_payload = prepare_forward_payload(original_request, controller_name)
+
+      Rails.logger.debug("User agent for telemetry is: #{http_user_agent original_request}")
```
xprazak2, line 13: How useful is logging user agent?

app/models/redhat_access/telemetry_configuration.rb
```diff
@@ -0,0 +1,14 @@
+module RedhatAccess
+  class TelemetryConfiguration < ActiveRecord::Base
```
ShimShtein, line 2: I suppose we can remove this model once we move to certificate authentication on the cloud.
@lphiri do you know about a schedule for this?

app/controllers/redhat_access/machine_telemetries_controller.rb
```diff
@@ -0,0 +1,71 @@
+module RedhatAccess
+  class MachineTelemetriesController < ::Api::V2::BaseController
+    layout false
+
+    include ::RedhatAccess::ClientAuthentication
+    include ::RedhatAccess::CandlepinRelated
+
+    before_action :ensure_telemetry_enabled_for_consumer, :only => [:forward_request]
```
ShimShtein, line 8: I suppose it's not needed - if insights-client decides it want to send data, it means the user has enabled it. I suppose we can pass it to the cloud, and make it cloud's responsibility to reject requests without proper subscriptions.

app/services/foreman_rh_cloud/cloud_request_forwarder.rb
```diff
@@ -0,0 +1,75 @@
+require 'rest-client'
+
+module ForemanRhCloud
+  class CloudRequestForwarder
+    include ::ForemanRhCloud::CloudAuth
```
ShimShtein, line 5: I suppose we should use certificate auth here, similar to what is done in [report upload job](https://github.com/theforeman/foreman_rh_cloud/blob/master/lib/foreman_inventory_upload/async/upload_report_job.rb#L32). We have a plan to get rid of token auth altogether.

app/services/foreman_rh_cloud/cloud_request_forwarder.rb
ShimShtein: It depends what information is sent in the user agent by insihgts-client. From looking at the code in `prepare_forward_params`, the user agent actually modifies the parameters of the outgoing request.

app/services/foreman_rh_cloud/cloud_request_forwarder.rb
```diff
@@ -0,0 +1,75 @@
...
+
+    def forward_request(original_request, controller_name, branch_id)
+      forward_params = prepare_forward_params(original_request, branch_id)
+      Rails.logger.debug{"Request parameters for telemetry request: #{forward_params}"}
```
ShimShtein, line 9: We probably need to use one of the foreman loggers here, with a proper label, like here: https://github.com/theforeman/foreman_rh_cloud/blob/master/lib/foreman_inventory_upload/async/shell_process.rb#L39

app/services/foreman_rh_cloud/cloud_request_forwarder.rb
```diff
@@ -0,0 +1,75 @@
...
+    def prepare_forward_params(original_request, branch_id)
+      forward_params = original_request.query_parameters
+      if original_request.user_agent && !original_request.user_agent.include?('redhat_access_cfme')
+        forward_params = forward_params.merge(:branch_id => branch_id)
+      end
+
+      forward_params
+    end
+
+    def prepare_cloud_url(request_path)
```
ShimShtein, line 56: We can move this method to `lib/insights_cloud.rb`, where all insights-related urls live.

db/migrate/20201006085432_create_redhat_access_telemetry_configuration.rb
```diff
@@ -0,0 +1,20 @@
+class CreateRedhatAccessTelemetryConfiguration < ActiveRecord::Migration[6.0]
```
ShimShtein, line 1: Older foreman versions do not support rails 6.0.
```suggestion
class CreateRedhatAccessTelemetryConfiguration < ActiveRecord::Migration[5.0]
```

ShimShtein, line 1: If we decide to use this model (depends on the auth model used), can you rename the file to `20201006085432_create_redhat_access_telemetry_configuration.foreman_rh_cloud.rb` (added `foreman_rh_cloud` label).

ShimShtein (changes requested):

app/controllers/redhat_access/machine_telemetries_controller.rb
lphiri: The reason we had this in the previous plugin was to give the Satellite admin control over whether the proxy functionality should be allowed.   This useful when the Satellite owner is not the same as the managed host owner. I suggest we keep this as its used in practice.

app/controllers/redhat_access/machine_telemetries_controller.rb
lphiri: This related to the previous comment.  You will need to keep the ability to disable the proxy functionality to maintain feature parity.

db/migrate/20201006085432_create_redhat_access_telemetry_configuration.rb
ShimShtein: I think we can exchange this model to a global setting: `Setting[:rh_cloud_disable_insights_proxy]` that will be checked by the `ensure_telemetry_enabled_for_consumer` method.

app/services/foreman_rh_cloud/cloud_request_forwarder.rb
xprazak2: Do we have any docs on how to do cert auth to cloud? Inspired by [inventory upload](https://github.com/theforeman/foreman_rh_cloud/blob/master/lib/foreman_inventory_upload/scripts/uploader.sh.erb#L39), I tried:

```ruby
    RestClient::Request.execute(
        method: req_method,
        url: cloud_url,
        verify_ssl: ForemanRhCloud.verify_ssl_method,
        payload: forward_payload,
        ssl_client_cert: OpenSSL::X509::Certificate.new(certs[:cert]),
        ssl_client_key: OpenSSL::PKey::RSA.new(certs[:key]),
        headers: {
          params: forward_params,
        }
```
which gives me 401: Authentication missing. @lphiri, could you point me in the right direction please?

xprazak2: I managed to authenticate with certs to cert.cloud however I was getting 404 for endpoints that were previously working, so it seems we'll need to stick to token based auth for the time being:

```
api/v1/static/release/insights-core.egg?branch_id=20d6765d-b5aa-4bf2-b8c3-efda358dff98", "Accept"=>"*/*", "User-Agent"=>"rest-client/2.1.0 (linux-gnu x86_64) ruby/2.5.5p157"
 c06cca6e |
2020-10-20T10:28:40 [D|app|c06cca6e] # => 404 NotFound | text/plain 31 bytes, 0.30s

/api/v1/static/uploader.v2.json?branch_id=20d6765d-b5aa-4bf2-b8c3-efda358dff98", "Accept"=>"*/*", "User-Agent"=>"rest-client/2.1.0 (linux-gnu x86_64) ruby/2.5.5p157"
 f604e6d6 |
2020-10-20T10:28:41 [D|app|f604e6d6] # => 404 NotFound | text/plain 31 bytes, 0.22s
 f604e6d6 | 
```

xprazak2:
I replaced telemetry config with a setting and cleaned up a bit.

app/services/foreman_rh_cloud/cloud_request_forwarder.rb
lphiri: @xprazak2 i would not expect changing auth type to be the cause of the 404s.  Are you saying you get a  404 when you use cert.cloud.redhat.com (vs cloud.redhat.com)?

xprazak2:
Now cert auth is used if possible. For several endpoints that cannot be reached on cert.cloud, token based auth is still used.

app/services/foreman_rh_cloud/cloud_request_forwarder.rb
xprazak2: Yes, exactly.

ShimShtein: As we discussed on the mail, we can use `cloud.redhat.com` for those endpoints. Even more so, they don't require any authentication at all.

ShimShtein (changes requested):

xprazak2:
I removed the token from header, cert url can be configured via env variable and moved `prepare_forward_cloud_url` to forwarder.

xprazak2:
Any additional concerns?

app/services/foreman_rh_cloud/cloud_request_forwarder.rb
```diff
@@ -0,0 +1,101 @@
...
+
+    def core_app_name
+      'Foreman'
+    end
+
+    def core_app_version
+      Foreman::Version.new
+    end
+
+    def http_user_agent(original_request)
```
ShimShtein, line 93: If I am not mistaken, this function is only used in `log.debug` message.

ShimShtein (changes requested):

app/services/foreman_rh_cloud/cloud_request_forwarder.rb
xprazak2: Yes, I understood from a previous [comment](https://github.com/theforeman/foreman_rh_cloud/pull/375#discussion_r503960544) that it is better to keep because we modify the outgoing params.

ShimShtein: I think we should modify the `user_agent` as part of the forwarding too.

xprazak2: Done.

ares:
What is the status here?

ShimShtein:
I think it can be closed in favor of https://github.com/theforeman/foreman_rh_cloud/pull/412. And about that one - I am waiting for @rexwhite to take it for a spin before I merge it, to make sure we're not missing something in the old plugin.

ShimShtein:
@xprazak2, I think this one should be closed, since #412 is already in.

xprazak2:
Right, closing. Thanks!

PR #400, opened 2020-10-26
Fix hits item structure to match changed response

laviro:


ShimShtein:
The change was introduced in foreman 2.3: https://community.theforeman.org/t/notice-root-node-on-json-dropped/20837

laviro:
lets keep it open until we'll open a new branch?

ShimShtein:
Merged, thanks @laviro!

PR #401, opened 2020-10-29
Link to foreman's host recommendations

laviro:
instead of linking to the cloud recommendations.


ShimShtein (changes requested):
We also need to update foreman requirement in `engine.rb`, since it depends on a new feature in foreman.

laviro:
Ready for another round @ShimShtein :)

ShimShtein:
Merged, thanks @laviro!

PR #404, opened 2020-11-02
Make sure mtu is integer

ShimShtein:


laviro (commented):
Tested and it works great!
can you also add a test for a case where `mtu` should not exist in the report?

ShimShtein:
@laviro checks added.

laviro (approved):
LGTM thanks @ShimShtein !
now for the :cherries:

PR #411, opened 2020-11-13
Add rake tasks for report upload and generate + upload together

mirekdlugosz:
Add two more rake tasks, to enable full plugin functionality from command line:
* `rh_cloud_inventory:report:generate_upload` will both generate and upload report
* `rh_cloud_inventory:report:upload` will upload report from directory specified by `target` environment variable; it very much requires to call `rh_cloud_inventory:report:generate` first

This is requested by Insights QE team, who are working on end-to-end FiFi scenarios, and they don't want to involve entire UI interaction machinery only to upload report.

### DIFF
--- a/lib/tasks/rh_cloud_inventory.rake
+++ b/lib/tasks/rh_cloud_inventory.rake
@@ -2,6 +2,21 @@ require 'tempfile'
 
 namespace :rh_cloud_inventory do
   namespace :report do
+    desc 'Generate inventory report and send it to Red Hat cloud'
+    task generate_upload: [:environment, 'dynflow:client'] do
+      unless ENV['organization_id'].nil?
+        organizations = [ Organization.where(:id => ENV['organization_id']).first ]
+      else
+        organizations = Organization.unscoped.all
+      end
+
+      User.as_anonymous_admin do
+        organizations.each do |organization|
+          ForemanInventoryUpload::Async::GenerateReportJob.perform_now(ForemanInventoryUpload.generated_reports_folder, organization.id)
+          puts "Generated and uploaded inventory report for organization '#{organization.name}'"
+        end
+      end
+    end
     desc 'Generate inventory report to be sent to Red Hat cloud'
     task generate: :environment do
       portal_user = ENV['portal_user']
@@ -26,6 +41,14 @@ namespace :rh_cloud_inventory do
           puts "Successfully generated #{target} for organization id #{organization}"
         end
       end
+    end
+    desc 'Upload generated inventory report to Red Hat cloud'
+    task upload: [:environment, 'dynflow:client'] do
+      base_folder = ENV['target'] || ForemanInventoryUpload.generated_reports_folder
+      organization_id = ENV['organization_id']
+      report_file = ForemanInventoryUpload.facts_archive_name(organization_id)
+      ForemanInventoryUpload::Async::QueueForUploadJob.perform_now(base_folder, report_file, organization_id)
+      puts "Uploaded #{report_file}"
     end
   end
 
--- a/README.md
+++ b/README.md
@@ -19,9 +19,22 @@ In UI: Configure -> Inventory Upload -> Restart
 
 From command-line:
 
+    # generate and upload report for all organizations
+    /usr/sbin/foreman-rake rh_cloud_inventory:report:generate_upload
+
+    # generate and upload report for specific organization
+    export organization_id=1
+    /usr/sbin/foreman-rake rh_cloud_inventory:report:generate_upload
+
+    # generate report for specific organization (don't upload)
     export organization_id=1
     export target=/var/lib/foreman/red_hat_inventory/generated_reports/
     /usr/sbin/foreman-rake rh_cloud_inventory:report:generate
+
+    # upload previously generated report (needs to be named 'report_for_#{organization_id}.tar.gz')
+    export organization_id=1
+    export target=/var/lib/foreman/red_hat_inventory/generated_reports/
+    /usr/sbin/foreman-rake rh_cloud_inventory:report:upload
 
 #### Fetch hosts remediation data
 


lib/tasks/rh_cloud_inventory.rake
```diff
@@ -2,6 +2,21 @@ require 'tempfile'
 
 namespace :rh_cloud_inventory do
   namespace :report do
+    desc 'Generate inventory report and send it to Red Hat cloud'
+    task generate_upload: [:environment, 'dynflow:client'] do
```
mirekdlugosz, line 6: I took `'dynflow:client'` part from https://github.com/theforeman/foreman-tasks/commit/81bf0968328e8eb3f921729eb42d9bdcb66c0a3a . Without it, trying to run task would result in exception:
```
rake aborted!
The Dynflow world was not initialized yet. If your plugin uses it, make sure to call Rails.application.dynflow.require! in some initializer
/opt/theforeman/tfm/root/usr/share/gems/gems/dynflow-1.4.7/lib/dynflow/rails.rb:75:in `world'
```

I don't understand why these two tasks require that, but existing tasks calling `execute_now` do not.

lib/tasks/rh_cloud_inventory.rake
```diff
@@ -2,6 +2,21 @@ require 'tempfile'
...
+      if ! ENV['organization_id'].nil?
+        organizations = [ Organization.where(:id => ENV['organization_id']).first ]
+      else
+        organizations = Organization.unscoped.all
+      end
+
+      User.as_anonymous_admin do
+        organizations.each do |organization|
+          ForemanInventoryUpload::Async::GenerateReportJob.perform_now(ForemanInventoryUpload.generated_reports_folder, organization.id)
```
mirekdlugosz, line 15: I decided to use that instead of `GenerateAllReportsJob`, since all reports contain:
```ruby
      ensure
        self.class.set(:wait => 24.hours).perform_later
      end
```

As I understand it, calling rake task would automatically schedule next run in 24 hours, effectively adding another upload task to Satellite queue. Using `GenerateReportJob` avoids that problem.

lib/tasks/rh_cloud_inventory.rake
```diff
@@ -27,6 +42,14 @@ namespace :rh_cloud_inventory do
         end
       end
     end
+    desc 'Upload generated inventory report to Red Hat cloud'
+    task upload: [:environment, 'dynflow:client'] do
+      base_folder = ENV['target'] || Dir.pwd
+      organization_id = ENV['organization_id']
+      report_file = "report_for_#{organization_id}.tar.gz"
```
ShimShtein, line 49: You can use our internal constant here:
```suggestion
      report_file = ForemanInventoryUpload.facts_archive_name(organization_id)
```

lib/tasks/rh_cloud_inventory.rake
```diff
@@ -27,6 +42,14 @@ namespace :rh_cloud_inventory do
         end
       end
     end
+    desc 'Upload generated inventory report to Red Hat cloud'
+    task upload: [:environment, 'dynflow:client'] do
+      base_folder = ENV['target'] || Dir.pwd
```
ShimShtein, line 47: I think `Dir.pwd` would point to the folder where the script resides in prod environment.
I would be better to set it to the default base folder:
```suggestion
      base_folder = ENV['target'] || ForemanInventoryUpload.generated_reports_folder
```

ShimShtein (changes requested):
A couple of nitpicks, and we're done.

Can you please rebase the PR too? It should fix the issue with travis tests.

lib/tasks/rh_cloud_inventory.rake
mirekdlugosz: Good idea, done.

lib/tasks/rh_cloud_inventory.rake
mirekdlugosz: I copied the line that we already have in `rh_cloud_inventory:report:generate`. But I do like your suggestion more, especially as it would fix one of issues reported in bugzilla (where running rake task on fresh Satellite fails, since directory structure does not exist yet).

Should I apply same change in `rh_cloud_inventory:report:generate`?

lib/tasks/rh_cloud_inventory.rake
mirekdlugosz: Ah, I see it now - I might not use `perform_later` myself, but `GenerateReportJob` will schedule next step in commands chain using `perform_later`. Makes sense, thanks for explanation.

mirekdlugosz (commented):
@ShimShtein thanks for your comments, I included them in my commit. Also, rebased it on top of fresh master. Let's see what travis thinks now.

lib/tasks/rh_cloud_inventory.rake
ShimShtein: I think defaulting to the `generated_reports_folder` could be problematic during report creation due to permissions - we can't be sure the user has permissions to write into foreman's folder. I suppose it would be better to default to creation in the current folder. 

We probably need to test it with `foreman-rake` to see if the folder is reported correctly though.

lib/tasks/rh_cloud_inventory.rake
```diff
@@ -27,6 +42,14 @@ namespace :rh_cloud_inventory do
         end
       end
     end
+    desc 'Upload generated inventory report to Red Hat cloud'
+    task upload: [:environment, 'dynflow:client'] do
```
ShimShtein, line 46: I see `dynflow:client` was not enough. Got exception thrown at me:
```
2020-11-17T18:10:44 [I|app|] Performing ForemanInventoryUpload::Async::QueueForUploadJob (Job ID: c7953845-ebe4-415d-b79e-df803b51d5f1) from Dynflow(default) enqueued at  with arguments: "/home/vagrant/foreman/red_hat_inventory/generated_reports/", "report_for_1.tar.gz", "1"
2020-11-17T18:10:44 [D|bac|] Ensuring objects
2020-11-17T18:10:44 [D|bac|] Copying report_for_1.tar.gz to /home/vagrant/foreman/red_hat_inventory/uploads/
2020-11-17T18:10:44 [D|bac|] Done copying report_for_1.tar.gz to /home/vagrant/foreman/red_hat_inventory/uploads/report_for_1.tar.gz
E, [2020-11-17T18:10:44.761724 #30440] ERROR -- /connector-database-core: No executor available (Dynflow::Error)
/home/vagrant/foreman/.vendor/ruby/2.5.0/gems/dynflow-1.4.7/lib/dynflow/connectors/database.rb:143:in `any_executor'
/home/vagrant/foreman/.vendor/ruby/2.5.0/gems/dynflow-1.4.7/lib/dynflow/connectors/database.rb:135:in `find_receiver'
...
/home/vagrant/foreman/.vendor/ruby/2.5.0/gems/activejob-6.0.3.4/lib/active_job/enqueuing.rb:22:in `perform_later'
/home/vagrant/foreman_rh_cloud/lib/foreman_inventory_upload/async/queue_for_upload_job.rb:15:in `perform'
```

@adamruzicka, maybe you know a better solution to this?

ShimShtein (changes requested):
I was not able to run the tasks, even with foreman up in a different ssh session.

lib/tasks/rh_cloud_inventory.rake
mirekdlugosz: It's `/usr/sbin/foreman-rake`, so it's already expected to be run by root/user with elevated permissions. Personally, I wouldn't worry about insufficient permissions here.

By the way, `Dir.pwd` isn't current working dir in shell - it seems to be changed to `/usr/share/foreman/` somewhere inside `foreman-rake` call:
```
# cd /tmp/
# export organization_id=1
# /usr/sbin/foreman-rake rh_cloud_inventory:report:generate
Must specify either portal_user or organization_id
I, [2020-11-18T06:34:00.043724 #25541]  INFO -- : Started generating hosts report in /tmp/d20201118-25541-suckmw
I, [2020-11-18T06:34:00.059385 #25541]  INFO -- : Adding slice 99a3216e-22bb-4a0c-8b54-934067e4a7ce
I, [2020-11-18T06:34:00.905352 #25541]  INFO -- : slice 99a3216e-22bb-4a0c-8b54-934067e4a7ce was created with 1 hosts
I, [2020-11-18T06:34:00.905479 #25541]  INFO -- : Report generation finished
I, [2020-11-18T06:34:00.905503 #25541]  INFO -- : Archiving generated report
I, [2020-11-18T06:34:00.909942 #25541]  INFO -- : tar: tar (child): /usr/share/foreman/report_for_1.tar.gz: Cannot open: Permission denied
tar (child): Error is not recoverable: exiting now
./
./metadata.json
./99a3216e-22bb-4a0c-8b54-934067e4a7ce.json

I, [2020-11-18T06:34:00.910009 #25541]  INFO -- : Tar command failed: pid 25562 exit 0
Successfully generated /usr/share/foreman/report_for_1.tar.gz for organization id 1
```

lib/tasks/rh_cloud_inventory.rake
```diff
@@ -2,6 +2,21 @@ require 'tempfile'
 
 namespace :rh_cloud_inventory do
   namespace :report do
+    desc 'Generate inventory report and send it to Red Hat cloud'
+    task generate_upload: [:environment, 'dynflow:client'] do
+      if ! ENV['organization_id'].nil?
```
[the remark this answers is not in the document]
mirekdlugosz: Changed in 50753ec965e28e616026e8f299fc3e677efd2bb6

lib/tasks/rh_cloud_inventory.rake
mirekdlugosz: Hm, I didn't get any exception on Satellite 6.8.1 (Foreman 2.1.2.20) or 6.9.0 (Foreman 2.3.0).

adamruzicka: Depending on `dynflow:client` marks the task as "allowed to communicate with dynflow, without having the capability to execute jobs on its own", which is fine. Since we're using dynflow as active job backend, submitting activejob jobs using `klass.perform_later` will require dynflow executor (dynflowd or dynflow-sidekiq@orchestrator) running. Without it it cannot perform those actions in background.

We could make the rake task act as an executor, but setting that up would be a bit more involved.

ShimShtein:
Works as expected. To test it in development environment I have started up another ssh session with `bundle exec rake dynflow:executor`.

ShimShtein:
@mirekdlugosz, if you want it in older foreman versions, ping me, I'll create cherry-picks into proper branches

mirekdlugosz:
@ShimShtein we are only really seeing Sat 6.9 snaps and Sat 6.8.z releases, so if you could cherry-pick into branch for 6.8, that would be great.
Older branches would be nice to have, but we can skip them.

ShimShtein:
Sure, will do

PR #413, opened 2020-11-17
remove rvm cache

laviro:


ShimShtein:
Couldn't find any other option except from disabling caching. Merging the PR.

PR #421, opened 2020-11-29
Add Insights page layout with search

laviro:
The plan is to add first the search bar,
and on another PR to create the table that will interact with it

### DIFF
--- a/app/controllers/insights_cloud/hits_controller.rb
+++ b/app/controllers/insights_cloud/hits_controller.rb
@@ -1,11 +1,25 @@
 module InsightsCloud
   class HitsController < ::ApplicationController
+    include Foreman::Controller::AutoCompleteSearch
+
     def index
+      hits = resource_base_search_and_page
+
+      render json: {
+        hits: hits,
+      }, status: :ok
+    end
+
+    def show
       host = Host.where(id: host_id_param).first
 
       render json: {
         hits: host.insights.hits,
       }, status: :ok
+    end
+
+    def model_of_controller
+      ::InsightsHit
     end
 
     private
--- a/webpack/InsightsCloudSync/InsightsCloudSync.js
+++ b/webpack/InsightsCloudSync/InsightsCloudSync.js
@@ -1,61 +1,64 @@
-import React from 'react';
-import { IntlProvider } from 'react-intl';
+/* eslint-disable react-hooks/exhaustive-deps */
+import React, { useEffect } from 'react';
+import { Button } from 'patternfly-react';
 import PropTypes from 'prop-types';
 import { translate as __ } from 'foremanReact/common/I18n';
-import { Button, Icon } from 'patternfly-react';
-import { INSIGHTS_SYNC_PAGE_TITLE } from './InsightsCloudSyncConstants';
-import InsightsSettings from './Components/InsightsSettings';
-import { cloudTokenSettingUrl } from './InsightsCloudSyncHelpers';
+import PageLayout from 'foremanReact/routes/common/PageLayout/PageLayout';
+import { useForemanSettings } from 'foremanReact/Root/Context/ForemanContext';
+import InsightsHeader from './Components/InsightsHeader';
+import {
+  INSIGHTS_SYNC_PAGE_TITLE,
+  INSIGHTS_SEARCH_PROPS,
+} from './InsightsCloudSyncConstants';
 
-const InsightsCloudSync = ({ syncInsights }) => {
-  document.title = INSIGHTS_SYNC_PAGE_TITLE;
+const InsightsCloudSync = ({
+  syncInsights,
+  query,
+  fetchInsights,
+  page,
+  perPage: urlPerPage,
+}) => {
+  const { perPage: appPerPage } = useForemanSettings();
+  const perPage = urlPerPage || appPerPage;
+
+  // acts as componentDidMount
+  useEffect(() => {
+    fetchInsights({ page, perPage, searchQuery: query });
+  }, []);
+
   return (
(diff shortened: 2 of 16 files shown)


app/controllers/insights_cloud/hits_controller.rb
```diff
@@ -1,17 +1,39 @@
 module InsightsCloud
   class HitsController < ::ApplicationController
+    include Foreman::Controller::AutoCompleteSearch
+
     def index
+      page, per_page, search = insights_search_params.values_at(:page, :per_page, :search)
+      hits = InsightsHit.search_for(search)
```
ShimShtein, line 7: You can use [`resource_base_search_and_page`](https://github.com/theforeman/foreman/blob/7a4899774a957bf7336db12f5643ed21024a929c/app/controllers/application_controller.rb#L240) here.

webpack/InsightsCloudSync/InsightsCloudSync.js
```diff
@@ -1,57 +1,61 @@
...
+  updateUrlQuery: PropTypes.func.isRequired,
+  fetchInsights: PropTypes.func.isRequired,
+  page: PropTypes.number,
+  perPage: PropTypes.number,
+  query: PropTypes.string,
+};
+
+InsightsCloudSync.defaultProps = {
+  page: 1,
+  perPage: 25,
```
ShimShtein, line 57: Does it mean it will send `25` even if nobody set this value? It's not consistent with the server side setting.

@mariaaga, is there a better way to sync pagination properties with the server?

ShimShtein (changes requested):

webpack/InsightsCloudSync/InsightsCloudSync.js
laviro: I think we could use foreman context as default for pagination,
those values are initially set when the foreman app starts,
another option is to create an initial API call on our page in order to get those values.

ShimShtein: I think we need to use foreman's context. It already handles pagination values for us.

laviro: done, using its value if `per_page` from `url` is `undefined`

laviro:
rebased on top of the router PR #424
and addressed your suggestions, thanks @ShimShtein !

MariaAga (changes requested):
Looks good! The only issue is that on search does a full page reload for me

webpack/InsightsCloudSync/InsightsCloudSyncActions.js
```diff
@@ -20,3 +28,35 @@ export const syncInsights = () => async dispatch => {
     );
   }
 };
+
+export const updateUrlQuery = (queryParams = {}) => dispatch => {
+  dispatch(
+    push({
```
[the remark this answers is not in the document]
laviro: I couldn't reproduce, maybe with some of the new changes, it's gone? can you check again? thanks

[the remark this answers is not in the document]
MariaAga: Yeah it's gone now :tada:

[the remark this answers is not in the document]
MariaAga: Can you remove `dispatch` here as well?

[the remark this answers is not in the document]
laviro: I need the `getState` now too so we have to use `dispatch`

MariaAga (approved):

MariaAga:
There was a bug opened in foreman tasks, that users want the search button to run search every time it's clicked and not only when the query changed.
@laviro maybe the behavior here should change to that?

webpack/InsightsCloudSync/InsightsCloudSync.js
laviro: had another thought about it, 
in this case, the table won't cause the scrollbar to appear when there are 5 rows per page,
maybe 5 should be our default?
above that, it will make the user to scroll down.

laviro:
> There was a bug opened in foreman tasks, that users want the search button to run search every time it's clicked and not only when the query changed.
> @laviro maybe the behavior here should change to that?

make sense @MariaAga,
I wanted to avoid the extra API call, but I guess some data could change/ error could accure serverside?
I will fix that :)

laviro:
Done, fixes are int the second commit, ready for another review :)

laviro:
Router PR got in, rebased on top of it

MariaAga:
I tried adding a bookmark and got this error: `/foreman_rh_cloud/insights_cloud is not a valid controller`

ShimShtein:
@laviro Found out why the bookmarks are not added. It comes from this method: [`valid_controllers_list`](https://github.com/theforeman/foreman/blob/812eb536e6047c88639d08b361715781e32ff3ba/app/validators/bookmark_controller_validator.rb#L8) method. I think we should add proper permissions for our controller and it will solve also the issue here.

laviro:
@ShimShtein would it be OK to do something like https://github.com/theforeman/foreman-tasks/blob/master/db/seeds.d/61-foreman_tasks_bookmarks.rb#L9? we could also create some useful bookmarks there like: `High Risk` or `with playbook`...

I will also look into the permissions area, thanks!

ShimShtein:
Yes it would be OK, but not in this PR. I think the permissions (and bookmarks as part of it) should go to the table PR.

ShimShtein (approved):

ShimShtein:
Merged, thanks @laviro!

PR #423, opened 2020-12-02
Add insights hits table

laviro:
the second commit adds the table,
the first commit contains the search bar and page layout which is in PR #421

blocked by #421


webpack/InsightsCloudSync/Components/InsightsTable/InsightsTable.js
```diff
@@ -0,0 +1,32 @@
...
+    <Table
+      key="insights-table"
+      columns={createInsightsTableSchema()}
+      rows={results}
+      id="insights-table"
+      style={{ marginBottom: -6 }}
+    />
+    <Pagination
+      viewType="list"
+      itemCount={8}
```
laviro, line 18: hard coded for testing, need to get the total-hits-count from server

webpack/InsightsCloudSync/Components/InsightsSettings/InsightsSettings.js
```diff
@@ -12,9 +12,17 @@ class InsightsSettings extends Component {
...
   render() {
     const { insightsSyncEnabled, setInsightsSyncEnabled } = this.props;
+    const insightsStyles = {
+      border: '1px solid #ededed',
+      borderRadius: '4px',
+      padding: '10px',
+      marginBottom: '15px',
+      float: 'right',
+    };
+
```
laviro, line 22: currently when using foreman router there's no css loaded, from the next version importing scss will be supported.

this is a workaround to get our styles to work.

lib/foreman_rh_cloud/engine.rb
```diff
@@ -79,7 +80,6 @@ class Engine < ::Rails::Engine
 
       ::Katello::UINotifications::Subscriptions::ManifestImportSuccess.include ForemanInventoryUpload::Notifications::ManifestImportSuccessNotificationOverride if defined?(Katello)
 
-      ::Host::Managed.include RhCloudHost
```
laviro, line 82: Big thanks for the help on this @tbrisker ! :1st_place_medal: 
Looks like it wasn't needed and caused that `NameError (undefined local variable or method 'taxonomy_conditions' for#<Host::Managed::ActiveRecord_Relation:0x00007f0eaff32720>)`

laviro:
Update also with sorting :)

Red Hat Insights (1)

laviro:
Updated with PF4 table, still a WIP:

Red Hat Insights (2)

ShimShtein:
This will also require permissions if we want the bookmarks to work and generally keep our hits table secure :grinning:

ShimShtein (commented):
We were talking about select all implementation through sending the query directly to the server.

laviro:
Still in progress, planning a refactor into smaller components, and adding tests

laviro:
Still need to add tests but feel free to debug it and comment @ShimShtein! thanks

webpack/InsightsCloudSync/Components/InsightsTable/InsightsTableHelpers.js
```diff
@@ -0,0 +1,36 @@
...
+import { columns } from './InsightsTableConstants';
+
+export const mapResultsToRows = (results, selectedIds) => {
+  if (results.length === 0) return [];
+
+  return results.asMutable().map(({ id, hostname, title, total_risk }) => {
+    const row = [
+      hostname,
+      title,
+      <Section className="insights-total-risk" type="icon-group">
```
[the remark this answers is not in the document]
laviro: I agree it's a good spot for that but on a first impression looks like it transforms only the column's header?
@MariaAga can you also try and take a look?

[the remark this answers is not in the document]
MariaAga: Maybe `cellTransforms` ?

ShimShtein (commented):
All in all looks good.

webpack/InsightsCloudSync/InsightsCloudSyncConstants.js
```diff
@@ -3,13 +3,15 @@ import { getControllerSearchProps } from 'foremanReact/constants';
...
+
 export const INSIGHTS_SYNC_PAGE_TITLE = __('Red Hat Insights');
+
 export const INSIGHTS_PATH = foremanUrl('/foreman_rh_cloud/insights_cloud');
-export const INSIGHTS_HITS_PATH = '/insights_cloud/hits';
+
 export const INSIGHTS_SAVE_AND_SYNC_PATH = foremanUrl(
   '/insights_cloud/save_token_and_sync'
 );
-export const INSIGHTS_HITS_API_KEY = 'INSIGHTS_HITS';
```
MariaAga, line 12: This is used in `webpack/InsightsCloudSync/Components/NoTokenEmptyState.js`

webpack/InsightsCloudSync/Components/InsightsTable/SelectAllAlert.js
```diff
@@ -0,0 +1,72 @@
...
+    return (
+      <Alert
+        isInline
+        variant="info"
+        title={`All ${selectedCount} recommendations on this page are selected.`}
+        actionClose={<AlertActionCloseButton onClose={onClose} />}
+        actionLinks={
+          <React.Fragment>
+            <AlertActionLink onClick={selectAll}>
+              Select all {itemCount} recommendations
```
MariaAga, line 30: Should be translated with `sprintf`

webpack/InsightsCloudSync/Components/InsightsTable/SelectAllAlert.js
```diff
@@ -0,0 +1,72 @@
...
+}) => {
+  if (!showSelectAllAlert) return null;
+  const selectedCount = Object.keys(selectedIds).length;
+  if (!selectedCount) return null;
+  if (itemCount > selectedCount) {
+    return (
+      <Alert
+        isInline
+        variant="info"
+        title={`All ${selectedCount} recommendations on this page are selected.`}
```
MariaAga, line 25: Should also be translated with `sprintf`

webpack/InsightsCloudSync/Components/InsightsTable/InsightsTableActions.js
```diff
@@ -0,0 +1,125 @@
...
+  };
+  dispatch(
+    get({
+      key: `${INSIGHTS_HITS_API_KEY}_ALL`,
+      url: INSIGHTS_HITS_PATH,
+      params: {
+        search: query,
+      },
+      handleSuccess,
+    })
```
MariaAga, line 75: This will not return all the results. I think it returns the default perpage amount of results. Also adding all the selected IDs might cause issue if there are thousands of them (not sure how many recommendations users can have)

webpack/InsightsCloudSync/Components/InsightsTable/InsightsTableConstants.js
```diff
@@ -0,0 +1,47 @@
...
+];
+
+export const actions = [
+  // {
+  //   title: __('Remediate'),
+  //   onClick: (event, rowId, rowData, extra) => {
+  //     alert('Remediate!');
+  //   },
+  // },
+];
```
MariaAga, line 39: left over or will this be in a different pr?

webpack/InsightsCloudSync/Components/InsightsTable/components/EmptyState.js
```diff
@@ -0,0 +1,54 @@
...
+import { translate as __, sprintf } from 'foremanReact/common/I18n';
+
+const TableEmptyState = ({ status, error }) => {
+  switch (status) {
+    case STATUS.PENDING:
+      return (
+        <EmptyState variant={EmptyStateVariant.small}>
+          <EmptyStateIcon variant="container" component={Spinner} />
+          <Title headingLevel="h2" size="lg">
+            Loading
```
MariaAga, line 23: Should be translated

webpack/InsightsCloudSync/Components/InsightsTable/components/EmptyState.js
```diff
@@ -0,0 +1,54 @@
+/* eslint-disable react-hooks/exhaustive-deps */
```
MariaAga, line 1: Is this needed?

webpack/InsightsCloudSync/InsightsCloudSync.js
```diff
@@ -1,61 +1,23 @@
 /* eslint-disable react-hooks/exhaustive-deps */
```
MariaAga, line 1: Is  this needed?

MariaAga (changes requested):
Nice PR! It's great seeing the pf4 table in use :) 
I had some issues with the select and perPage options, as in the settings my perPage is 7, so I left some comments about that inline.
Also if I use select all and then unselect using the checkbox (instead of  the "clear all") it only clears the selected boxes in the page so if I change the page there are boxes still selected, I think it makes more sense to unselect all the items and not only the items in the page.

webpack/InsightsCloudSync/Components/InsightsTable/InsightsTableHelpers.js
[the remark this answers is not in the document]
laviro: That works :)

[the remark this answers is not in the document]
laviro: @ShimShtein as we discussed offline on using the [IRow](https://github.com/patternfly/patternfly-react/blob/master/packages/react-table/src/components/Table/TableTypes.ts#L204) for selecting it on the initial stage,
looks like [TS interfaces are not compiled to JS](https://www.tutorialsteacher.com/typescript/typescript-interface#:~:text=The%20TypeScript%20compiler%20does%20not,function%20or%20an%20arrow%20function.) :(
worth thinking about using typescript across our app, at least for better integration with PF4,
cc @theforeman/ui-ux

webpack/InsightsCloudSync/Components/InsightsTable/InsightsTableConstants.js
laviro: I guess we can have it now for visibility though it will do nothing

webpack/InsightsCloudSync/Components/InsightsTable/InsightsTableActions.js
laviro: Looking at [resource_base_search_and_page](https://github.com/theforeman/foreman/blob/7a4899774a957bf7336db12f5643ed21024a929c/app/controllers/application_controller.rb#L242) which we use in our controller,
I think it will return all of the results when `per_page` is `nil`
cc @ShimShtein

webpack/InsightsCloudSync/Components/InsightsTable/InsightsTableConstants.js
```diff
@@ -0,0 +1,47 @@
...
+    transforms: [cellWidth(65), sortable],
+  },
+  {
+    sortKey: 'total_risk',
+    title: __('Total Risk'),
+    transforms: [cellWidth(15), sortable],
+  },
+];
+
+export const perPageOptions = [
```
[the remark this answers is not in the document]
laviro: great, I will use that!

webpack/InsightsCloudSync/Components/InsightsTable/InsightsTableActions.js
MariaAga: I think per_page is not nil if it's set in the settings as it returns  7 results for me

webpack/InsightsCloudSync/InsightsCloudSyncConstants.js
laviro: good catch!

laviro:
Thanks, @ShimShtein @MariaAga I think I answered most of your comments,
still need to add tests to this PR

webpack/InsightsCloudSync/Components/InsightsTable/InsightsTable.js
```diff
@@ -0,0 +1,117 @@
...
+      <TableEmptyState status={status} error={error} />
+      <Pagination
+        itemCount={itemCount}
+        widgetId="recommendation-pagination"
+        perPage={perPage}
+        page={page}
+        variant={PaginationVariant.bottom}
+        onSetPage={onTableSetPage}
+        onPerPageSelect={onTablePerPageSelect}
+        perPageOptions={getPerPageOptions(perPage)}
```
MariaAga, line 78: This adds the current option to the array instead of adding the settings & url custom options. So for example when perPage is 5 it shows up like this: 
Screenshot from 2021-01-07 11-17-14

webpack/InsightsCloudSync/Components/InsightsTable/InsightsTableActions.js
ShimShtein: We have discussed an option to send the search condition in case "select all" is chosen. I think the UI should lock the checkboxes in selected state:
So basically we have two different operation modes:
1. _Some_ of the items are selected: then ids are collected and will be sent to the action buttons.
2. _All_ items are selected (across all pages): then ids won't be selected (hence locked to the user) and the search condition will be sent to the action buttons instead of ids list. This will remove the need to send tons of ids to the client and then back to the server.

webpack/InsightsCloudSync/Components/InsightsTable/InsightsTable.js
laviro: Thanks, solved it in commit https://github.com/theforeman/foreman_rh_cloud/pull/423/commits/0e74f4752048333972cf2b3ed651ad596bf5b4ad

webpack/InsightsCloudSync/Components/InsightsTable/InsightsTableActions.js
laviro: Thanks @MariaAga @ShimShtein! Addressed it in https://github.com/theforeman/foreman_rh_cloud/pull/423/commits/f1b0be7c67170dd83f17e14f622bb1427bb1bb50, let me know if this works for you :)

MariaAga:
@laviro Select all checkbox state doesn't get saved for me when changing the table page, is this still a WIP or a bug? :)

laviro:
> @laviro Select all checkbox state doesn't get saved for me when changing the table page, is this still a WIP or a bug? :)

As discussed with @ShimShtein offline, we decided to follow the gmail select all ux, which also clear all selected when transitioning or updating the page/table

> Also it looks like I can't submit a bookmark. I get this error:
> 
> ```
> /foreman_rh_cloud/insights_cloud is not a valid controller
> ```

@ShimShtein what are we deciding? Would this be a parallel effort in another PR?

webpack/InsightsCloudSync/Components/InsightsTable/__tests__/fixtures.js
```diff
@@ -0,0 +1,102 @@
+import Immutable from 'seamless-immutable';
+import { noop } from 'patternfly-react';
+
+const hits = Immutable([
```
MariaAga, line 4: Can hits only have the needed props? It doesn't have here `hostname` which is used in the table but has `last_seen`

MariaAga (changes requested):
> As discussed with ShimShtein offline, we decided to follow the gmail select all ux, which also clear all selected when transitioning or updating the page/table

Makes sense, I didn't know gmail did that.

the PR looks good! I have only a few small request:

webpack/InsightsCloudSync/Components/InsightsTable/__tests__/fixtures.js
ShimShtein: I prefer the API responses to be as accurate as possible

webpack/InsightsCloudSync/Components/InsightsTable/InsightsTableActions.js
```diff
@@ -0,0 +1,128 @@
...
+  });
+};
+
+export const onTableSetPage = (_event, pageNumber) =>
+  fetchInsights({ page: pageNumber });
+
+export const onTablePerPageSelect = (_event, perPageNumber) =>
+  fetchInsights({ perPage: perPageNumber });
+
+export const onTableSelect = (
```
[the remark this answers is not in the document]
ShimShtein: Is there any advantage in adding extra nesting?

ShimShtein:
@MariaAga, I think it's ready for another round of review.

webpack/InsightsCloudSync/Components/InsightsTable/InsightsTableActions.js
```diff
@@ -0,0 +1,128 @@
...
+
+    showSelectAllAlert = true;
+  } else {
+    isSelected
+      ? (selectedIds[results[rowId].id] = true)
+      : delete selectedIds[results[rowId].id];
+  }
+
+  dispatch(selectByIds(selectedIds));
+  return dispatch(setSelectAllAlert(showSelectAllAlert));
```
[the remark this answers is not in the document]
MariaAga: @ShimShtein Looks like I was wrong, it can't be just dispatch :)
Can you change it back to return setSelectAllAlert(showSelectAllAlert) ?

webpack/InsightsCloudSync/InsightsCloudSyncConstants.js
```diff
@@ -3,13 +3,15 @@ import { getControllerSearchProps } from 'foremanReact/constants';
...
+
 export const INSIGHTS_SEARCH_PROPS = {
   ...getControllerSearchProps('/insights_cloud/hits'),
   controller: INSIGHTS_PATH,
```
MariaAga, line 17: This should be `insights_hits` for the bookmarks to work

webpack/InsightsCloudSync/Components/InsightsTable/InsightsTableSelectors.js
```diff
@@ -0,0 +1,74 @@
...
+  query: selectSearch(state),
+  sortBy: selectSortBy(state),
+  sortOrder: selectSortOrder(state),
+});
+
+export const selectStatus = state =>
+  selectAPIStatus(state, INSIGHTS_HITS_API_KEY);
+
+export const selectError = state =>
+  selectAPIError(state, INSIGHTS_HITS_API_KEY);
```
MariaAga, line 41: Last issue: I think this should be `selectAPIErrorMessage` as `selectAPIError` returns an object and not a string so I get this warning:
```
Warning: Failed prop type: Invalid prop `error` of type `object` supplied to `InsightsTable`, expected `string`.
```

MariaAga (approved):

PR #424, opened 2020-12-10
Add React router for insights page

laviro:


webpack/InsightsCloudSync/Components/InsightsSettings/InsightsSettings.js
```diff
@@ -12,9 +12,17 @@ class InsightsSettings extends Component {
 
   render() {
     const { insightsSyncEnabled, setInsightsSyncEnabled } = this.props;
+    const insightsStyles = {
+      border: '1px solid #ededed',
```
ShimShtein, line 16: I am a bit reluctant about moving colors to inline style. I suppose it would be harder to themify it.
Maybe we should have some class or global  constant for it, so it would be easier to themify it in the same place?

webpack/InsightsCloudSync/InsightsCloudSync.js
```diff
@@ -25,13 +26,17 @@ const InsightsCloudSync = ({ settingsUrl, syncInsights }) => {
...
               rel="noopener noreferrer"
             >
-              access.redhat.com <Icon name="external-link" size="xs" />
+              access.redhat.com <Icon name="external-link" />
             </a>
             <br />
             {__("2. Copy the token to 'Red Hat Cloud token' setting: ")}
-            <a href={settingsUrl} target="_blank" rel="noopener noreferrer">
+            <a
+              href={foremanUrl('/settings?search=name+%3D+rh_cloud_token')}
```
ShimShtein, line 34: Are there any helpers for url construction either in foreman or in the framework? It feels a bit prone to errors to construct urls manually.

ShimShtein (commented):
All in all looks good.
A couple of inline comments, but if they aren't realistic, I will merge.

webpack/InsightsCloudSync/InsightsCloudSync.js
laviro: thanks, I assume we could use `urijs` for that :+1:

webpack/InsightsCloudSync/Components/InsightsSettings/InsightsSettings.js
laviro: good point.. I tried to do fetch the plugin css file on the global-js file (in the next foreman version it will be available from core)
though there is a delay in the view so you see the page breaking and changing (this happens only if you refresh the insights page) if you come from another page the styles will be already predefined.

MariaAga (approved):
LGTM!

laviro:
@ShimShtein as discussed offline, I updated deface to `insert_top`

laviro:
also removed the styles from being loaded twice on the inventory page which uses `application.html.erb` layout

ShimShtein (approved):

ShimShtein:
Merged, thanks @laviro!

PR #412, opened 2020-11-16
Add endpoints for proxying payload

xprazak2:
Contains #375 and #325 in a single branch for easier testing. Needs https://github.com/redhataccess/foreman-plugin/pull/56

PR #425, opened 2020-12-14
New insights status for the host

stejskalleos:
```  
REPORTING             = 0 # host_registration_insights = true & getting data
NO_REPORT             = 1 # host_registration_insights = true & not getting data
NOT_MANAGED           = 2 # host_registration_insights = false
NOT_MANAGED_WITH_DATA = 3 # host_registration_insights = false & getting data
```
host_registration_insights | sending data? | Insights status
-- | -- | --
`nil` | yes | `Reporting (not set by registration)`
`nil` | no | `Not reporting (not set by registration)`
`TRUE` | yes | `Reporting`
`TRUE` | no | `Not reporting`
`FALSE` | yes | `Reporting (not set by registration)`
`FALSE` | no | `Not reporting (not set by registration)`

### DIFF
--- a/app/controllers/insights_cloud/api/machine_telemetries_controller.rb
+++ b/app/controllers/insights_cloud/api/machine_telemetries_controller.rb
@@ -10,27 +10,28 @@ module InsightsCloud::Api
 
     skip_after_action :log_response_body, :only => [:forward_request]
     skip_before_action :check_media_type, :only => [:forward_request]
+    after_action :update_host_insights_status, only: [:forward_request]
 
     # The method that "proxies" requests over to Cloud
     def forward_request
       certs = candlepin_id_cert @organization
-      cloud_response = ::ForemanRhCloud::CloudRequestForwarder.new.forward_request(request, controller_name, @branch_id, certs)
+      @cloud_response = ::ForemanRhCloud::CloudRequestForwarder.new.forward_request(request, controller_name, @branch_id, certs)
 
-      if cloud_response.code == 401
+      if @cloud_response.code == 401
         return render json: {
           :message => 'Authentication to the Insights Service failed.',
           :headers => {},
         }, status: :bad_gateway
       end
 
-      if cloud_response.headers[:content_disposition]
-        return send_data cloud_response, disposition: cloud_response.headers[:content_disposition], type: cloud_response.headers[:content_type]
+      if @cloud_response.headers[:content_disposition]
+        return send_data @cloud_response, disposition: @cloud_response.headers[:content_disposition], type: @cloud_response.headers[:content_type]
       end
 
-      assign_header(response, cloud_response, :x_resource_count, true)
-      assign_header(response, cloud_response, :x_rh_insights_request_id, false)
+      assign_header(response, @cloud_response, :x_resource_count, true)
+      assign_header(response, @cloud_response, :x_rh_insights_request_id, false)
 
-      render json: cloud_response, status: cloud_response.code
+      render json: @cloud_response, status: @cloud_response.code
     end
 
     def branch_info
@@ -68,6 +69,15 @@ module InsightsCloud::Api
     def ensure_branch_id
       @branch_id = cp_owner_id(@organization)
       return render_message "Branch ID not found for organization #{@organization.title}", :status => 400 unless @branch_id
+    end
+
+    def update_host_insights_status
+      return unless request.path == '/redhat_access/r/insights/platform/ingress/v1/upload' ||
+                    request.path.include?('/redhat_access/r/insights/uploads/')
+
+      host_status = @host.get_status(InsightsClientReportStatus)
+      host_status.update(reported_at: DateTime.now)
+      host_status.update(status: host_status.to_status(data: @cloud_response.code.to_s.start_with?('2')))
     end
   end
 end
--- a/app/controllers/insights_cloud/api/machine_telemetries_controller.rb
+++ b/app/controllers/insights_cloud/api/machine_telemetries_controller.rb
@@ -75,9 +75,9 @@ module InsightsCloud::Api
       return unless request.path == '/redhat_access/r/insights/platform/ingress/v1/upload' ||
                     request.path.include?('/redhat_access/r/insights/uploads/')
 
+      data = @cloud_response.code.to_s.start_with?('2')
       host_status = @host.get_status(InsightsClientReportStatus)
-      host_status.update(reported_at: DateTime.now)
-      host_status.update(status: host_status.to_status(data: @cloud_response.code.to_s.start_with?('2')))
+      host_status.update(reported_at: Time.now.utc, status: host_status.to_status(data: data))
     end
   end
 end
--- a/app/models/insights_client_report_status.rb
+++ b/app/models/insights_client_report_status.rb
@@ -36,7 +36,7 @@ class InsightsClientReportStatus < HostStatus::Status
     end
   end
 
-  def to_status(data: nil)
+  def to_status(data: false)
     if insights_param
       return REPORTING if data
       return in_interval? ? REPORTING : NO_REPORT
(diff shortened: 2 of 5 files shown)


test/models/insight_status_test.rb
```diff
@@ -0,0 +1,50 @@
...
+  describe 'to_status' do
+    let(:host) { FactoryBot.create(:host, :managed) }
+    let(:host_status) { host.get_status(InsightsStatus) }
+
+    setup do
+      CommonParameter.where(name: 'host_registration_insights').destroy_all
+    end
+
+    test 'Should be registered & is getting data' do
+      CommonParameter.create(name: 'host_registration_insights', key_type: 'boolean', value: true, host: host)
```
ShimShtein, line 18: You can use `FactoryBot` to create parameters

app/models/insights_status.rb
```diff
@@ -0,0 +1,59 @@
...
+  def to_label(_options = {})
+    case status
+      when REGISTER_DATA
+        N_('Recieving data')
+      when REGISTER_NO_DATA
+        N_('No data')
+      when NO_REGISTER
+        N_('Not required')
+      when NO_REGISTER_DATA
+        N_('Not installed but recieving data')
```
ShimShtein, line 21: I think a better term would be "Not managed". Insights-client is obviously installed on the host.

app/models/insights_status.rb
```diff
@@ -0,0 +1,59 @@
+class InsightsStatus < HostStatus::Status
```
ShimShtein, line 1: Can we rename it to `InsightsClientReportStatus`. "Insights" is too broad term to understand what sort of status it will be.

app/models/insights_status.rb
```diff
@@ -0,0 +1,59 @@
...
+      NO_REGISTER_DATA
+    end
+  end
+
+  def in_interval?
+    (Time.now.utc - reported_at).to_i < REPORT_INTERVAL.to_i
+  end
+
+  def insights_param
+    host.host_param('host_registration_insights').nil? ? true : host.host_param('host_registration_insights')
```
ShimShtein, line 57: If the parameter is not set, I think the function should return `false` and not `true`.

ShimShtein (changes requested):

ShimShtein:
@stejskalleos what's the status here?

stejskalleos:
Sorry for late response, I've been busy lately.
This feature have been postponed to Sat 6.10, so right now I'm waiting for required PR (https://github.com/theforeman/foreman_rh_cloud/pull/412) to be done & merged ,and after that I'll start working on it.

stejskalleos:
@ShimShtein rebased & updated

app/models/insights_client_report_status.rb
```diff
@@ -0,0 +1,60 @@
...
+  end
+
+  def to_label(_options = {})
+    case status
+      when REPORTING
+        N_('Reporting')
+      when NO_REPORT
+        N_('Not reporting')
+      when NOT_MANAGED
+        N_('Not reporting')
```
ares, line 20: Not reporting (not set by registration)

app/models/insights_client_report_status.rb
```diff
@@ -0,0 +1,60 @@
...
+      when NOT_MANAGED_WITH_DATA
+        N_('Reporting')
```
ares, line 22: Reporting (not set by registration)

app/models/insights_client_report_status.rb
```diff
@@ -0,0 +1,60 @@
...
+  def to_global(_options = {})
+    case status
+      when REPORTING
+        ::HostStatus::Global::OK
+      when NO_REPORT
+        ::HostStatus::Global::ERROR
+      when NOT_MANAGED
+        ::HostStatus::Global::OK
+      when NOT_MANAGED_WITH_DATA
+        ::HostStatus::Global::WARN
```
ares, line 35: I was thinking about this but always come back to the conclusion, warning is probably the right value. While users may deploy insights-client through other means (ansible role, puppet module, manually) the parameter says it should not be deployed so users should either change the param or be warned about that. So :+1:

ares (commented):
looks good to me, just a suggestion to distinguish the states human representations

ShimShtein (approved):

ares:
When we'll be ready to merge this? It should be in for the release that goes with Foreman 2.5.

laviro:
@ShimShtein can we merge this now?

ShimShtein:
@laviro, I want to branch 2.3 out before we merge it.

app/subscribers/foreman_rh_cloud/insights_subscriber.rb
```diff
@@ -0,0 +1,8 @@
+module ForemanRhCloud
+  class InsightsSubscriber < ::Foreman::BaseSubscriber
+    def call(*args)
+      host = args.first.payload[:object]
+      host.get_status(InsightsClientReportStatus).refresh!
```
ShimShtein, line 5: This causes non-managed hosts with `host_registration_insights` param to be displayed as "Reporting" even though the host is not provisioned at all.

ShimShtein (changes requested):
Apparently there are some functionality glitches

app/controllers/insights_cloud/api/machine_telemetries_controller.rb
```diff
@@ -69,5 +70,14 @@ def ensure_branch_id
       @branch_id = cp_owner_id(@organization)
       return render_message "Branch ID not found for organization #{@organization.title}", :status => 400 unless @branch_id
     end
+
+    def update_host_insights_status
+      return unless request.path == '/redhat_access/r/insights/platform/ingress/v1/upload'
```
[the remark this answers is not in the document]
stejskalleos: Could you please give me info how to reproduce it?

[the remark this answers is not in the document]
ShimShtein: I think it's the compatibility=false setting

stejskalleos:
@ShimShtein 
Rebased, updated & fixed both issues (status for new host & legacy upload by insights client)

stejskalleos:
@ShimShtein updated

This is current behavior:
host_registration_insights | sending data? | Insights status (code)
-- | -- | --
`nil` | yes | `NOT_MANAGED_WITH_DATA (3)`
`nil` | no | `NOT_MANAGED (2)`
`TRUE` | yes | `REPORTING (0)`
`TRUE` | no | `NO_REPORT (1)`
`FALSE` | yes | `NOT_MANAGED_WITH_DATA (3)`
`FALSE` | no | `NOT_MANAGED (2)`

app/controllers/insights_cloud/api/machine_telemetries_controller.rb
```diff
@@ -69,5 +70,14 @@ def ensure_branch_id
       @branch_id = cp_owner_id(@organization)
       return render_message "Branch ID not found for organization #{@organization.title}", :status => 400 unless @branch_id
     end
+
+    def update_host_insights_status
+      return unless request.path == '/redhat_access/r/insights/platform/ingress/v1/upload' ||
+                    request.path.include?('/redhat_access/r/insights/uploads/')
+
+      host_status = @host.get_status(InsightsClientReportStatus)
+      host_status.update(reported_at: DateTime.now)
```
ShimShtein, line 79: Instead of updating the database twice, you can update the in-memory record and then save it.

app/subscribers/foreman_rh_cloud/insights_subscriber.rb
```diff
@@ -0,0 +1,9 @@
+module ForemanRhCloud
+  class InsightsSubscriber < ::Foreman::BaseSubscriber
+    def call(*args)
+      host = args.first.payload[:object]
+      host_status = host.get_status(InsightsClientReportStatus)
+      host_status.update(status: host_status.to_status)
```
ShimShtein, line 6: I don't mind checking in as it is, but there is `host.refresh_statuses([InsightsClientReportStatus])` method that would do the same.

ShimShtein (changes requested):
One small comment and we're ready to go

app/subscribers/foreman_rh_cloud/insights_subscriber.rb
stejskalleos: I'd like to keep it as it is because `refresh_statuses` also refresh timestamps which is something I don't want in this step.

ShimShtein:
Finally merged. Thanks, @stejskalleos!

PR #426, opened 2020-12-22
Add empty states to insights page

MariaAga:
Depends on: https://github.com/theforeman/foreman_rh_cloud/pull/421
Added the following empty states:
- Loading: when the table pr will be available I'll change the condition to: `status === 'PENDING' && !data.length`
Screenshot from 2020-12-22 16-53-24
- Error: shows the error that was received
Screenshot from 2020-12-22 16-53-34

- No token: if the token length is less than 1 then the user is shown instructions on how to add a token:
Screenshot from 2020-12-22 16-49-47

Also changed the default value for the token setting to be empty

### DIFF
--- a/app/controllers/insights_cloud/hits_controller.rb
+++ b/app/controllers/insights_cloud/hits_controller.rb
@@ -7,6 +7,7 @@ module InsightsCloud
 
       render json: {
         hits: hits,
+        hasToken: Setting[:rh_cloud_token].length > 1,
       }, status: :ok
     end
 
--- a/app/models/setting/rh_cloud.rb
+++ b/app/models/setting/rh_cloud.rb
@@ -1,4 +1,6 @@
 class Setting::RhCloud < Setting
+  ::Setting::BLANK_ATTRS.concat %w{rh_cloud_token}
+
   def self.default_settings
     return unless ActiveRecord::Base.connection.table_exists?('settings')
     return unless super
@@ -7,7 +9,7 @@ class Setting::RhCloud < Setting
       set('allow_auto_insights_sync', N_('Allow recommendations synchronization from Red Hat cloud'), false),
       set('obfuscate_inventory_hostnames', N_('Obfuscate host names sent to Red Hat cloud'), false),
       set('obfuscate_inventory_ips', N_('Obfuscate ip addresses sent to Red Hat cloud'), false),
-      set('rh_cloud_token', N_('Authentication token to Red Hat cloud services. Used to authenticate requests to cloud APIs'), 'DEFAULT', N_('Red Hat Cloud token'), nil, encrypted: true),
+      set('rh_cloud_token', N_('Authentication token to Red Hat cloud services. Used to authenticate requests to cloud APIs'), nil, N_('Red Hat Cloud token'), nil, encrypted: true),
       set('exclude_installed_packages', N_('Exclude installed packages from Red Hat cloud inventory report'), false),
     ]
   end
--- a/package.json
+++ b/package.json
@@ -38,7 +38,8 @@
     "raf": "^3.4.0",
     "stylelint": "^9.3.0",
     "stylelint-config-standard": "^18.0.0",
-    "surge": "^0.20.3"
+    "surge": "^0.20.3",
+    "redux-mock-store": "^1.2.2"
   },
   "dependencies": {
     "jed": "^1.1.1",
--- a/webpack/InsightsCloudSync/Components/ErrorEmptyState.js
+++ b/webpack/InsightsCloudSync/Components/ErrorEmptyState.js
@@ -1,0 +1,30 @@
+import React from 'react';
+import { useSelector } from 'react-redux';
+import {
+  EmptyState,
+  EmptyStateIcon,
+  EmptyStateVariant,
+  EmptyStateBody,
+} from '@patternfly/react-core';
+import { ExclamationCircleIcon } from '@patternfly/react-icons';
+import { translate as __, sprintf } from 'foremanReact/common/I18n';
+import { selectError } from '../InsightsCloudSyncSelectors';
+
+export const ErrorEmptyState = () => {
+  const error = useSelector(selectError);
+  return (
+    <EmptyState variant={EmptyStateVariant.xl}>
+      <EmptyStateIcon icon={ExclamationCircleIcon} />
+      <EmptyStateBody>
+        <p>
+          {sprintf(
+            __(
+              'Something went wrong while getting the data, the server returned the following error: %s'
+            ),
+            error
+          )}
+        </p>
+      </EmptyStateBody>
+    </EmptyState>
+  );
+};
--- a/webpack/InsightsCloudSync/Components/InsightsHeader/index.js
+++ b/webpack/InsightsCloudSync/Components/InsightsHeader/index.js
@@ -1,38 +1,14 @@
 import React from 'react';
-import { Icon } from 'patternfly-react';
 import { translate as __ } from 'foremanReact/common/I18n';
 import InsightsSettings from '../InsightsSettings';
-import { cloudTokenSettingUrl } from '../../InsightsCloudSyncHelpers';
+import './InsightsHeader.scss';
 
 const InsightsHeader = () => (
-  <div className="insights-header" style={{ marginBottom: '15px' }}>
+  <div className="insights-header">
     <InsightsSettings />
     <p>
       {__(`Insights synchronization process is used to provide Insights
              recommendations output for hosts managed here`)}
-    </p>
-    <p>
-      {__(`1. Obtain an Red Hat API token: `)}
-      <a
-        href="https://access.redhat.com/management/api"
-        target="_blank"
-        rel="noopener noreferrer"
-      >
-        access.redhat.com <Icon name="external-link" />
-      </a>
-      <br />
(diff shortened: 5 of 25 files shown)


webpack/InsightsCloudSync/Components/InsightsHeader/index.js
```diff
@@ -0,0 +1,21 @@
+import React from 'react';
+import { translate as __ } from 'foremanReact/common/I18n';
+import InsightsSettings from '../InsightsSettings';
+
+const InsightsHeader = () => (
+  <div
+    className="insights-header"
+    style={{
+      marginBottom: '15px',
+      overflow: 'auto',
```
MariaAga, line 10: Need to move to scss file

app/models/setting/rh_cloud.rb
```diff
@@ -1,4 +1,6 @@
 class Setting::RhCloud < Setting
+  ::Setting::BLANK_ATTRS.concat %w{rh_cloud_token}
```
laviro, line 2: what is it actually doing?

app/models/setting/rh_cloud.rb
```diff
@@ -7,7 +9,7 @@ def self.default_settings
       set('allow_auto_insights_sync', N_('Allow recommendations synchronization from Red Hat cloud'), false),
       set('obfuscate_inventory_hostnames', N_('Obfuscate host names sent to Red Hat cloud'), false),
       set('obfuscate_inventory_ips', N_('Obfuscate ip addresses sent to Red Hat cloud'), false),
-      set('rh_cloud_token', N_('Authentication token to Red Hat cloud services. Used to authenticate requests to cloud APIs'), 'DEFAULT', N_('Red Hat Cloud token'), nil, encrypted: true),
+      set('rh_cloud_token', N_('Authentication token to Red Hat cloud services. Used to authenticate requests to cloud APIs'), nil, N_('Red Hat Cloud token'), nil, encrypted: true),
```
laviro, line 12: Do we need a migration for that? @ShimShtein ?

package.json
```diff
@@ -38,7 +38,8 @@
     "raf": "^3.4.0",
     "stylelint": "^9.3.0",
     "stylelint-config-standard": "^18.0.0",
-    "surge": "^0.20.3"
+    "surge": "^0.20.3",
+    "redux-mock-store": "^1.2.2"
```
laviro, line 42: I think it would be better to use `theforeman/test` package or an alternative than introducing another npm package that doesn't exist in packaging 2.3

webpack/InsightsCloudSync/Components/ErrorEmptyState.js
```diff
@@ -0,0 +1,30 @@
...
+export const ErrorEmptyState = () => {
+  const error = useSelector(selectError);
+  return (
+    <EmptyState variant={EmptyStateVariant.xl}>
+      <EmptyStateIcon icon={ExclamationCircleIcon} />
+      <EmptyStateBody>
+        <p>
+          {sprintf(
+            __(
+              'Something went wrong while getting the data, the server returned the following error: %s'
```
laviro, line 22: Is it a network error? maybe we can just say "Network error, looks like something went wrong, please try to refresh the page"

webpack/InsightsCloudSync/Components/NoTokenEmptyState.js
```diff
@@ -0,0 +1,86 @@
...
+  const onSave = () => {
+    dispatch(
+      put({
+        key: SETTINGS_API_KEY,
+        url: settingsUrl,
+        params: { setting: { value: token } },
+        handleSuccess: reloadData,
+      })
+    );
+  };
```
laviro, line 34: nice work! though Ithink it should be a `POST` to `hits_controller` which will save the setting and trigger the sync task.

app/models/setting/rh_cloud.rb
MariaAga: Allowing the setting to be empty

app/models/setting/rh_cloud.rb
MariaAga: When I added `::Setting::BLANK_ATTRS.concat %w{rh_cloud_token}` I didn't need a migration for it to get the new default

webpack/InsightsCloudSync/Components/ErrorEmptyState.js
MariaAga: For example the error that was fixed in the table PR (undefined taxonamy_conditions). It's not a network error so refreshing won't fix it

package.json
MariaAga: @theforeman/packaging This package is also used in foreman 2.3. Is it a problem that it doesn't exist in packaging 2.3 if it's only used in testing?

evgeni: If it's for tests only, then we don't need it packaged, just added to the exclude list (if it's not there yet)

MariaAga: https://github.com/theforeman/foreman-packaging/blob/rpm/develop/update-requirements#L37
Looks like it's there

evgeni: Then you're safe from the packaging 👻!

laviro: Awesome!

webpack/InsightsCloudSync/Components/NoTokenEmptyState.js
MariaAga: Added a path and a function for that

evgeni: Totally of topic, but when I've seen this, I wondered: would there be an interest (and need) for having a "real" api client in JS, that can build those paths etc automatically? Like we have for Ruby and Python?

ShimShtein:
@MariaAga, can you please rebase this?

webpack/InsightsCloudSync/Components/NoTokenEmptyState.js
ShimShtein: > Totally of topic, but when I've seen this, I wondered: would there be an interest (and need) for having a "real" api client in JS, that can build those paths etc automatically? Like we have for Ruby and Python?

I think you're talking about something like this one: https://github.com/railsware/js-routes.
Are you sure you want the packaging hassle of it?

app/controllers/insights_cloud/hits_controller.rb
```diff
@@ -22,6 +23,14 @@ def model_of_controller
       ::InsightsHit
     end
 
+    def save_setting_and_sync
+      token = Setting::RhCloud.find_by_name("rh_cloud_token")
+      token.value = params[:value]
+      token.save
```
ShimShtein, line 29: Let's use the version that throws exception on validation.
```suggestion
      token.save!
```

app/controllers/insights_cloud/hits_controller.rb
```diff
@@ -22,6 +23,14 @@ def model_of_controller
       ::InsightsHit
     end
 
+    def save_setting_and_sync
+      token = Setting::RhCloud.find_by_name("rh_cloud_token")
+      token.value = params[:value]
```
ShimShtein, line 28: To have proper exception thrown
```suggestion
      token.value = params.require(:value)
```

app/controllers/insights_cloud/hits_controller.rb
```diff
@@ -22,6 +23,14 @@ def model_of_controller
       ::InsightsHit
     end
 
+    def save_setting_and_sync
```
ShimShtein, line 26:
```suggestion
    def save_token_and_sync
```
I think it adds clarity.

Besides that, I think the proper place for this would be the settings controller, and the response could do a redirect to `insights_cloud_hits_url`

app/controllers/insights_cloud/hits_controller.rb
```diff
@@ -7,6 +7,7 @@ def index
 
       render json: {
         hits: hits,
+        isToken: Setting[:rh_cloud_token].length > 1,
```
ShimShtein, line 10:
```suggestion
        hasToken: Setting[:rh_cloud_token].length > 1,
```
I think it describes the state, rather than what the response is.

webpack/InsightsCloudSync/InsightsCloudSync.js
```diff
@@ -53,12 +90,16 @@ InsightsCloudSync.propTypes = {
   page: PropTypes.number,
   perPage: PropTypes.number,
   query: PropTypes.string,
+  isToken: PropTypes.bool,
```
ShimShtein, line 93:
```suggestion
  hasToken: PropTypes.bool,
```

ShimShtein (changes requested):

app/controllers/insights_cloud/hits_controller.rb
MariaAga: Where would the sync fit in this way?

ShimShtein: I think it will still fit to the settings page. Imagine we want other processes start immediately after saving the token.

In later stages, we will refactor the direct call to the sync task into a separate object (maybe a dynflow task) that will be responsible for handling all "state initialization" process.

MariaAga:
Thanks Shim for the suggestions. I updated the commit

ShimShtein:
Merged, thanks @MariaAga!

PR #428, opened 2020-12-22
Move inventory upload to react router

MariaAga:
Had to make some css changes because after I changed to `PageLayout` it looked less good. Changed the class names to fit the convention :)
Screenshot from 2020-12-22 19-09-36

### DIFF
--- a/webpack/ForemanInventoryUpload/Components/PageHeader/PageHeader.scss
+++ b/webpack/ForemanInventoryUpload/Components/PageHeader/PageHeader.scss
@@ -1,0 +1,3 @@
+.inventory-upload-header {
+  overflow: auto;
+}
--- a/app/views/foreman_rh_cloud/react/inventory_upload.html.erb
+++ b/app/views/foreman_rh_cloud/react/inventory_upload.html.erb
@@ -1,1 +1,0 @@
-<%= react_component('ForemanInventoryUpload') %>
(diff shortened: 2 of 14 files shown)


webpack/ForemanInventoryUpload/Components/PageHeader/PageHeader.scss
```diff
@@ -0,0 +1,7 @@
+.inventory_title {
```
ShimShtein, line 1: Do we still need it? I don't see you use it anywhere

ShimShtein (changes requested):

webpack/ForemanInventoryUpload/Components/PageHeader/PageHeader.scss
MariaAga: You're right, it's not used, deleted :tada:

ShimShtein (changes requested):
the contents table got smaller, is it intentional?
After PR:
Screenshot from 2020-12-27 15-06-28
Before PR:
Screenshot from 2020-12-27 15-09-11

MariaAga:
@ShimShtein No, but I'm changing the design in the next pr (https://github.com/theforeman/foreman_rh_cloud/pull/429), so should I fix it here?

ShimShtein:
I'm OK with not having it fixed here, as long as we make sure it's fixed in the next one.

PR #429, opened 2020-12-27
Inventory upload setting redesign to pf4

MariaAga:
Depends on https://github.com/theforeman/foreman_rh_cloud/pull/428
Starting the redesign of  the settings. Next I will need to add "red hat inventory" and "cloud connected"  settings and move the other settings to a modal.
Screenshot from 2020-12-27 12-39-18

### DIFF
--- a/webpack/ForemanInventoryUpload/Components/AutoUploadSwitcher/AutoUploadSwitcher.js
+++ b/webpack/ForemanInventoryUpload/Components/AutoUploadSwitcher/AutoUploadSwitcher.js
@@ -1,20 +1,17 @@
 import React from 'react';
 import PropTypes from 'prop-types';
-import { Switch, FieldLevelHelp } from 'patternfly-react';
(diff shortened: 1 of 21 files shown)


ShimShtein:
@MariaAga, can you rebase please?

MariaAga:
Rebased, if you have some suggestions on the design I would love to hear them :smile:

ShimShtein (changes requested):
image
Can we make them not so close to each other?

Also the main "card view" is a bit smaller than the screen. Can we get it back to the same width?

MariaAga:
I removed the grid as it's not necessary any more and it added the padding. About spacing the settings: I don't see them like this, can you check if you have this css rule applied? `.foreman-switcher{ margin-bottom: 5px; }`

MariaAga:
I made the css rule more specific and added a classname to the page so we only change the plugins css

ShimShtein:
Merged, thanks @MariaAga!

PR #430, opened 2020-12-27
Avoid host base including RhCloudHost module

laviro:
There are some errors related to the fact we want to use the managed host and not the base host.


ShimShtein:
Merged, thanks @laviro!

PR #433, opened 2020-12-28
Add Ruby tests to GH actions

laviro:


laviro:
@ShimShtein added ruby tests, took 18 min without cache, after we merge this I guess it will reduce..

PR #438, opened 2020-12-30
Change compression algo for reports to xz

ShimShtein:
the default gzip files become too large, let's change it to xz that should compress the files much more efficiently.

This can go in once https://github.com/quipucords/yupana/pull/336 goes in.

### DIFF
--- a/lib/foreman_inventory_upload.rb
+++ b/lib/foreman_inventory_upload.rb
@@ -48,7 +48,7 @@ module ForemanInventoryUpload
   end
 
   def self.facts_archive_name(organization)
-    "report_for_#{organization}.tar.gz"
+    "report_for_#{organization}.tar.xz"
   end
 
   def self.upload_url
--- a/lib/foreman_inventory_upload/generators/archived_report.rb
+++ b/lib/foreman_inventory_upload/generators/archived_report.rb
@@ -28,7 +28,7 @@ module ForemanInventoryUpload
 
           @logger.info 'Archiving generated report'
           # success = system('tar', '-zcvf', @target, '-C', tmpdir, '.')
-          Open3.popen2e('tar', '-zcvf', @target, '-C', tmpdir, '.') do |_in, out, wait_thr|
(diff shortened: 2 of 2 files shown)


laviro (approved):
Thanks @ShimShtein ! LGTM, waiting on the yupana PR :+1:

adamruzicka:
While you're touching this, wouldn't zstd give even better results?

ShimShtein:
> While you're touching this, wouldn't zstd give even better results?

Depends on yupana, if they can read it, I don't mind. @kgaikwad what do you think?

kgaikwad:
@ShimShtein @adamruzicka 
Yupana currently uses [tarfile](https://docs.python.org/3/library/tarfile.html) module that reads and writes gzip, bz2 and lzma compressed archives. 
So on Yupana side, it needs to change a library that supports `zstd`. Example - [xtarfile](https://pypi.org/project/xtarfile/)

gnurag:
Objectively speaking, `zstd` does perform better with our reports when we consider overall compression ratio, cpu utilization and time taken. 

```
$ ls -lh
-rw-rw-r--. 1 anurag anurag 102M Jan  4 18:06 report-gzip.tar.gz     # 33.360s
-rw-rw-r--. 1 anurag anurag 459M Jan  4 17:07 report.tar             # original
-rw-rw-r--. 1 anurag anurag 3.5M Jan  4 17:55 report-xz.tar.xz       # 56.000s
-rw-rw-r--. 1 anurag anurag 7.3M Jan  4 17:55 report-zstd.tar.zst    # 1.916s
```
Thanks @kgaikwad for looking into library support. `xtarfile` seems to be a wrapper over python `tarfile`, so this is something we can explore too at some point in the future.

gnurag:
I'd vote for keeping `xz` as the preferred algo, since our primary aim is to reduce the uploaded archive size.

ShimShtein:
Thanks for the awesome analysis! So let's wait for yupana to go online, and then we can merge this PR.

PR #440, opened 2020-12-31
Move inventory toolbar buttons to title

MariaAga:
In `inventory_upload`, move the "Actions history" and "Documentation" links / buttons from the toolbar to a kebab next to the title.
This can also probably be done using `customBreadcrumbs` option in the page layout but there's a bug that makes it unusable in foreman 
2.3 (Solved here: https://github.com/theforeman/foreman/pull/8217)
The dropdown items appear on the left of the icon instead of the right (which is the default) so they will be shown inside of the page limits

Screenshot from 2020-12-31 11-40-18


MariaAga:
I'm not sure what to do with the failing ruby tests, the file that fails is mocked and the command doesn't fail in my env

ShimShtein:
Now the tests should work as expected. Can you rebase please?

MariaAga:
Still fail with `ModuleNotFoundError: Module not found: Error: Can't resolve 'foremanReact/components/Head' in '/home/runner/work/foreman_rh_cloud/foreman_rh_cloud/projects/foreman_rh_cloud/webpack/ForemanInventoryUpload/Components/PageHeader'`

ShimShtein:
Thanks to @laviro, found the reason: in Foreman 2.3 there is no `Head` component, but you are asking for it: https://github.com/theforeman/foreman_rh_cloud/pull/440/checks?check_run_id=1639285226#step:20:61

MariaAga:
Oh I see, good test :)
I didn't realize how old 2.3 is. Removed the `Head`

ShimShtein:
Merged, thanks @MariaAga!

PR #451, opened 2021-01-20
Add backend for calling cloud connector setup

ehelms:


webpack/ForemanInventoryUpload/Components/AutoUploadSwitcher/AutoUploadSwitcherActions.js
```diff
@@ -8,7 +8,7 @@ export const handleToggle = currentAutoUploadEnabled => async dispatch => {
   try {
     const {
       data: { autoUploadEnabled },
-    } = await API.post(inventoryUrl('auto_upload'), {
+    } = await API.post(inventoryUrl('enable_cloud_connector'), {
```
ehelms, line 11: Ignore this, will need to be reverted. This was simply to make testing for me easier at the start.

app/services/foreman_rh_cloud/cloud_connector.rb
```diff
@@ -0,0 +1,57 @@
...
+
+    def service_user
+      user = User.find_by_login(CLOUD_CONNECTOR_USER)
+
+      if user.nil?
+        user = User.create!(
+          :login => CLOUD_CONNECTOR_USER,
+          :password => SecureRandom.base64(15),
+          :description => "This is a service user used by cloud connector to talk to the Satellite API",
+          :auth_source => AuthSource.find_by_name('Internal'),
```
ares, line 44: How about `AuthSourceInternal.first`?

ares (commented):
nice, few comments from reading but generally thumbs up

app/services/foreman_rh_cloud/cloud_connector.rb
```diff
@@ -0,0 +1,57 @@
...
+      user = service_user
+      token_value = personal_access_token(user)
+
+      composer = ::JobInvocationComposer.for_feature(
+        'ansible_configure_cloud_connector',
+        [foreman_host.id],
+        {:satellite_user => service_user.login, :satellite_password => token_value}
+      )
+      composer.trigger!
+      composer.job_invocation
```
ShimShtein, line 19: From UI point of view we will want to show that the job is still running. For that, we will need to fetch the correct job invocation from the UI, where we don't have the ID. Is there a way we can identify the correct invocation?

adamruzicka: `compose.job_invocation` should be the job invocation that was created, so it is just a matter of serializing it into json in the controller.

ShimShtein: @adamruzicka, in the current browser session it will be there, sure. What about refresh? I need a way to find the invocation with fresh browser (redux store)

adamruzicka: That should be a responsibility of that part of the frontend code which calls this api, no?

config/routes.rb
```diff
@@ -11,6 +11,9 @@
     post 'installed_packages_inclusion', to: 'uploads#installed_packages_inclusion'
     post 'ips_obfuscation', to: 'uploads#ips_obfuscation'
 
+    get 'cloud_connector', to: 'uploads#cloud_connector_status'
+    post 'cloud_connector', to: 'uploads#enable_cloud_connector'
```
ehelms, line 15: @MariaAga I updated the routes here. The POST route should kick things off and return a task. The GET should return the most recent job.

app/controllers/foreman_inventory_upload/uploads_controller.rb
```diff
@@ -24,6 +24,21 @@ def throw_flash_error(message)
...
 
+    def enable_cloud_connector
+      cloud_connector = ForemanRhCloud::CloudConnector.new
+      cloud_connector.install
+    end
+
+    def cloud_connector_status
+      cloud_connector = ForemanRhCloud::CloudConnector.new
+
+      if (job = cloud_connector.running_job)
```
ares, line 35: Why can't we always return the job? Even if it's running, we have the link to the running job and not the task. I'm probably missing something. But `running_job` seems to either return `nil` or the job itself. Why do we load task from it if it's job already?

app/services/foreman_rh_cloud/cloud_connector.rb
```diff
@@ -0,0 +1,68 @@
...
+    end
+
+    def latest_job
+      feature_id = RemoteExecutionFeature.find_by_label(CLOUD_CONNECTOR_FEATURE).id
+      JobInvocation.where(:remote_execution_feature_id => feature_id).first
+    end
+
+    def running_job
+      return unless latest_job.status == HostStatus::ExecutionStatus::RUNNING
```
MariaAga, line 29: Can you please change this to `latest_job&.status`?

app/controllers/foreman_inventory_upload/uploads_controller.rb
```diff
@@ -24,6 +24,21 @@ def throw_flash_error(message)
...
+        job.task
+      else
+        cloud_connector.latest_job
```
MariaAga, line 38: Also can you change this to :
```
else
        cloud_connector.latest_job
elif
 render json something
 ```
 When cloud_connector.latest_job is null the server returns an error for me

File: app/controllers/foreman_inventory_upload/uploads_controller.rb (hunk 5 above)
Comment (ehelms) [reply app/controllers/foreman_inventory_upload/uploads_controller.rb:35]: I was not sure how the UI would know from the returned data here that the job was still executing. Are you saying return the job and have the UI then query the Jobs API for status?

File: app/services/foreman_rh_cloud/cloud_connector.rb [hunk 8]
```diff
@@ -0,0 +1,68 @@
...
+    CLOUD_CONNECTOR_USER = 'cloud_connector_user'.freeze
+    CLOUD_CONNECTOR_TOKEN_NAME = 'Cloud connector access token'.freeze
+    CLOUD_CONNECTOR_FEATURE = 'ansible_configure_cloud_connector'.freeze
+
+    def install
+      user = service_user
+      token_value = personal_access_token(user)
+
+      composer = ::JobInvocationComposer.for_feature(
+        CLOUD_CONNECTORE_FEATURE,
```
MariaAga, line 15: Should be CLOUD_CONNECTOR_FEATURE instead of CLOUD_CONNECTOR**E**_FEATURE ( without an E in CONNECTOR)

ehelms:
Pushed a few updates, the status endpoint now returns the job as json in all cases. For now, this means querying the Job API for status after acquiring the job.

laviro (commented):
Left some comments about the UI part,
as for the server-side, I couldn't test it yet as I still need to add smart-proxy to my dev environment

webpack/ForemanInventoryUpload/Components/PageHeader/components/CloudConnectorButton/CloudConnectorActions.js
```diff
@@ -0,0 +1,9 @@
+import { post } from 'foremanReact/redux/API';
+import { CONFIGURE_CLOUD_CONNECTOR } from './CloudConnectorConstants';
+import { inventoryUrl } from '../../../../ForemanInventoryHelpers';
+
+export const configureCloudConnector = () =>
+  post({
+    key: CONFIGURE_CLOUD_CONNECTOR,
+    url: inventoryUrl('enable_cloud_connector'),
+  });
```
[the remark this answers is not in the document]
MariaAga: @laviro maybe Instead of `${response.data.jobID}` it should be the url to the job/task?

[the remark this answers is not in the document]
laviro: yeah, sure.. it was just an example :D

app/controllers/foreman_inventory_upload/uploads_controller.rb
ares: I think I'm saying return the job and the job status in the custom json response. User shouldn't be punished by our insufficient APIs, we should give them a link to the job instead of a task and their browser shouldn't need to do 2 queries for that in my opinion.

MariaAga: Does this makes sense?
https://github.com/theforeman/foreman_rh_cloud/blob/80e08cd4a3ed2bace536e4acc4343c983f0667aa/app/controllers/foreman_inventory_upload/uploads_controller.rb#L32-L35
https://github.com/theforeman/foreman_rh_cloud/blob/80e08cd4a3ed2bace536e4acc4343c983f0667aa/webpack/ForemanInventoryUpload/Components/PageHeader/components/CloudConnectorButton/CloudConnectorSelectors.js#L8-L12
from my PR https://github.com/theforeman/foreman_rh_cloud/pull/458

ehelms:
Closing given this was included https://github.com/theforeman/foreman_rh_cloud/pull/458

app/controllers/foreman_inventory_upload/uploads_controller.rb
ares: yeah, I definitely mind the implementation less, e.g. we can use task and we can do extra API query to fetch the job, it just feels suboptimal since we already have all of that info on the server side

PR #452, opened 2021-01-24
Add select all to url in insights

MariaAga:
Insights table will be opened with selectAll= true from the navigation.
un-selecting all/one of the hosts will set selectAll =false in the url.
url params can't be boolean so the selector is ```isSelectAll === true ||  isSelectAll === 'true'; ``` because `!!'false' === true`.
Missing action tests.

### DIFF
--- a/webpack/InsightsCloudSync/Components/InsightsTable/InsightsTableActions.js
+++ b/webpack/InsightsCloudSync/Components/InsightsTable/InsightsTableActions.js
@@ -1,10 +1,7 @@
 import URI from 'urijs';
 import { push } from 'connected-react-router';
 import { get } from 'foremanReact/redux/API';
-import {
-  selectIsAllSelected,
-  selectQueryParams,
-} from './InsightsTableSelectors';
+import { selectQueryParams } from './InsightsTableSelectors';
 import { INSIGHTS_PATH } from '../../InsightsCloudSyncConstants';
 import {
   columns,
@@ -17,7 +14,7 @@ import {
 
 export const fetchInsights = (queryParams = {}) => (dispatch, getState) => {
   const state = getState();
-  const { page, perPage, query, sortBy, sortOrder } = {
+  const { page, perPage, query, sortBy, sortOrder, isSelectAll } = {
     ...selectQueryParams(state),
     ...queryParams,
   };
@@ -32,6 +29,12 @@ export const fetchInsights = (queryParams = {}) => (dispatch, getState) => {
         search: query,
         order: `${sortBy} ${sortOrder}`,
       },
+      handleSuccess: response => {
+        if (isSelectAll) {
+          selectAllIds(dispatch, response.data.hits || []);
+          dispatch(selectAll());
+        }
+      },
     })
   );
 
@@ -42,6 +45,7 @@ export const fetchInsights = (queryParams = {}) => (dispatch, getState) => {
     search: query,
     sort_by: sortBy,
     sort_order: sortOrder,
+    select_all: isSelectAll,
   });
 
   dispatch(
@@ -51,13 +55,19 @@ export const fetchInsights = (queryParams = {}) => (dispatch, getState) => {
     })
   );
 
-  const isAllSelected = selectIsAllSelected(state);
-  if (isAllSelected) {
-    dispatch(clearAllSelection());
-  } else {
+  if (!isSelectAll) {
     dispatch(setSelectAllAlert(false));
-    dispatch(setSelectAll(false));
   }
+};
+
+const selectAllIds = (dispatch, results, prevSelectedIds = {}) => {
+  const selectedIds = { ...prevSelectedIds };
+  results.forEach(row => {
+    if (row.disableCheckbox) return;
+    selectedIds[row.id] = true;
+  });
+  dispatch(selectByIds(selectedIds));
+  dispatch(setSelectAllAlert(true));
 };
 
 export const setSelectAllAlert = showSelectAllAlert => ({
@@ -70,10 +80,13 @@ export const selectByIds = selectedIds => ({
   payload: { selectedIds },
 });
 
-export const setSelectAll = isAllSelected => ({
-  type: INSIGHTS_SET_SELECT_ALL,
-  payload: { isAllSelected },
-});
+export const setSelectAll = isAllSelected => dispatch => {
+  dispatch(setSelectAllUrl(isAllSelected));
+  dispatch({
+    type: INSIGHTS_SET_SELECT_ALL,
+    payload: { isAllSelected },
+  });
+};
 
 export const selectAll = () => setSelectAll(true);
 
@@ -100,31 +113,42 @@ export const onTablePerPageSelect = (_event, perPageNumber) =>
   fetchInsights({ perPage: perPageNumber });
 
 export const onTableSelect = (
-  _event,
   isSelected,
   rowId,
   rows,
   prevSelectedIds
 ) => dispatch => {
-  const selectedIds = { ...prevSelectedIds };
-  let showSelectAllAlert = false;
-  // for select all
+  const handleRegularCheckbox = () => {
+    const selectedIds = { ...prevSelectedIds };
+    if (isSelected) {
+      selectedIds[rows[rowId].id] = true;
+    } else {
+      dispatch(setSelectAllAlert(false));
+      dispatch(setSelectAll(false));
+      delete selectedIds[rows[rowId].id];
+    }
+
+    dispatch(selectByIds(selectedIds));
+  };
+
   if (rowId === -1) {
     if (!isSelected) return dispatch(clearAllSelection());
-
-    rows.forEach(row => {
-      if (row.disableCheckbox) return;
-      selectedIds[row.id] = true;
-    });
-
-    showSelectAllAlert = true;
+    selectAllIds(dispatch, rows, prevSelectedIds);
   } else {
-    isSelected
-      ? (selectedIds[rows[rowId].id] = true)
-      : delete selectedIds[rows[rowId].id];
+    handleRegularCheckbox();
   }
 
-  dispatch(selectByIds(selectedIds));
-  dispatch(setSelectAllAlert(showSelectAllAlert));
   return null;
+};
+
+const setSelectAllUrl = selectAllValue => dispatch => {
+  const uri = new URI();
+  uri.setSearch({ select_all: selectAllValue });
+
+  dispatch(
+    push({
+      pathname: INSIGHTS_PATH,
+      search: uri.search(),
+    })
+  );
 };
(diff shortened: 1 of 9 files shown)


lib/foreman_rh_cloud/engine.rb
```diff
@@ -66,7 +66,7 @@ class Engine < ::Rails::Engine
         # Adding a sub menu after hosts menu
         divider :top_menu, caption: N_('RH Cloud'), parent: :configure_menu
         menu :top_menu, :inventory_upload, caption: N_('Inventory Upload'), url: '/foreman_rh_cloud/inventory_upload', url_hash: { controller: :react, action: :index }, parent: :configure_menu
-        menu :top_menu, :insights_hits, caption: N_('Insights'), url: '/foreman_rh_cloud/insights_cloud', url_hash: { controller: :react, action: :index }, parent: :configure_menu
+        menu :top_menu, :insights_hits, caption: N_('Insights'), url: '/foreman_rh_cloud/insights_cloud?isSelectAll=true', url_hash: { controller: :react, action: :index }, parent: :configure_menu
```
laviro, line 69: @ShimShtein do we want it to open the plugin page always with `selectAll=true` from the menu?
or just to apply `selectAll` from the `url` from specific locations such as `host view`?

ShimShtein: I suppose by default it should be `select_all = false`.

webpack/InsightsCloudSync/Components/InsightsTable/InsightsTableSelectors.js
```diff
@@ -64,6 +65,10 @@ export const selectSelectedIds = state =>
 export const selectIsAllSelected = state =>
   selectInsightsCloudTable(state).isAllSelected || false;
 
+export const selectIsAllSelectedQuery = state =>
+  selectQuery(state).isSelectAll === true ||
+  selectQuery(state).isSelectAll === 'true';
```
laviro, line 70: the `query` values are always strings,

```suggestion
export const selectIsAllSelectedQuery = state => selectQuery(state).isSelectAll === 'true';
```

laviro:
Thanks @MariaAga! can you also rebase?

webpack/InsightsCloudSync/Components/InsightsTable/InsightsTableActions.js
```diff
@@ -32,6 +29,11 @@ export const fetchInsights = (queryParams = {}) => (dispatch, getState) => {
         search: query,
         order: `${sortBy} ${sortOrder}`,
       },
+      handleSuccess: response => {
+        if (isSelectAll) {
+          selectAllIdsFromResonse(dispatch, response);
+        }
+      },
```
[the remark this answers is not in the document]
MariaAga: From what I understand I need to select all the results so all the checkboxes will be shown as selected, but the remediation will just send a query for select all.

webpack/InsightsCloudSync/Components/InsightsTable/InsightsTableActions.js
```diff
@@ -117,13 +132,29 @@ export const onTableSelect = (
...
+  const uri = new URI();
+  uri.setSearch({ isSelectAll: selectAllValue });
+
+  dispatch(
+    push({
+      pathname: INSIGHTS_PATH,
+      search: uri.search(),
+    })
+  );
+};
```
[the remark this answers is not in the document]
MariaAga: Do you mean to put 
``` js
  dispatch(
    push({
      pathname: INSIGHTS_PATH,
      search: uri.search(),
    })
  );
```
in a function?

MariaAga:
This PR has the wrong commit

MariaAga:
I think I was able to recover the commit :partying_face:

webpack/InsightsCloudSync/Components/InsightsTable/InsightsTableActions.js
```diff
@@ -51,15 +54,24 @@ export const fetchInsights = (queryParams = {}) => (dispatch, getState) => {
...
+  const selectedIds = {};
+  const results = response?.data?.hits || [];
+
+  dispatch(selectAll());
+  results.forEach(row => {
+    selectedIds[row.id] = true;
+  });
+  dispatch(selectByIds(selectedIds));
+  dispatch(setSelectAllAlert(true));
+  dispatch(selectAll());
```
ShimShtein, line 72: Why do you need to dispatch `selectAll` twice?

MariaAga: I don't, removed

webpack/InsightsCloudSync/Components/InsightsTable/InsightsTableActions.js
```diff
@@ -17,7 +14,7 @@ import {
 
 export const fetchInsights = (queryParams = {}) => (dispatch, getState) => {
   const state = getState();
-  const { page, perPage, query, sortBy, sortOrder } = {
+  const { page, perPage, query, sortBy, sortOrder, isSelectAll = false } = {
```
laviro, line 17: No need to set it here as the selector will always return a boolean value
```suggestion
  const { page, perPage, query, sortBy, sortOrder, isSelectAll } = {
```

webpack/InsightsCloudSync/Components/InsightsTable/InsightsTableActions.js
```diff
@@ -42,6 +44,7 @@ export const fetchInsights = (queryParams = {}) => (dispatch, getState) => {
     search: query,
     sort_by: sortBy,
     sort_order: sortOrder,
+    isSelectAll,
```
laviro, line 47: can you use the snake_case convention we have for the other params?

```suggestion
    select_all: isSelectAll,
```

webpack/InsightsCloudSync/Components/InsightsTable/InsightsTableActions.js
```diff
@@ -51,15 +54,23 @@ export const fetchInsights = (queryParams = {}) => (dispatch, getState) => {
...
     dispatch(clearAllSelection());
-  } else {
-    dispatch(setSelectAllAlert(false));
-    dispatch(setSelectAll(false));
   }
 };
 
+const selectAllIdsFromResonse = (dispatch, response) => {
+  const selectedIds = {};
+  const results = response?.data?.hits || [];
```
laviro, line 64: no need, because on success we will get the response.data object in any case
```suggestion
  const results = response.data.hits || [];
```

webpack/InsightsCloudSync/Components/InsightsTable/InsightsTableActions.js
```diff
@@ -117,13 +131,29 @@ export const onTableSelect = (
...
+    selectedIds[results[rowId].id] = true;
+    if (results.length === Object.keys(selectedIds).length) {
+      showSelectAllAlert = true;
+    }
   } else {
-    isSelected
-      ? (selectedIds[results[rowId].id] = true)
-      : delete selectedIds[results[rowId].id];
+    dispatch(setSelectAll(false));
+    delete selectedIds[results[rowId].id];
```
laviro, line 141: did you mean to put it inside the following `else` on line 139?

webpack/InsightsCloudSync/Components/InsightsTable/InsightsTableActions.js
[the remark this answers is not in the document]
laviro: I thought about using the same method in `fetchInsights` and here to update the URL, but can be added as a follow-up

webpack/InsightsCloudSync/Components/InsightsTable/InsightsTableActions.js
MariaAga: Sure I'll change to that, do you have any resource where I can read about that? :)

webpack/InsightsCloudSync/Components/InsightsTable/InsightsTableActions.js
MariaAga: What do you mean?

webpack/InsightsCloudSync/Components/InsightsTable/InsightsTableActions.js
laviro: that's what Axios and the API middleware will return on success,
what's in the data object is what returned from the server in the render JSON,
and in this case, if hits are undefined it will return an empty array

webpack/InsightsCloudSync/Components/InsightsTable/InsightsTableActions.js
laviro: I mean that in line `126` we check if the main checkbox was selected and in line `139` else we refer to all other rows checkboxes. in each of the if's context, you can do whatever you want, but do you really need the new `else if`?

MariaAga: The new else if replaces the condition that was in the else so
``` javascript
else {
   isSelected? dispatch(setSelectAll(false)) : (selectedIds[results[rowId].id] = true)
}
```
is now 
``` javascript
else if(isSelected) {
   dispatch(setSelectAll(false))
   ...
} else {
   (selectedIds[results[rowId].id] = true)
   ...
}
```

MariaAga:
Changed `selectAllFromResopnse` to `SelectAllIds` so now the fetch and the `onTableSelect` can use it.

webpack/InsightsCloudSync/Components/InsightsTable/InsightsTableActions.js
MariaAga: Moved the else code to a function for readability

laviro (commented):
nice work @MariaAga !
found a bit of changed behavior:

Before, the user selects all the current viewable items, unless they press to select all others
Red Hat Insights (4)

Now, it selects all from other pages too immediatley.
Red Hat Insights (3)

MariaAga:
Moved dispatch(selectAll());
to the handle success instead of being in `selectAllIds`

laviro (commented):
Now it seems that if you select individual recommendations,
they will be completely gone if you switch between the table pages:

Red Hat Insights (5)

MariaAga:
@laviro Since we want this functionality I changed the alert text to say `%s Recommendations are selected` so for example, per 
 page is 5 , and user selects one recommendation in page 1 and then select all page on page 2 and the alert will show `6  Recommendations are selected` instead of `all 6 in this page are selected

laviro (commented):
tested and it works great!
one last thing before merging, can you add a translation for a singular case when one recommendation is selected?

Screenshot (63)

MariaAga:
@laviro Changed to `Recommendations selected: X `

laviro (commented):
Thanks @MariaAga @ShimShtein,
waiting on the tests to go green before merging

PR #454, opened 2021-01-26
[CP 2.1] Change compression algo for reports to xz (#438)

ShimShtein:


ShimShtein:
@adamruzicka, Can you take it for a spin? Should be simple enough to test.

adamruzicka (approved):
Haven't really tested it, but `man tar` confirms. ACK

PR #459, opened 2021-01-27
Remove empty ipv6 elements

ShimShtein:


ShimShtein:
@adamruzicka, would be nice to test this one

lib/foreman_inventory_upload/generators/slice.rb
```diff
@@ -105,8 +105,8 @@ def report_system_profile(host, host_ips_cache)
         @stream.array_field('network_interfaces') do
           @stream.raw(host.interfaces.map do |nic|
             {
-              'ipv4_addresses': [host_ips_cache[nic.ip]].compact,
-              'ipv6_addresses': [nic.ip6].compact,
+              'ipv4_addresses': [host_ips_cache[nic.ip]].reject(&:empty?),
+              'ipv6_addresses': [nic.ip6].reject(&:empty?),
```
adamruzicka, line 109: How would I go about getting a nic where the address would be an empty string like you have in tests?

adamruzicka (commented):
In general I'm in favor, I'm just curious how we can get into a situation where this is needed.

lib/foreman_inventory_upload/generators/slice.rb
ShimShtein: I suppose it comes from hosts that are created in foreman from imported facts. We have a customer that we have seen this behaviour. I can get more details, if you like.

adamruzicka: Please do

adamruzicka (approved):
ACK

ShimShtein:
@laviro, ready for review/merge

laviro (approved):

laviro:
Thanks @ShimShtein !

PR #472, opened 2021-02-07
Update gh-action lock files after cloud connector was added

laviro:


ShimShtein:
It took 23 mins previously. Let's see how long it will take with warm cache

ShimShtein:
Now it's 17 min. Merged, thanks @laviro!

PR #473, opened 2021-02-07
Set default autoupload value to true

ShimShtein:
@ezr-ondrej, does it look like a valid solution?

The behaviour I am looking for is for new users set the default to `true`.
For users that are upgrading from previous versions and were relying on the fact that the default was `false`, I want to set the value to `false` explicitly.

### DIFF
--- a/app/models/setting/rh_cloud.rb
+++ b/app/models/setting/rh_cloud.rb
@@ -1,11 +1,20 @@
 class Setting::RhCloud < Setting
   ::Setting::BLANK_ATTRS.concat %w{rh_cloud_token}
 
+  def self.load_defaults
+    return false unless table_exists?
+    transaction do
+      # If the user had auto_upload default setting, we will not surprise him, and force the value to false
+      # for new users, the default will be set to true and the value will remain nil
+      Setting.where(name: 'allow_auto_inventory_upload', value: nil).where("settings.default LIKE '%false%'").update_all(value: "--- false\n...")
+      super
+    end
+  end
+
   def self.default_settings
     return unless ActiveRecord::Base.connection.table_exists?('settings')
-    return unless super
     [
-      set('allow_auto_inventory_upload', N_('Allow automatic upload of the host inventory to the Red Hat cloud'), false, N_('Allow automatic inventory uploads')),
+      set('allow_auto_inventory_upload', N_('Allow automatic upload of the host inventory to the Red Hat cloud'), true, N_('Allow automatic inventory uploads')),
       set('allow_auto_insights_sync', N_('Allow recommendations synchronization from Red Hat cloud'), false, N_('Allow recommendations synchronization')),
       set('obfuscate_inventory_hostnames', N_('Obfuscate host names sent to Red Hat cloud'), false, N_('Obfuscate host names')),
       set('obfuscate_inventory_ips', N_('Obfuscate ip addresses sent to Red Hat cloud'), false, N_('Obfuscate IPs')),


app/models/setting/rh_cloud.rb
```diff
@@ -4,8 +4,11 @@ class Setting::RhCloud < Setting
   def self.default_settings
     return unless ActiveRecord::Base.connection.table_exists?('settings')
     return unless super
+    # If the user had auto_upload default setting, we will not surprise him, and force the value to false
+    # for new users, the default will be set to true and the value will remain nil
+    Setting.where(name: 'allow_auto_inventory_upload', default: "--- false\n...", value: nil).update_all(value: "--- false\n...")
```
ezr-ondrej, line 9: I think this would be better off in [`load_defaults`]( https://github.com/theforeman/foreman/blob/0112426cd294fda6fe65fd5f0fdb28d95024221b/app/models/setting.rb#L283).
`default_settings` shouldn't have side effects I believe.
Apart of that it looks reasonable :+1:  (read doesn't look great, but it's probably our only option xD)

app/models/setting/rh_cloud.rb
```diff
@@ -4,8 +4,11 @@ class Setting::RhCloud < Setting
   def self.default_settings
     return unless ActiveRecord::Base.connection.table_exists?('settings')
     return unless super
```
ezr-ondrej, line 6: BTW this has been moved to `load_defaults`, this is useless now as super here is always `[]`

ShimShtein:
@ezr-ondrej Ready for another round!

ezr-ondrej (approved):
Make sense to me :+1:

ShimShtein:
@laviro, final round?

laviro (commented):
Thanks @ShimShtein,
I tested this two cases:
- User who didn't touch the setting before this change,
   the value stayed `nil` but the `default` result to `true` after this change
-  When the user had set it to `off` prior to this change - seems like it still updates the setting to `true`
- And if it was `true` it remains `true`

ShimShtein:
@laviro, what I intended to happen is like this:
first, the default should be set to `true` in all cases (no matter the record was there or not and no matter the user changed its value or not).
Now if the user has change the value through foreman's UI, that value will remain as it was (it's the one that is stored in the value column in the DB).
Now if the setting is new for the user (as in the record was not in the database before - new install), the default would be set to `true`, hence the user will see it as `true`.
If the record was already existing, and its default was `false`, but the user didn't touch its value (meaning the user was relying on the default value to be false), we will change the default to `true` as in all cases, but we will set the `value` to `false` similar to what happens when the user explicitly sets the value. This way the users that upgrade the system and were relying on the default to be false will have effective value remain false.

app/models/setting/rh_cloud.rb
```diff
@@ -1,11 +1,20 @@
 class Setting::RhCloud < Setting
   ::Setting::BLANK_ATTRS.concat %w{rh_cloud_token}
 
+  def self.load_defaults
+    return false unless table_exists?
+    transaction do
+      # If the user had auto_upload default setting, we will not surprise him, and force the value to false
+      # for new users, the default will be set to true and the value will remain nil
+      Setting.where(name: 'allow_auto_inventory_upload', default: "--- false\n...", value: nil).update_all(value: "--- false\n...")
```
laviro, line 9: I think I found the issue, the `\n` doesn't catch the default value:

```suggestion
      Setting.where(name: 'allow_auto_inventory_upload', default: "--- false", value: nil).update_all(value: "--- false\n...")
```

Selection_002

Btw, can't the default value be just `false`? why are the `---` additions needed?

ShimShtein: We're saving the values as yaml-encoded. As for the `\n` - I think it's encoded with `\n`, take a look at the value after the change, if it's `true\n`, then it was `false\n` beforehand.

laviro: @ShimShtein when switching with your PR, after seed to default `true` it also shows without the `/n`:
Selection_003

but looks like in older versions of Foreman, like 2.1-stable, it has `default: "--- false\n..."` in it, @ezr-ondrej any idea? :D

laviro (commented):
Thanks @ShimShtein !

now it works,
- from {default: false, value: nil} to {default: true, value: false}
- from {default: false, value: false} to {default: true, value: false}

PR #477, opened 2021-02-09
Fix inventory file download error handling

laviro:
Since we moved to react-router pages, there is a problem to throw an error doing a full-page reload.
In this PR we will get the file and the error with a JSON request.

### DIFF
--- a/app/controllers/foreman_inventory_upload/uploads_controller.rb
+++ b/app/controllers/foreman_inventory_upload/uploads_controller.rb
@@ -14,14 +14,7 @@ module ForemanInventoryUpload
       files = Dir["{#{ForemanInventoryUpload.uploads_file_path(filename)},#{ForemanInventoryUpload.done_file_path(filename)}}"]
 
       return send_file files.first, disposition: 'attachment', filename: filename unless files.empty?
-      throw_flash_error "File doesn't exist"
-    end
-
-    def throw_flash_error(message)
-      process_error(
-        :redirect => foreman_inventory_upload_index_path,
-        :error_msg => message
-      )
+      raise ::Foreman::Exception.new("The report file doesn't exist")
     end
 
     def enable_cloud_connector
--- a/webpack/ForemanInventoryUpload/Components/Dashboard/DashboardActions.js
+++ b/webpack/ForemanInventoryUpload/Components/Dashboard/DashboardActions.js
@@ -67,7 +67,7 @@ export const setActiveTab = (accountID, tabName) => ({
 });
 
 export const downloadReports = accountID => {
-  window.location.href = inventoryUrl(`${accountID}/uploads/file`);
+  window.open(inventoryUrl(`${accountID}/uploads/file`), '_blank');
   return {
     type: INVENTORY_REPORTS_DOWNLOAD,
     payload: {
--- a/webpack/ForemanInventoryUpload/Components/Dashboard/__tests__/DashboardActions.test.js
+++ b/webpack/ForemanInventoryUpload/Components/Dashboard/__tests__/DashboardActions.test.js
@@ -36,15 +36,15 @@ const fixtures = {
 };
 
 describe('Dashboard actions', () => {
-  const { location } = window;
+  const { open } = window;
 
   beforeAll(() => {
-    delete window.location;
-    window.location = { href: jest.fn() };
+    delete window.open;
+    window.open = jest.fn();
   });
 
   afterAll(() => {
-    window.location = location;
+    window.open = open;
   });
 
   return testActionSnapshotWithFixtures(fixtures);


laviro:
On the second commit, I added the file from the response as a `Blob` in one API call,
unlike the first commit where there are two API calls.

@ShimShtein @MariaAga what are your thoughts about this?

webpack/ForemanInventoryUpload/Components/Dashboard/DashboardActions.js
```diff
@@ -66,16 +71,47 @@ export const setActiveTab = (accountID, tabName) => ({
...
-    toastError: error => error.response.message,
-    handleSuccess: () => {
-      window.location.href = url;
-    },
-  });
+export const downloadReports = accountID => async dispatch => {
+  dispatch({ type: REPORT_DOWNLOAD_REQUEST, payload: { accountID } });
+  try {
+    const filename = `report_for_${accountID}.tar.xz`;
+    const { data } = await Axios.get(
```
MariaAga, line 78: `webpack/assets/javascripts/react_app/redux/API/API.js` contains a wrap for axios, can you use it?

webpack/ForemanInventoryUpload/Components/Dashboard/DashboardActions.js
```diff
@@ -66,16 +71,47 @@ export const setActiveTab = (accountID, tabName) => ({
...
+    // IE variant
+    if (window.navigator && window.navigator.msSaveOrOpenBlob) {
+      window.navigator.msSaveOrOpenBlob(blob, filename);
+    } else {
+      const url = window.URL.createObjectURL(blob);
+      const link = document.createElement('a');
+      link.href = url;
+      link.setAttribute('download', filename);
+      document.body.appendChild(link);
+      link.click();
```
MariaAga, line 97: I wonder if you can use `import { saveAs } from 'file-saver';` here instead, foreman uses it in `TemplateGeneratorActions.js`

app/controllers/foreman_inventory_upload/uploads_controller.rb
```diff
@@ -13,15 +13,8 @@ def download_file
...
-      throw_flash_error "File doesn't exist"
-    end
-
-    def throw_flash_error(message)
-      process_error(
-        :redirect => foreman_inventory_upload_index_path,
-        :error_msg => message
-      )
+      return render json: { message: "File doesn't exist" }, status: :not_found if files.empty?
+      render :json => File.read(files.first)
```
ShimShtein, line 17: I have a better solution: Let's leave the browser to handle the download in a separate page.
If there will be an error, instead of rendering a flash error, we will render a 404, or a specific "file not found" page.

ShimShtein (changes requested):

app/controllers/foreman_inventory_upload/uploads_controller.rb
laviro: After all of the hacks? :angel: just kidding,
I like this approach too,
and I decided to go with `raise ActionController::BadRequest.new("The report file doesn't exist")`:
Screenshot (50)

instead of just a plain empty page with a small message:
Screenshot (51)

app/controllers/foreman_inventory_upload/uploads_controller.rb
```diff
@@ -14,14 +14,7 @@ def download_file
...
       return send_file files.first, disposition: 'attachment', filename: filename unless files.empty?
-      throw_flash_error "File doesn't exist"
-    end
-
-    def throw_flash_error(message)
-      process_error(
-        :redirect => foreman_inventory_upload_index_path,
-        :error_msg => message
-      )
+      raise ActionController::BadRequest.new("The report file doesn't exist")
```
ShimShtein, line 17: Better to use either `::Foreman::Exception` or `::Foreman::WrappedException`. I think they are shown in a better way, plus we have a way to mark them in documentation.

ShimShtein (changes requested):

app/controllers/foreman_inventory_upload/uploads_controller.rb
laviro: updated, thanks @ShimShtein !

app/controllers/foreman_inventory_upload/uploads_controller.rb
```diff
@@ -14,14 +14,7 @@ def download_file
...
       return send_file files.first, disposition: 'attachment', filename: filename unless files.empty?
-      throw_flash_error "File doesn't exist"
-    end
-
-    def throw_flash_error(message)
-      process_error(
-        :redirect => foreman_inventory_upload_index_path,
-        :error_msg => message
-      )
+      raise ::Foreman::Exception.new("The report file doesn't exist")
```
ShimShtein, line 17: We should probably use `not_found` method from [`ApplicationController`](https://github.com/theforeman/foreman/blob/develop/app/controllers/application_controller.rb#L105) with the exception:

```suggestion
      return not_found ::Foreman::Exception.new("The report file doesn't exist")
```

ShimShtein (changes requested):
Can you also move the link to be opened in a new tab, otherwise on error it will overwrite the main tab.

app/controllers/foreman_inventory_upload/uploads_controller.rb
laviro: `return not_found` ends up with a page like this one, looks like the warning is based on the `controller`:

Selection_001

I think the `::Foreman::Exception` is better, have more info in it:

Screenshot (58)

laviro:
Thanks Shim, now if there's an error it will be opened in another tab

ShimShtein:
Let's start with this solution and change it later if needed

PR #479, opened 2021-02-16
refactor cloud token with boolean value

laviro:


laviro:
@ShimShtein a small fix we talked about,
@MariaAga while you are working on the settings refactor you can rebase on top of this PR

app/controllers/foreman_inventory_upload/accounts_controller.rb
```diff
@@ -24,7 +24,7 @@ def index
         autoUploadEnabled: Setting[:allow_auto_inventory_upload],
         hostObfuscationEnabled: Setting[:obfuscate_inventory_hostnames],
         ipsObfuscationEnabled: Setting[:obfuscate_inventory_ips],
-        cloudToken: Setting[:rh_cloud_token],
+        cloudToken: Setting[:rh_cloud_token].length > 1,
```
MariaAga, line 27: Suggestion: `cloudToken: Setting[:rh_cloud_token] && Setting[:rh_cloud_token].length > 1`
When I got a new env it was set to null so this gave me an error, `Setting[:rh_cloud_token]?.length > 1` didn't work because ruby doesn't want to compare null to 1

laviro: Using `!Setting[:rh_cloud_token].empty?` to catch both `nil` and empty strings `''`

MariaAga (approved):

ShimShtein:
Merged, thanks @laviro!

PR #481, opened 2021-02-16
Add insights has playbook column

laviro:


laviro:
@ShimShtein does the idea match your backend work?

webpack/InsightsCloudSync/Components/InsightsTable/SelectAllAlert.js
```diff
@@ -25,7 +24,7 @@ const SelectAllAlert = ({
         )}
         actionLinks={
           <AlertActionLink onClick={selectAll}>
-            {sprintf('Select all %s recommendations', itemCount)}
+            {__('Select recommendations from all pages')}
```
laviro, line 27: We decided to drop the actual number of all recommendations, also on other strings here
as we can't really tell if they are selectable.

also, for the `selectAll` functionality which updates the store with the `flag`,
the exact number doesn't really matter.

MariaAga:
Can the table show the disabled rows as selected with `disableSelection` so that the select all at the top of the page will work? this can be needed for example if the user has 24 rows with playbook, 1 without and they click on the select all, all rows will be selected but the user won't have a way to unselect all of them because the select all checkbox is not checked

MariaAga (approved):
Works with and without playbooks :tada: 
Thanks!

ShimShtein:
Merged. Thanks @laviro!

PR #482, opened 2021-02-17
Fix 500 on insights page when token is empty

ShimShtein:


app/controllers/insights_cloud/hits_controller.rb
```diff
@@ -6,7 +6,7 @@ def index
       hits = resource_base_search_and_page.preload(:host)
 
       render json: {
-        hasToken: Setting[:rh_cloud_token].length > 1,
+        hasToken: !Setting[:rh_cloud_token].empty?,
```
laviro, line 9: also fixed in #479 :)

PR #483, opened 2021-02-17
Add rule and resolution models

ShimShtein:


lib/insights_cloud/async/insights_full_sync.rb
```diff
@@ -79,8 +95,19 @@ def to_model_hash(hit_hash)
           total_risk: hit_hash['total_risk'].to_i,
           likelihood: hit_hash['likelihood'].to_i,
           results_url: hit_hash['results_url'],
+          rule_id: to_rule_id(hit_hash['results_url']),
```
laviro, line 98: can we also add `has_playbook` data here based on the `rule_id`?

app/models/insights_hit.rb
```diff
@@ -4,6 +4,9 @@ class InsightsHit < ApplicationRecord
   # since the facet is one-to-one association with a host, we can connect
   # through host_id column on both this model and facet.
   belongs_to :insights_facet, foreign_key: 'host_id', primary_key: 'host_id', counter_cache: :hits_count
+
+  has_one :rule, class_name: 'InsightsRule', foreign_key: 'rule_id', primary_key: 'rule_id'
```
laviro, line 8: can you add `rule_id` to `scoped_search` and maybe `has_playbook` too?

laviro (commented):
NIce work @ShimShtein! left a few suggestions

laviro:
Can you please rebase?

lib/insights_cloud/async/insights_full_sync.rb
ShimShtein: At this point I don't know which rules have playbooks, since I didn't fetch recommendations yet.

ShimShtein:
@laviro, rebased and ready for another round.
Added scoped search for `with_playbook=true` plus the obvious `rule_id` field.
Now you can also use `InsightsHit.with_playbook` too. I have added a scope for that.

laviro (commented):
Thanks @ShimShtein, tested and it works great!
as we discussed offline, as a follow up we could add:
- sort by `has_playbook`
- scoped search for `has_playbook = false`

PR #485, opened 2021-02-22
Bump version to 3.0.17

ShimShtein:


laviro (commented):
Thanks @ShimShtein, can you also update [package.json version](https://github.com/theforeman/foreman_rh_cloud/blob/master/package.json#L3)

PR #489, opened 2021-02-25
Add tags to inventory_export_url

laviro:
returning all of the hosts from c.r.c can take a lot of time,
we can filter to the specific hosts with the satellite instance tag


ShimShtein:
LGTM, merged!

PR #490, opened 2021-02-25
Refactor Inventory full-sync task to be async

laviro:
Now the task is async:

https___10 46 9 103_foreman_rh_cloud_inventory_upload (1)


app/controllers/foreman_inventory_upload/tasks_controller.rb
```diff
@@ -13,12 +13,21 @@ def create
...
 
       render json: {
-        syncHosts: host_statuses[:sync],
-        disconnectHosts: host_statuses[:disconnect],
+        task: task,
+      }, status: :ok
+    end
+
+    def index
+      task = ForemanTasks::Task.where(label: 'InventorySync::Async::InventoryFullSync')
```
ShimShtein, line 25: `task` can be nil, if it was never executed.
It will fail later in the `render` statement

lib/foreman_inventory_upload.rb
```diff
@@ -62,6 +62,7 @@ def self.ensure_folder(folder)
   end
 
   def self.inventory_export_url
-    ForemanRhCloud.base_url + '/api/inventory/v1/hosts'
+    tags = URI.encode("satellite/satellite_instance_id=#{Foreman.instance_id}")
+    ForemanRhCloud.base_url + "/api/inventory/v1/hosts?tags=#{tags}"
```
ShimShtein, line 66: You can rebase this

ShimShtein (changes requested):

app/controllers/foreman_inventory_upload/tasks_controller.rb
laviro: I think if it's `nil` we can return status `404 :not_found` to the client
and we hide the chart/show some empty state for "run inventory sync"

laviro:
a bit more work is needed

laviro:
ready for a review @ShimShtein @adamruzicka

app/controllers/foreman_inventory_upload/tasks_controller.rb
```diff
@@ -13,12 +13,23 @@ def create
...
       render json: {
-        syncHosts: host_statuses[:sync],
-        disconnectHosts: host_statuses[:disconnect],
+        task: task,
+      }, status: :ok
+    end
+
+    def show
+      task = ForemanTasks::Task.where(id: params[:id]).first
+      return render json: { message: N_('No task was found') }, status: :internal_server_error unless task || task.id
```
adamruzicka, line 27: `task || task.id` if task is `nil`, call `.id` on it? Shouldn't there be `&&`?

app/controllers/foreman_inventory_upload/uploads_settings_controller.rb
```diff
@@ -22,5 +23,12 @@ def self.cloud_connector_status
       return nil unless job
       { id: job.id, task: ForemanTasks::Task.where(:id => job.task_id).first }
     end
+
+    def last_successful_inventory_sync_task
+      task = ForemanTasks::Task.where(label: 'InventorySync::Async::InventoryFullSync', result: 'success')
+                        .reorder('ended_at desc').first
```
adamruzicka, line 29: Is the `reorder` necessary? Why use `.where(...).first` instead of `find_by(...)`?

ShimShtein (changes requested):

app/controllers/foreman_inventory_upload/uploads_settings_controller.rb
adamruzicka: `ForemanTasks::Task.reorder('ended_at desc').find_by(...)` should work. The query it generates is identical to `where(...).reorder(...).first` so it doesn't really matter. Nevermind then

app/controllers/foreman_inventory_upload/tasks_controller.rb
laviro: updated

laviro:
Thanks! ready for another round :)

webpack/ForemanInventoryUpload/Components/PageHeader/components/SyncButton/SyncButtonActions.js
```diff
@@ -1,57 +1,92 @@
...
-    );
-  } catch ({
-    message,
-    response: { data: { message: toastMessage } = {} } = {},
-  }) {
-    dispatch({
-      type: INVENTORY_SYNC_ERROR,
-      payload: {
-        error: message,
+export const handleSync = () => dispatch => {
```
laviro, line 14: @MariaAga can you also take a look at this part? thanks

MariaAga (commented):
The code looks good but I can't check the functionality, I think the tests could be a bit better so I left some comments :)

webpack/ForemanInventoryUpload/Components/PageHeader/components/SyncButton/__tests__/__snapshots__/SyncButtonSelectors.test.js.snap
```diff
@@ -1,18 +1,3 @@
...
-`;
-
-exports[`SyncButton selectors should return InventorySync SyncHosts 1`] = `1`;
-
-exports[`SyncButton selectors should return InventorySync disconnectHosts 1`] = `0`;
-
-exports[`SyncButton selectors should return InventorySync error 1`] = `"some-error"`;
-
-exports[`SyncButton selectors should return InventorySync status 1`] = `"RESOLVED"`;
+exports[`SyncButton selectors should return InventorySync status 1`] = `null`;
```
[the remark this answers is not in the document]
laviro: fixed

webpack/ForemanInventoryUpload/Components/PageHeader/components/SyncButton/__tests__/__snapshots__/SyncButtonActions.test.js.snap
```diff
@@ -4,32 +4,11 @@ exports[`SyncButton actions should handleSync 1`] = `
...
-          "sticky": true,
-          "type": "success",
-        },
-      },
-      "type": "TOASTS_ADD",
+      "errorToast": [Function],
+      "handleSuccess": [Function],
+      "key": "INVENTORY_SYNC",
+      "type": "post-some-type",
+      "url": "/foreman_inventory_upload/tasks",
```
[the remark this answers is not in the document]
laviro: basically, you are right...
unfortunately, with the current integration test environment, I couldn't mimic the full cycle of the API middleware
even though I mocked the whole foreman redux folder, checkout: https://github.com/laviro/foreman_rh_cloud/blob/wip/inventory_fullsync_integration_test

when I used the integration helper snapshot, I got only the initial `pending` state.. https://github.com/laviro/foreman_rh_cloud/blob/wip/inventory_fullsync_integration_test/webpack/ForemanInventoryUpload/Components/PageHeader/components/SyncButton/__tests__/__snapshots__/integration.test.js.snap#L25

I think we should setup something with foreman and `react-testing-library` for this to actually work properly in tests, any other thoughts @ShimShtein @MariaAga ?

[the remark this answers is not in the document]
MariaAga: I sent Ron an example on how I did it in REX to check if it fits here:
https://github.com/theforeman/foreman_remote_execution/blob/7260e7da180e1792e0f70efecc09f0d9db0fa0f1/webpack/react_app/components/JobWizard/__tests__/integration.test.js#L17

laviro:
the last commit adds the integration test, but there is more to cover
thanks for the mocking tip @MariaAga!

ShimShtein:
Merged, thanks @laviro!

PR #491, opened 2021-03-01
Sync rh_cloud tags with branch_info

ShimShtein:
Now the tags that are reported by rh_cloud are identical to those reported by insights-client (via the `branch_info` method)


lphiri (approved):
@ShimShtein  - this looks good to me. Its exactly as I had imagined we would do it :).

laviro (approved):

laviro:
Thanks @ShimShtein @lphiri !!

PR #495, opened 2021-03-02
handle accounts by label nor id

laviro:


ShimShtein:
Merged, thanks @laviro!

PR #496, opened 2021-03-02
avoid Inventory page components scss overriding other pages

laviro:


laviro:
@ShimShtein @MariaAga can you check if I missed anything?
looks like our scss overrides other foreman/plugins pages as we load it with deface for 2.3
added a wrapping class in all of the inventory components scss

for insights and the host-details-tab pages looks like they are already explicit,

to review this PR it's best to view it in unified mode and hide spaces so changes are reduced to ~30:
Screenshot (72)

laviro:
@adamruzicka that's the one we were talking about,
can you check also if it solves your issue?

ShimShtein (changes requested):
Found two more files that need attention:
 - [x] `./webpack/ForemanInventoryUpload/Components/PageHeader/components/SyncButton/components/modal.scss` I would rename to something less generic: `#inventory_sync_modal` for example

 - [x] `./webpack/ForemanInventoryUpload/Components/TabHeader/tabHeader.scss` needs wrapping

MariaAga:
Also:

- [ ] webpack/ForemanInventoryUpload/Components/InventoryFilter/inventoryFilter.scss
- [ ] webpack/ForemanInventoryUpload/Components/InventorySettings/InventorySettings.scss
- [ ] webpack/ForemanInventoryUpload/Components/PageHeader/components/ToolbarButtons/toolbarButtons.scss

Found by searching `.` with files to include: `**/webpack/ForemanInventoryUpload/**/*.scss` in vscode

laviro:
> Also:
> 
> * [ ]  webpack/ForemanInventoryUpload/Components/InventoryFilter/inventoryFilter.scss
> * [ ]  webpack/ForemanInventoryUpload/Components/InventorySettings/InventorySettings.scss
> * [ ]  webpack/ForemanInventoryUpload/Components/PageHeader/components/ToolbarButtons/toolbarButtons.scss
> 
> Found by searching `.` with files to include: `**/webpack/ForemanInventoryUpload/**/*.scss` in vscode

I think those are actually specific enough as they all start with `.inventory-`,
from now on I think when we can be more specific in class names/ids that's enough I guess,
I just didn't want to break all of the other pages before the release so I found the solution of wrapping them with the page's class to be the most secure.

ShimShtein (approved):

ShimShtein:
Merged, thanks @laviro!

PR #497, opened 2021-03-02
Do not set taxonomy when enabling cloud connector

ehelms:
Given the host that represents Foreman can only exist in one organization,
if a user has set a different taxonomy on the uploads page the
cloud connector will fail with a 404. The code needs to ignore
taxonomies when running cloud connector to ensure that it runs
no matter the context the user is in.

### DIFF
--- a/app/services/foreman_rh_cloud/cloud_connector.rb
+++ b/app/services/foreman_rh_cloud/cloud_connector.rb
@@ -9,13 +9,18 @@ module ForemanRhCloud
     def install
       user = service_user
       token_value = personal_access_token(user)
+      target_host = foreman_host
+      composer = nil
 
-      composer = ::JobInvocationComposer.for_feature(
-        CLOUD_CONNECTOR_FEATURE,
-        [foreman_host.id],
-        {:satellite_user => service_user.login, :satellite_password => token_value}
-      )
-      composer.trigger!
+      Taxonomy.as_taxonomy(target_host.organization, target_host.location) do
+        composer = ::JobInvocationComposer.for_feature(
+          CLOUD_CONNECTOR_FEATURE,
+          [target_host.id],
+          {:satellite_user => service_user.login, :satellite_password => token_value}
+        )
+        composer.trigger!
+      end
+
       composer.job_invocation
     end
 
@@ -54,7 +59,7 @@ module ForemanRhCloud
     end
 
     def foreman_host
-      ::Host.friendly.find(::SmartProxy.default_capsule.name)
+      ::Host.unscoped.friendly.find(::SmartProxy.default_capsule.name)
     end
   end
 end


ehelms:
This stems from the following BZ -- https://bugzilla.redhat.com/show_bug.cgi?id=1934190

ShimShtein (changes requested):
Although this one looks quite elegant, I would prefer a bit more specific approach:
We can use `as_taxonomy` to wrap the action in the proper context:
``` ruby
      Taxonomy.as_taxonomy(self_host.organization, self_host.location) do
        composer = ::JobInvocationComposer.for_feature(
          CLOUD_CONNECTOR_FEATURE,
          [self_host.id],
          {:satellite_user => service_user.login, :satellite_password => token_value}
        )
        composer.trigger!
      end
```
Plus we need to fetch the host object according to user permissions:

``` ruby
     def foreman_host
      ::Host.unscoped.authorized.friendly.find(::SmartProxy.default_capsule.name)
     end
```

@ehelms could you please update the PR?

ehelms:
> Plus we need to fetch the host object according to user permissions:
> 
> ```ruby
>      def foreman_host
>       ::Host.unscoped.authorized.friendly.find(::SmartProxy.default_capsule.name)
>      end
> ```

I don't think we want this with an authorized check here do we? I feel like that should be at the controller level whether the action can be initiated from the API.



> Although this one looks quite elegant, I would prefer a bit more specific approach:
> We can use `as_taxonomy` to wrap the action in the proper context:
> 
> ```ruby
>       Taxonomy.as_taxonomy(self_host.organization, self_host.location) do
>         composer = ::JobInvocationComposer.for_feature(
>           CLOUD_CONNECTOR_FEATURE,
>           [self_host.id],
>           {:satellite_user => service_user.login, :satellite_password => token_value}
>         )
>         composer.trigger!
>       end
> ```

What is `self_host`?

ShimShtein:
> I don't think we want this with an authorized check here do we?

If the user is not authorized for the host, I suppose it won't be good if he will be able to configure it.

> What is self_host?

it's `foreman_host` result. I have cached it, since we want it in multiple places. The name could be better, no doubt in that :slightly_smiling_face:

ShimShtein:
@ehelms could you rebase? We have already fixed the tests failure.

ehelms:
> > I don't think we want this with an authorized check here do we?
> 
> If the user is not authorized for the host, I suppose it won't be good if he will be able to configure it.

We should then also wrap the controller action in an authorize check and do not display the button on the UI page either. My assumption is, currently, about the only person who can do any of this operation is an admin? Is that fair?

ShimShtein:
Permissions are already set here: https://github.com/theforeman/foreman_rh_cloud/blob/master/lib/foreman_rh_cloud/engine.rb#L40. I suppose we have missed a permission for cloud connector action.
We do not hide UI elements, but makes sense to disable if the user doesn't have permissions. Let's do it gradually though.

> My assumption is, currently, about the only person who can do any of this operation is an admin? Is that fair?

By default it's only admins, true. Although there is an option to create custom roles and assign them to the user.

ehelms:
I'll send along updates in a day or two. I need to re-create a test environment to ensure the proposed changes will work.

ehelms:
Testing the proposed changes here things are not working for me when I switch to an org that the host is not a part of.

ehelms:
Updated and rebased

ShimShtein:
Merged, thanks @ehelms!

PR #499, opened 2021-03-03
use foreman-tasks lower than 4.0.0

laviro:


adamruzicka:
Do newer tasks break things for you?

laviro:
Newer tasks require foreman 2.4 and our master branch depends on 2.3

PR #500, opened 2021-03-03
Don't show hits for hosts that are filtered out by authorization

ShimShtein:


laviro (approved):
works great!! thanks @ShimShtein :)

PR #501, opened 2021-03-04
fix layout change org undefined error

laviro:
in foreman-2.3 org isn't passed in the layout action payload,
instead it got changed into `organization`


ShimShtein:
Merged, thanks @laviro!

PR #508, opened 2021-03-08
Fix inventory settings scss rule

laviro:


ShimShtein:
Much better, thanks @laviro!

PR #509, opened 2021-03-08
[CP 2.1] Escape string values in inventory json stream (#507)

ShimShtein:


laviro (approved):
tested and it works as expected! :)

PR #514, opened 2021-03-10
Add foreman 2.3 and latest test matrix

ShimShtein:


.github/workflows/ruby_tests.yml
```diff
@@ -75,16 +93,17 @@ jobs:
       - name: Install system dependencies
         run: |
           sudo apt-get update -yqq
-          sudo apt-get -yqq install libsystemd-dev make tar git ruby-libvirt build-essential patch ruby-dev zlib1g-dev liblzma-dev libxml2-dev libcurl4-openssl-dev libpq-dev qemu-kvm libvirt-daemon-system libvirt-dev nodejs
-      - name: Add foreman_rh_cloud and katello to Gemfile
+          sudo apt-get -yqq install libsystemd-dev make tar git ruby-libvirt build-essential patch ruby-dev zlib1g-dev liblzma-dev libxml2-dev libcurl4-openssl-dev libpq-dev qemu-kvm libvirt-daemon-system libvirt-dev nodejs libqpid-proton*
```
laviro, line 96: Maybe `python-qpid-proton` is also needed? or some other proton lib for ubuntu
```suggestion
          sudo apt-get -yqq install libsystemd-dev make tar git ruby-libvirt build-essential patch ruby-dev zlib1g-dev liblzma-dev libxml2-dev libcurl4-openssl-dev libpq-dev qemu-kvm libvirt-daemon-system libvirt-dev nodejs libqpid-proton* python-qpid-proton
```

ShimShtein: What is missing is proton's c libraries to build the gem. I think ubuntu currently does not have the correct version and that's why the build is failing.

Personally I think it would be easier to move to Jenkins or a centos container than to try and fix this build.

adamruzicka:
What's the status here?

PR #516, opened 2021-03-16
Fix initial filter term value

laviro:
move organization value updates into `useEffect` and using the layout selector.
Also, using the PF4 input now


ShimShtein:
Merged, thanks @laviro!

PR #517, opened 2021-03-16
Forward also content-type to insights-client forwarding code

ShimShtein:
To prevent 400 from the server on registration.


laviro (approved):

laviro:
Thanks @ShimShtein!

PR #529, opened 2021-04-04
Switch jobs to tasks

ShimShtein:


lib/insights_cloud/async/insights_full_sync.rb
```diff
@@ -2,17 +2,19 @@
 
 module InsightsCloud
   module Async
-    class InsightsFullSync < ::ApplicationJob
+    class InsightsFullSync < ::Actions::EntryAction
       include ::ForemanRhCloud::CloudAuth
 
-      def perform
+      def plan
```
laviro, line 8: You should also change the way the task is being triggered in https://github.com/theforeman/foreman_rh_cloud/blob/master/app/controllers/insights_cloud/tasks_controller.rb#L4

otherwise there's the following error:
```
NoMethodError (undefined method `perform_later' for InsightsCloud::Async::InsightsFullSync:Class):
```

laviro (commented):
Thanks, tested and it works as expected :)

do we also need to update [the rake task](https://github.com/theforeman/foreman_rh_cloud/blob/aedfc8041b987e95f26eb9e77d568cb56dae43cb/lib/tasks/rh_cloud_inventory.rake#L57) ?

ShimShtein:
I did update it: https://github.com/theforeman/foreman_rh_cloud/pull/529/files#diff-9ca8d337c7041b111f4dee13b8c59910e8ad72ca97ca82b0a3b845f68893555aR57

laviro (approved):
Oh, sorry about that - all good than, thanks!

PR #530, opened 2021-04-05
Fix and split resolutions

ShimShtein:
Contains also https://github.com/theforeman/foreman_rh_cloud/pull/529.

Adds a proper resolutions sync task, so the resolutions could be used in playbook generation.


ShimShtein:
@laviro rebased, now ready for review

laviro (approved):
All looks great, tested and it works as expected, no issues with migration :)

PR #531, opened 2021-04-06
Fix missing to_status method.

ShimShtein:
It causes https://github.com/theforeman/foreman/blob/f5bfce8bc5cab43fa66c362d40ee843e0783fe72/app/models/host/managed.rb#L856 to fail.


sbernhard (approved):
Make sense. Thanks.

laviro (approved):

laviro:
Thanks @ShimShtein !

PR #537, opened 2021-04-29
Limit organization size for auto upload

ShimShtein:


lib/foreman_inventory_upload.rb
```diff
@@ -57,10 +57,14 @@ def self.upload_url
   end
 
   def self.slice_size
-    # for testing set ENV to 'https://ci.cloud.redhat.com/api/ingress/v1/upload'
     @slice_size ||= (ENV['SATELLITE_INVENTORY_SLICE_SIZE'] || '1000').to_i
   end
 
+  def self.max_org_size
+    # Set max amount of hosts per organization for automatic uploads
+    @max_org_size ||= (ENV['SATELLITE_INVENTORY_MAX_ORG_SIZE'] || '150000').to_i
```
adamruzicka, line 65: The default can be an integer and `#to_i` is then just a noop

lib/foreman_inventory_upload/async/generate_all_reports_job.rb
```diff
@@ -13,7 +13,13 @@ def perform
         organizations = Organization.unscoped.all
 
         organizations.map do |organization|
```
adamruzicka, line 15: This previously returned an array of active jobs and now returns an array where elements can be either active jobs or `nil`s. Is the result used anywhere? If not, could we switch `#map` for `#each`?

adamruzicka (commented):
Two nitpicks, otherwise lgtm

lib/foreman_inventory_upload/async/generate_all_reports_job.rb
ShimShtein: The method anyway was returning the `ensure` block. I think the `map` could be replaced with `each` anyway.

ShimShtein:
@adamruzicka, ready for another round

lib/foreman_inventory_upload/async/generate_all_reports_job.rb
adamruzicka: I always thought `ensure` blocks get evaluated last, but do not alter the return value

ShimShtein: Yep, you're right. `ensure` does not return value.

Well, removed `nil` values. Now the return value will filter out `nil` values.

ShimShtein:
@adamruzicka, ready for another round.

adamruzicka (approved):

laviro (approved):
Well, good to know that someone uses Foreman with that amount of hosts :D

laviro:
Thanks @ShimShtein @adamruzicka !

PR #538, opened 2021-04-29
Add missing translations

laviro:


ShimShtein:
does it require a specific version of foreman?

laviro:
> does it require a specific version of foreman?

nope, it is using another library as a `devDependency` so we can use it with which version we'd like :)

laviro:
for some reason, I can't disable the specific rule for files...
getting the error: `Definition for rule 'translation/no-strings-without-translations' was not found  translation/no-strings-without-translations`

adamruzicka:
What's the status here?

laviro:
Hey @adamruzicka when I will have more capacity I will update fix the eslint lib structure so no errors like the above will appear,
until then, I can make this PR to just add the translations where needed. any thoughts @ShimShtein ?

laviro:
@ShimShtein added also the translations for the table titles and dropped the eslint-plugin that was used,
as there's a bit more work needed on it and will probably be added to foreman-js
ready to review :)

ShimShtein (changes requested):
I think I have found a couple of skipped places:

https://github.com/theforeman/foreman_rh_cloud/blob/fd4c1a2e01a59835daa3444171b2a03802f6fc39/webpack/InsightsCloudSync/Components/InsightsTable/SelectAllAlert.js#L21

https://github.com/theforeman/foreman_rh_cloud/blob/master/webpack/ForemanInventoryUpload/Components/StatusChart/StatusChart.js#L9 and a couple of other instances in the same file

https://github.com/theforeman/foreman_rh_cloud/blob/fd4c1a2e01a59835daa3444171b2a03802f6fc39/webpack/InsightsHostDetailsTab/components/ListItem/ListItem.js#L22 and a couple of instances in the same file

ShimShtein:
Merged, thanks @laviro!

PR #539, opened 2021-04-29
Add proxy port fix

ShimShtein:
Override default behavior of URI class to hide default port for a schema.

### DIFF
--- a/lib/foreman_rh_cloud.rb
+++ b/lib/foreman_rh_cloud.rb
@@ -38,10 +38,23 @@ module ForemanRhCloud
   end
 
   def self.proxy_setting(logger: Foreman::Logging.logger('background'))
+    fix_port(proxy_string(logger: logger))
+  end
+
+  def self.proxy_string(logger: Foreman::Logging.logger('background'))
     HttpProxy.default_global_content_proxy&.full_url ||
     ForemanRhCloud.cdn_proxy(logger: logger) ||
     ForemanRhCloud.global_foreman_proxy ||
     ''
+  end
+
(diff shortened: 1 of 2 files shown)


lib/foreman_rh_cloud.rb
```diff
@@ -38,12 +38,25 @@ def self.transformed_http_proxy_string(logger: Foreman::Logging.logger('backgrou
...
     ForemanRhCloud.cdn_proxy(logger: logger) ||
     ForemanRhCloud.global_foreman_proxy ||
     ''
   end
 
+  def self.fix_port(uri_string)
+    return '' if uri_string.empty?
+
+    uri = URI(uri_string)
+    uri.send(:define_singleton_method, :default_port, -> { nil })
```
adamruzicka, line 55: This feels a bit brutal. Could we do this using refinements instead?

ShimShtein: Didn't know about this structure...
I don't think it will work in our case, since `default_port` is a class method: https://github.com/ruby/uri/blob/f3c4280bf63de60e482bbb87f576dd5ec94c5ea9/lib/uri/generic.rb#L32

ShimShtein: I am changing only a single instance of a class, so I think it's the least damage I can do

adamruzicka: Alright

adamruzicka (approved):

laviro (approved):

laviro:
Thanks @ShimShtein @adamruzicka !!

PR #540, opened 2021-05-02
Create a notification about hits on foreman host

ShimShtein:


db/seeds.d/179_ui_notifications.rb
```diff
@@ -0,0 +1,10 @@
+blueprints = [
+  {
+    group: N_('Red Hat Insights'),
+    name: 'insights_satellite_hits',
+    message: N_('Satellite server has %{subject} recommendations by Red Hat'),
+    level: 'warning',
+  },
+]
+
+blueprints.each { |blueprint| UINotifications::Seed.new(blueprint).configure }
```
laviro, line 10: nit: extra line here

lib/insights_cloud/async/insights_generate_notifications.rb
```diff
@@ -0,0 +1,58 @@
...
+
+      def blueprint
+        self.class.blueprint
+      end
+
+      def add_notification(hits_count)
+        Notification.create!(
+          initiator: User.anonymous_admin,
+          audience: ::Notification::AUDIENCE_ADMIN,
+          message: UINotifications::StringParser.new(blueprint.message, {subject: hits_count}).to_s,
```
laviro, line 44: should it be:
```suggestion
          message: UINotifications::StringParser.new(blueprint.message, {hits_count: hits_count}).to_s,
```

and then in the [message](https://github.com/theforeman/foreman_rh_cloud/pull/540/files#diff-cd6cf07e82a109107cad0037db33477d19d1e8ad65fb3973cea2ccef81f49274R5) it will be more readable

laviro (commented):
Nice addition @ShimShtein! Looks great, left some inline comments

db/seeds.d/179_ui_notifications.rb
ShimShtein: This is the line that actually creates the blueprint objects in the DB. The lines above are just definitions of the `blueprints` attributes array.

lib/insights_cloud/async/insights_generate_notifications.rb
ShimShtein: Done. This was kind of leftover from earlier version: in theory `message` can be generated automatically by the [framework](https://github.com/theforeman/foreman/blob/3347fa49d500964f0209122d8d36c920d1feafcc/app/models/notification.rb#L79), but the framework injects only the `subject` and `initiator`. Unfortunately `subject` has to be an ActiveRecord, so I had to parse the string manually.

ShimShtein:

db/seeds.d/179_ui_notifications.rb
adamruzicka: No newline at end of file

adamruzicka (commented):
Looks sensible

ShimShtein:
@adamruzicka, ready for another round

laviro (approved):
Thanks! @ShimShtein looks good to me :)

PR #543, opened 2021-05-18
Add scheduled inventory sync task

ShimShtein:
Including a recurring task that will schedule it each night.


ShimShtein:
@adamruzicka, does this sound reasonable?

lib/foreman_rh_cloud/engine.rb
```diff
@@ -128,6 +128,14 @@ class Engine < ::Rails::Engine
         description: N_('Run remediation playbook generated by Insights'),
         host_action_button: false
       )
+
+      unless ForemanTasks::RecurringLogic.joins(:tasks).merge(
+        ForemanTasks::Task.where(label: 'InventorySync::Async::InventoryScheduledSync')
+      ).exists?
+        recurring_logic = ForemanTasks::RecurringLogic.new_from_cronline("0 0 * * *")
+        recurring_logic.save!
+        recurring_logic.start(InventorySync::Async::InventoryScheduledSync)
```
adamruzicka, line 137: `#start` calls `#start_after`, which calls `#save!` under the hood so you should be able to reduce this to
```suggestion
        ForemanTasks::RecurringLogic.new_from_cronline("0 0 * * *")
                                    .start(InventorySync::Async::InventoryScheduledSync)
```

lib/inventory_sync/async/inventory_scheduled_sync.rb
```diff
@@ -0,0 +1,19 @@
+module InventorySync
+  module Async
+    class InventoryScheduledSync < ::Actions::EntryAction
+      include ::Actions::RecurringAction
+
+      def plan
+        concurrence do
```
adamruzicka, line 7: concurrence is the default flow so you can drop this block and reduce the nesting by one level

lib/inventory_sync/async/inventory_scheduled_sync.rb
```diff
@@ -0,0 +1,19 @@
...
+          Organization.unscoped.each do |org|
+            plan_org_sync(org)
```
adamruzicka, line 9: Is having a method for it really necessary?

lib/foreman_rh_cloud/engine.rb
```diff
@@ -128,6 +128,14 @@ class Engine < ::Rails::Engine
         description: N_('Run remediation playbook generated by Insights'),
         host_action_button: false
       )
+
+      unless ForemanTasks::RecurringLogic.joins(:tasks).merge(
+        ForemanTasks::Task.where(label: 'InventorySync::Async::InventoryScheduledSync')
+      ).exists?
+        recurring_logic = ForemanTasks::RecurringLogic.new_from_cronline("0 0 * * *")
+        recurring_logic.save!
```
adamruzicka, line 136: This isn't really thread safe. Is that an issue here?

adamruzicka (commented):
A couple of nitpicks, but otherwise looks good

lib/foreman_rh_cloud/engine.rb
ShimShtein: For some reason it didn't save the logic unless I saved it manually. Should I debug it further, or leaving it as it is right now is also OK?

lib/inventory_sync/async/inventory_scheduled_sync.rb
ShimShtein: I am using those methods for testing. It's much easier to mock a method on this class than the actual call to foreman tasks.

lib/foreman_rh_cloud/engine.rb
ShimShtein: I am not sure, but I suppose there shouldn't be many intializers running in parallel. If we see a problem with multiple schedules, we will need to solve it.

adamruzicka:
Looks good, kicking-off the tests again

lib/foreman_rh_cloud/engine.rb
```diff
@@ -122,12 +122,24 @@ class Engine < ::Rails::Engine
...
+
+        unless ForemanTasks::RecurringLogic.joins(:tasks).merge(
+          ForemanTasks::Task.where(label: 'InventorySync::Async::InventoryScheduledSync')
+        ).exists?
+          recurring_logic = ForemanTasks::RecurringLogic.new_from_cronline("0 0 * * *")
+          recurring_logic.save!
+          recurring_logic.start(InventorySync::Async::InventoryScheduledSync)
+        end
+      end
+    rescue ActiveRecord::NoDatabaseError
```
adamruzicka, line 142: Do we need the rescue if we're checking if the table exists or does the check actually raise this?

ShimShtein: Yes, we need. I have tried it. The rescue is helping to run `rake db:create` and the table existence test helps to run `rake db:migrate`

adamruzicka (approved):

ShimShtein:
@laviro, ready for your review

laviro (commented):
Thanks @ShimShtein, I see it was added as scheduled in tasks,
but I don't see it being added to recurring logic on my end if it's needed...

ShimShtein:
@laviro found the bug - during initialization there is no valid user for the changes and permissions scope prevented the `RecurringLogic` object from being created.
Now wrapped it in forced user scope, that does not exist in test DB, hence the check for existence.

laviro (approved):
Thanks @ShimShtein, works now !

PR #551, opened 2021-06-02
Omit malformed bios_uuid values

ShimShtein:


lib/foreman_inventory_upload/generators/slice.rb
```diff
@@ -47,7 +47,7 @@ def report_host(host)
           @stream.simple_field('account', account_id(host.organization).to_s)
           @stream.simple_field('subscription_manager_id', host.subscription_facet&.uuid)
           @stream.simple_field('satellite_id', host.subscription_facet&.uuid)
-          @stream.simple_field('bios_uuid', fact_value(host, 'dmi::system::uuid'))
+          @stream.simple_field('bios_uuid', bios_uuid(host))
```
laviro, line 50: can we do the same regex also for:
- subscription_manager_id
- satellite_id
- vm_uuid
- report_slice_id

ShimShtein:
@laviro ready for another round!

laviro (approved):
Thanks! Looks good to me :)

PR #553, opened 2021-06-03
Convert all reported tag values to string

ShimShtein:


adamruzicka:
Isn't this a bit non-standard json dumping behavior?

ShimShtein:
Are you referring to the `@stream` pattern? Yes, it's not standard, the problem is there is no standard json streaming library, and I didn't want to keep all the report object in memory, it's just too big.

adamruzicka:
I commented before knowing what's going on, I wasn't aware that the thing processing the uploaded data requires all values to be strings. Feel free to ignore the previous comment

laviro (approved):
Thanks @ShimShtein !

PR #561, opened 2021-06-08
Refresh hits after sync

ShimShtein:


laviro:
@MariaAga, @amirfefer could you review it please? I am on child sick leave and this is kinda a blocker for the release

webpack/common/ForemanTasks/ForemanTasksActions.js
```diff
@@ -0,0 +1,64 @@
...
+        if (data.result === 'success') {
+          stopTaskInterval();
+          dispatch(onTaskSuccess(data));
+        }
+        if (data.result === 'error') {
+          stopTaskInterval();
+          if (taskErrorMessage === undefined) {
+            taskErrorMessage = errorData =>
+              `${__('The task failed with the following error:')} ${
+                errorData.humanized.errors[0]
```
MariaAga, line 31: Is `.humanized.errors[0]` always available? might be good to be on the safe side with `errorData?.humanized?.errors[0]`

MariaAga:
@ShimShtein, is it ok that on error the `Recommendation sync has failed` is shown twice?

ShimShtein:
> @ShimShtein, is it ok that on error the Recommendation sync has failed is shown twice?

It shouldn't. I wonder why is it happening.

MariaAga:
looks like the interval was too fast. when I added a longer delay it only showed the error once

webpack/InsightsCloudSync/InsightsCloudSyncActions.js
```diff
@@ -1,25 +1,42 @@
...
+
+const syncInsightsStartedToast = taskId =>
+  taskRelatedToast(taskId, 'info', __('Recommendation sync has started: '));
+
+const setupInsightsTaskPolling = (taskId, fetchInsights, query, dispatch) =>
+  setupTaskPolling({
+    taskId,
+    key: INSIGHTS_CLOUD_SYNC_TASK,
+    onTaskSuccess: () => {
+      fetchInsights({ query, page: 1 });
```
laviro, line 38: looks like I get the success callback, but for some reason I see a toast with the error: 
`Could not get task details: TypeError: Cannot read property 'type' of undefined`

laviro (commented):
Thanks @ShimShtein, added a few inline comments

webpack/InsightsCloudSync/InsightsCloudSyncActions.js
MariaAga: I needed to change in `add_satellite_notifications` the line:
`blueprint&.notifications&.destroy_all`
 Because I got an error that notifications is undefined and then I started getting this error as well

webpack/common/ForemanTasks/ForemanTasksActions.js
```diff
@@ -0,0 +1,64 @@
...
+    })
+  );
+
+export const taskRelatedToast = (taskID, type, message) =>
+  addToast({
+    sticky: false,
+    type,
+    message: (
+      <span>
+        {message}
```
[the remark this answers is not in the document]
MariaAga: I wanted to suggested maybe a `<br>` instead?

webpack/InsightsCloudSync/InsightsCloudSyncActions.js
laviro: seem that because of the `dispatch` in https://github.com/theforeman/foreman_rh_cloud/pull/561/files#diff-062e73f73ae1ac5114f5c53151d1230ee336d3c844290d4d966c8f1236b8a26cR24
and because this `onTaskSuccess` returns nothing, eventually what's happening is `dispatch(undefined)`
action should be in a type of object with `{type: 'SOME-TYPE', payload: {}}`
I would also take a closer look on https://github.com/theforeman/foreman_rh_cloud/pull/561/files#diff-5e22a22b04f74adeb26948f4c823ba96e6fa74fffe6823468cb0d65df73c2f43R40

laviro (approved):
Overall looks great! that issue is gone now
although we need to investigate why the double API call in insights

webpack/InsightsCloudSync/InsightsCloudSyncActions.js
```diff
@@ -1,25 +1,49 @@
...
+
+const syncInsightsStartedToast = taskId =>
+  taskRelatedToast(taskId, 'info', __('Recommendation sync has started: '));
+
+const setupInsightsTaskPolling = (taskId, fetchInsights, query, dispatch) =>
+  setupTaskPolling({
+    taskId,
+    key: INSIGHTS_CLOUD_SYNC_TASK,
+    onTaskSuccess: () => {
+      dispatch(fetchInsights({ query, page: 1 }));
```
laviro, line 38: to fix the double API call, I removed the `dispatch` here:

```suggestion
      fetchInsights({ query, page: 1 });
```

and also reverted the change of return in https://github.com/theforeman/foreman_rh_cloud/pull/561/files#diff-7dfe910e17369a7213356b886dabc440bb547dbec9b42ad7bda757f9b368f3b5R43

but I am not entirely sure why it works to be honest... 
maybe it gets the `dispatch` from the scope above? 🤔

ShimShtein: I have left the internal return, but removed the dispatch. It was enough to stop the double call.

From looking at the call stacks attached to redux, it looks like first the `fetchInsights` is executed and then the result is passed to `dispatch` which executes it again for some reason.

Inner call:

Screenshot from 2021-06-13 21-56-55

Outer call:

Screenshot from 2021-06-13 21-55-22

laviro: awesome! let's do that then :)

ShimShtein:
@laviro, done!

laviro (commented):
That fixed the earlier issue, thanks @ShimShtein !

one last thing that I found, when an error occurs in the inventory page it shows:
Screenshot - 2021-06-15T112123 701
 
and before there was this message that is visible in the network:
```json
{"message":"Nothing to sync, there are no hosts with subscription for this organization."}
```

laviro (approved):
great work @ShimShtein ! thanks @MariaAga for reviewing!
rails tests have already passed, only JS code changes, merging

PR #571, opened 2021-06-20
Add UUID to insights facets that are missing it

ShimShtein:


lib/inventory_sync/async/inventory_hosts_sync.rb
```diff
@@ -20,14 +20,19 @@ def create_facets
...
           {
             host_id: host_id,
             uuid: uuid,
           }
         end
         InsightsFacet.create(missing_facets)
+        missing_uuids = existing_facets.select { |host_id, uuid| uuid.empty? }.map(&:first)
+
+        missing_uuids.each do |host_id|
+          InsightsFacet.where(host_id: host_id).update_all(uuid: uuids_hash[host_id])
```
adamruzicka, line 34: How about something like this this?

```suggestion
        missing_uuids = existing_facets.select { |host_id, uuid| uuid.empty? }.each do |host_id, uuid|
          InsightsFacet.where(host_id: host_id).update_all(uuid: uuid)
```

adamruzicka (commented):
One suggestion, but it might be down to personal taste. Otherwise looks good

ShimShtein:
@adamruzicka Done!

adamruzicka (approved):

laviro (approved):
Thanks @ShimShtein !

PR #573, opened 2021-06-21
Fix no user and no org warning in generate task

ShimShtein:


lib/tasks/rh_cloud_inventory.rake
```diff
@@ -29,8 +29,9 @@ namespace :rh_cloud_inventory do
         puts "Using #{base_folder} for the output"
       end
 
-      unless portal_user || organizations.empty?
+      if portal_user.empty? && organizations.empty?
         puts "Must specify either portal_user or organization_id"
+        return
```
adamruzicka, line 34: If I'm reading it right, if a user provides both `portal_user` and `organization_id` then `portal_user` takes precedence, correct? Even though in theory a provided `organization_id` might be outside of the organizations associated with `portal_user`?

ShimShtein: Yes, portal user will take precedence, and the organizations list would be rewritten with a notification: https://github.com/theforeman/foreman_rh_cloud/blob/7f0f967d4870f0fce14d4fc948a5189f7783f54f/lib/tasks/rh_cloud_inventory.rake#L41

adamruzicka (approved):

PR #575, opened 2021-06-22
Add remediations button to host details

ShimShtein:


app/helpers/foreman_insights_host_helper.rb
```diff
@@ -0,0 +1,19 @@
+module ForemanInsightsHostHelper
+  def insights_host_overview_buttons(host)
+    search_condition = "hostname=#{host.fqdn}"
+    [
+      {
+        button: link_to_if_authorized(
+          _("Recommendations"),
```
laviro, line 7: how about renaming it to `Insights`?

laviro (approved):

app/helpers/foreman_insights_host_helper.rb
ShimShtein: "Insights" has a bit too broad meaning. It's the whole concept of Red Hat looking on your infrastructure. Here we want to be more specific and tell the user that he will be redirected to recommendations about this host.

PR #577, opened 2021-06-27
Add text about cloud connector

ShimShtein:
Added another line to the header text according to @MariSvirik's design:

image


ShimShtein:
@laviro, I see the tests are failing without me changing anything. Should we pin Foreman version?

laviro:
I would suggest adding the package-lock file into our CI cache and then our test should pass,
other than that the error isn't related so ACK from me,
if we have time to do that, let's rebase right after this PR and merge,
if we see that we don't have time before the release, it would be OK to merge and fix later IMO.

laviro (approved):

laviro:
Thanks @ShimShtein for the fast fix!!
that works great, we might want to try caching on a follow-up?

PR #578, opened 2021-06-28
Fix hits controller to always use authorized resource_base

ShimShtein:


ShimShtein:
@adamruzicka , here's another one waiting for review

app/controllers/insights_cloud/hits_controller.rb
```diff
@@ -23,9 +23,9 @@ def show
 
     def resolutions
       if remediation_all_selected_param
-        hits = InsightsHit.with_playbook.search_for(params[:query])
+        hits = resource_base.with_playbook.search_for(params[:query])
```
adamruzicka, line 26: How about using `resource_base_with_search`?

app/controllers/insights_cloud/hits_controller.rb
```diff
@@ -23,9 +23,9 @@ def show
...
       else
-        hits = resource_base_search_and_page.where(id: remediation_ids_param)
+        hits = resource_base_search_and_page.with_playbook.where(id: remediation_ids_param)
```
adamruzicka, line 28: How does this work? `resource_base_search_and_page` expects the search query in `params[:search]`, yet the truthy branch uses `params[:query]`. Which is right?

app/controllers/insights_cloud/hits_controller.rb
ShimShtein: `_with_search` hard codes `:search` parameter instead of `:query`.

app/controllers/insights_cloud/hits_controller.rb
ShimShtein: It depends on a switch on user side: when the user switches the "select all from all pages", we pass the `:query` param. Otherwise we use the regular paginated table parameters (`:search`, `:page` e.t.c.)

adamruzicka: A bit unexpected but fine

adamruzicka (approved):

laviro (approved):

PR #592, opened 2021-07-06
Inject deprecation warning to old plugin

ShimShtein:
Now on all pages you should see the same deprecation:

image

@MariSvirik Does it makes sense to you?

### DIFF
--- a/app/helpers/foreman_insights_deprecations_helper.rb
+++ b/app/helpers/foreman_insights_deprecations_helper.rb
@@ -1,0 +1,9 @@
+module ForemanInsightsDeprecationsHelper
+  def old_plugin_deprecation_warning
+    alert(
+      class: 'alert-warning',
+      close: false,
+      text: _('redhat_access plugin is deprecated and will be removed in Satellite 6.10. You can find recommendations for your hosts on hosts index and details pages.')
+    )
+  end
+end
--- a/app/overrides/old_plugin_deprecation.rb
+++ b/app/overrides/old_plugin_deprecation.rb
@@ -1,0 +1,20 @@
+Deface::Override.new(
+  virtual_path: 'redhat_access/analytics_dashboard/welcome',
+  name: 'deprecation_warning',
+  insert_before: 'div#welcome',
+  text: '<%= old_plugin_deprecation_warning %>'
+)
+
+Deface::Override.new(
+  virtual_path: 'redhat_access/analytics_dashboard/index',
+  name: 'deprecation_warning_index_not_met',
+  insert_before: 'article#content',
+  text: '<%= old_plugin_deprecation_warning %>'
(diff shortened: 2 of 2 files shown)


MariSvirik:
@ShimShtein
Is the deprecation warning always on the page without content such as this one? (if yes, we can use an empty state, if not we should use an inline notification)
Is this warning connected to the fact that Insights is not enabled? If not I don't see a reason to show it there.
What happens if the users use recommendations counts? 
How does it help them to solve the situation? What is recommendation counts? 
Can we be more specific and link them to that page?

Probably out of this PR but:
PAGE CONTENT:  I would use an empty state design with a warning icon and primary button to enable whatever we want them to enable. Do we want the users to enable Insights, right? 

Btw. thanks for tagging me

ShimShtein:
@MariSvirik Let me share some context: foreman_rh_cloud plugin should replace the older redhat_access plugin.
All pages under the `Insights` top menu would be gone.

What I want to do in this PR is to add a deprecation warning on top of all those pages, to warn the user that these pages will be gone soon.
What you see in the screenshot is their "welcome" screen. This screen is shown when insights plugin is not configured. 
I have implemented the same warning also for the internal screens, it's a warning you see above the content of those pages. I don't have the screenshots handy, sorry, but I will generate them later I hope.

About the recommendations counts: I wanted to point the user where the same information is displayed by the new plugin. The idea is that the user now has the recommendations count column on hosts index page and recommendations tab on host details page. Does `You can find recommendations for your hosts on hosts index and details pages` sound better?

ShimShtein:
Latest version:

image

laviro (commented):
Overall looks good to me, waiting for more responses

laviro (approved):

MariSvirik:
Ok, now I get it. If you can't link users anywhere, just add a period at the end of the sentence and leave it as it is.

app/helpers/foreman_insights_deprecations_helper.rb
```diff
@@ -0,0 +1,9 @@
+module ForemanInsightsDeprecationsHelper
+  def old_plugin_deprecation_warning
+    alert(
+      class: 'alert-warning',
+      close: false,
+      text: _('redhat_access plugin is deprecated and will be removed in Satellite 6.10. You can find recommendations for your hosts on hosts index and details pages')
```
laviro, line 6:
```suggestion
      text: _('redhat_access plugin is deprecated and will be removed in Satellite 6.10. You can find recommendations for your hosts on hosts index and details pages.')
```

ShimShtein:
@laviro done!

laviro (approved):

laviro:
Thanks @ShimShtein & @MariSvirik !

PR #593, opened 2021-07-06
Add self host sync task

ShimShtein:
Since foreman host will not be registered to itself, it will not be synced using the regular hosts sync task. This is because hosts sync task assumes all hosts have subscription facet and on the cloud side they will be marked as managed by Satellite.

### DIFF
--- a/lib/foreman_inventory_upload.rb
+++ b/lib/foreman_inventory_upload.rb
@@ -70,8 +70,16 @@ module ForemanInventoryUpload
     folder
   end
 
+  def self.inventory_base_url
+    ForemanRhCloud.base_url + "/api/inventory/v1/hosts"
+  end
+
   def self.inventory_export_url
     tags = URI.encode("satellite/satellite_instance_id=#{Foreman.instance_id}")
-    ForemanRhCloud.base_url + "/api/inventory/v1/hosts?tags=#{tags}"
+    inventory_base_url + "?tags=#{tags}"
+  end
+
+  def self.inventory_self_url
+    inventory_base_url + "?hostname_or_id=#{ForemanRhCloud.foreman_host.fqdn}"
   end
 end
--- a/lib/inventory_sync/async/inventory_hosts_sync.rb
+++ b/lib/inventory_sync/async/inventory_hosts_sync.rb
@@ -4,6 +4,12 @@ module InventorySync
       set_callback :iteration, :around, :setup_facet_transaction
       set_callback :step, :around, :create_facets
 
+      def plan
+        # by default the tasks will be executed concurrently
+        plan_self
+        plan_self_host_sync
+      end
+
       def setup_facet_transaction
         InsightsFacet.transaction do
           yield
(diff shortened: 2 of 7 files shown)


lib/inventory_sync/async/inventory_self_host_sync.rb
```diff
@@ -0,0 +1,30 @@
...
+      def add_missing_insights_facet(uuids_hash)
+        facet = InsightsFacet.find_or_create_by(host_id: ForemanRhCloud.foreman_host.id) do |facet|
+          facet.uuid = uuids_hash.values.first
+        end
+
+        # fix empty uuid in case the facet already exists
+        facet.update(uuid: uuids_hash.values.first) unless facet.uuid
+      end
+
+      def inventory_url
```
laviro, line 25: should it be `request_url`?
iiuc it should override the method in `query_inventory_job.rb` ?

laviro (commented):
> Since foreman host will not be registered to itself

is there a small chance that someone did register the Satellite host?
do we want to verify that before doing another call?

laviro:
@adamruzicka can you also have a look?

lib/foreman_inventory_upload.rb
```diff
@@ -70,8 +70,16 @@ def self.ensure_folder(folder)
     folder
   end
 
+  def self.inventory_base_url
+    ForemanRhCloud.base_url + "/api/inventory/v1/hosts"
```
adamruzicka, line 74: Is there an url helper for this?

lib/inventory_sync/async/inventory_hosts_sync.rb
```diff
@@ -4,6 +4,13 @@ class InventoryHostsSync < QueryInventoryJob
       set_callback :iteration, :around, :setup_facet_transaction
       set_callback :step, :around, :create_facets
 
+      def plan
+        concurrence do
```
adamruzicka, line 8: Concurrence is the default, there's no need to be explicit about it

lib/foreman_inventory_upload.rb
ShimShtein: This is the URL helper :) It's cloud address, the full address would be similar to `https://cloud.redhat.com/api/inventory/v1/hosts`. It's not a local Foreman URL.

ShimShtein:
Ready for another round. Once we will move to cert auth instead of API token we use currently, we will need to change the implementation a bit.

lib/inventory_sync/async/inventory_self_host_sync.rb
ShimShtein: Nice catch!

lib/inventory_sync/async/inventory_hosts_sync.rb
ShimShtein: Removed redundant call

laviro (approved):

adamruzicka (approved):
Code-wise looks good

laviro:
Thanks @ShimShtein @adamruzicka !!

PR #607, opened 2021-07-25
BZ 1967138 - remediation modal vertical nav overlap

laviro:
the remediation modal is not visible properly in reduced browser window.


ShimShtein:
Merged, thanks @laviro!

PR #608, opened 2021-07-25
Skip scheduled actions if token is absent

ShimShtein:


ShimShtein:
@adamruzicka, your opinion would be much appreciated. The request is quite simple: I want to skip planning an action, if the token setting is empty.

app/services/foreman_rh_cloud/remediations_retriever.rb
```diff
@@ -12,6 +12,11 @@ def initialize(hit_remediation_pairs, logger: Logger.new(IO::NULL))
     end
 
     def create_playbook
+      unless cloud_auth_available?
+        logger.debug('Cloud authentication is not available, cannot continue')
+        return nil
```
adamruzicka, line 17: The `nil` is not really needed

adamruzicka (approved):
This is the approach I'd take if I were to do this myself, lgtm

ShimShtein:
Removed the `nil`

laviro (approved):
Thanks @ShimShtein ! LGTM

PR #611, opened 2021-07-28
Skip empty parameter tags

ShimShtein:


lib/foreman_inventory_upload/generators/tags.rb
```diff
@@ -19,7 +19,9 @@ def generate
       def generate_parameters
         return [] unless Setting[:include_parameter_tags]
 
-        (@host.host_inherited_params_objects || []).map { |item| [item.name, item.value] }
+        (@host.host_inherited_params_objects || [])
+          .map { |item| [item.name, item.value] }
+          .select { |_name, value| value.present? || value.is_a?(FalseClass) }
```
laviro, line 24: is it possible to use [filter_map](https://blog.saeloun.com/2019/05/25/ruby-2-7-enumerable-filter-map.html) here?

ShimShtein: This method is introduced in ruby 2.7, but we ship foreman 2.3 with ruby 2.5.

laviro (approved):

laviro:
Thanks @ShimShtein !

PR #612, opened 2021-07-28
Remove remediations from experimental features

ShimShtein:


webpack/InsightsCloudSync/Components/RemediationModal/RemediateButton.js
```diff
@@ -1,59 +1,28 @@
...
-        }
-      >
-        {button}
-      </Popover>
-    );
-  }
-  return button;
-};
+const RemediateButton = ({ selectedIds, toggleModal }) => (
+  <Button
```
laviro, line 8: Yay! this is much simpler now,
I would even suggest moving it to `RemediationModal.js` for simplicity and to avoid the extra file for a simple button

laviro (commented):
Work great, thanks @ShimShtein !

ShimShtein:

laviro (approved):
Thanks @ShimShtein, LGTM

PR #617, opened 2021-07-29
Skip inventory full sync without token

ShimShtein:
This is a leftover after https://github.com/theforeman/foreman_rh_cloud/pull/608

### DIFF
--- a/lib/inventory_sync/async/inventory_full_sync.rb
+++ b/lib/inventory_sync/async/inventory_full_sync.rb
@@ -5,6 +5,11 @@ module InventorySync
       set_callback :step, :around, :update_statuses_batch
 
       def plan(organization)
+        unless cloud_auth_available?
+          logger.debug('Cloud authentication is not available, skipping inventory hosts sync')
+          return
+        end
+
         plan_self(organization_id: organization.id)
       end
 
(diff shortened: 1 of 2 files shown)


ShimShtein:
@adamruzicka this one is the last task I have missed from the previous PR.
Can you take a quick look?

lib/inventory_sync/async/inventory_full_sync.rb
```diff
@@ -5,6 +5,11 @@ class InventoryFullSync < InventoryHostsSync
       set_callback :step, :around, :update_statuses_batch
 
       def plan(organization)
+        unless cloud_auth_available?
+          logger.debug('Cloud authentication is not available, skipping inventory hosts sync')
```
laviro, line 9: From the UI we also disable to trigger this task if no token was found: https://github.com/theforeman/foreman_rh_cloud/blob/b7c22c6b2428f465cc5f06ac5ae5e7efa4aac577/webpack/ForemanInventoryUpload/Components/PageHeader/components/SyncButton/components/Modal.js#L24

though for rake tasks or some other ways to trigger the sync I guess it's better to return early, LGTM

laviro (approved):

laviro:
Thanks @ShimShtein !

PR #621, opened 2021-08-03
Remove foreman access

ShimShtein:


foreman_rh_cloud.gemspec
```diff
@@ -19,7 +19,6 @@ Gem::Specification.new do |s|
   s.test_files -= Dir['test/**/*.orig']
 
   s.add_dependency 'katello'
-  s.add_dependency 'redhat_access'
```
laviro, line 22: Do we also need to update https://github.com/theforeman/foreman_rh_cloud/blob/master/config/Gemfile.lock.gh_test ?

laviro (commented):
Overall looks good to me,
@adamruzicka could you also take a look? thanks!

ShimShtein:
@laviro, fixed gemfile.lock too

laviro (approved):

PR #624, opened 2021-08-12
Finally fix --test-connection request

ShimShtein:


jameerpathan111:
@ShimShtein I have tried this PR and it seems to fix the issue.

ShimShtein:
@lphiri, can you please review this one?

app/services/foreman_rh_cloud/cloud_request_forwarder.rb
```diff
@@ -76,6 +82,10 @@ def platform_request?
       ->(request_path) { request_path.include? '/platform' }
     end
 
+    def connection_test_request?
+      ->(request_path) { request_path =~ /redhat_access\/r\/insights\/?$/ }
```
lphiri, line 86: If the insights-client  is configured to run with "legacy_upload=False", test connections will come
in via the "/platform" path, so this regex will not catch it.
But I think that is fine because the client calls  "/r/insights/platform/apicast-tests/ping" itself.

lphiri (approved):
@ShimShtein  - this looks good to me - minor comment about behavior with legacy_upload=False

app/services/foreman_rh_cloud/cloud_request_forwarder.rb
ShimShtein: Double checked, it works with `legacy_upload=False` too

laviro (approved):

laviro:
Thanks @ShimShtein ! tested and it fixes the issue!

PR #629, opened 2021-08-19
Add failure strategy for all tasks

ShimShtein:


lib/inventory_sync/async/inventory_self_host_sync.rb
```diff
@@ -20,6 +20,10 @@ def create_facets
         results
       end
 
+      def rescue_strategy_for_self
+        Dynflow::Action::Rescue::Skip
+      end
```
adamruzicka, line 25: `InventoryHostsSync` plans `InventorySelfHostSync`, but they both run in concurrence. If both fail, the whole execution plan will pause itself, since one of the actions suggests to `Skip` and other to `Fail` and dynflow can't decide which is right.

We should either make them match or make `InventoryHostsSync` override the behavior by defining the strategy as `rescue_strategy` instead of `rescue_strategy_for_self`

lib/insights_cloud/async/insights_generate_notifications.rb
```diff
@@ -53,6 +53,10 @@ def add_notification(hits_count)
           }
         )
       end
+
+      def rescue_strategy_for_self
+        Dynflow::Action::Rescue::Skip
+      end
```
adamruzicka, line 59: Similar situation between `InsightsGenerateNotifications` and `InsightsRulesSync` as between `InventoryHostsSync`  and `InventorySelfHostSync`

ShimShtein:
@adamruzicka for now marked all task as `Fail`. Will think of a more relaxed strategy if it will become an issue.

adamruzicka (approved):
Looks good

laviro:
Thanks @adamruzicka @ShimShtein !!

PR #632, opened 2021-08-23
Fix inventory page header

laviro:


webpack/ForemanInventoryUpload/Components/PageHeader/PageTitle.js
```diff
@@ -5,6 +5,7 @@ import {
   KebabToggle,
   DropdownPosition,
 } from '@patternfly/react-core';
+import Head from 'foremanReact/components/Head';
```
laviro, line 8: exist since Foreman 2.5: https://github.com/theforeman/foreman/blob/2.5-stable/webpack/assets/javascripts/react_app/components/Head/index.js

laviro:
@ShimShtein looks like our ci is using an older version of foreman,
that's why the assets:precompile fails, opened https://github.com/theforeman/foreman_rh_cloud/pull/633 to update it

laviro:
@ShimShtein can you take a look?

laviro:
Rebased,

here is the screenshot of how it used to be:
Screenshot-20210929112024-1845x566

and with this PR:
Screenshot-20210929111639-2098x481

and also fixed the browser tab title

laviro:
oh, ruby test assets:compile fails because it is still against foreman v2.3
the Head component I used is there since foreman v2.5: https://github.com/theforeman/foreman/blob/2.5-stable/webpack/assets/javascripts/react_app/components/Head/index.js

ShimShtein:
LGTM, merging. Thanks @laviro!

PR #643, opened 2021-09-09
Use a prebuilt container for tests

ShimShtein:


.github/images/Dockerfile
```diff
@@ -0,0 +1,59 @@
...
+
+RUN \
+  yum install -y centos-release-scl; \
+  yum install -y systemd-devel make tar libvirt-devel rh-ruby25-ruby-devel zlib-devel libxml2-devel openssl-libs libvirt-devel nodejs rh-ruby25-rubygems automake gcc gcc-c++ kernel-devel libcurl-devel sqlite-devel rh-nodejs12-npm sclo-git25 rh-postgresql12-postgresql-devel rh-postgresql12-postgresql; \
+  source scl_source enable $SCLS; \
+  git config --global user.name $GIT_COMMITTER_NAME; \
+  git config --global user.email $GIT_COMMITTER_EMAIL; \
+  mkdir /projects; \
+  cd /projects; \
+  git clone --depth 1 --branch 2.3-stable https://github.com/theforeman/foreman.git; \
```
laviro, line 22: for master this should be `develop` I guess and then we should change it to `3.1-stable` once it will be branched.

.github/images/Dockerfile
```diff
@@ -0,0 +1,59 @@
...
+  git clone --depth 1 --branch KATELLO-3.18 https://github.com/Katello/katello.git; \
+  git clone --depth 1 --branch v3.0.5 https://github.com/theforeman/foreman-tasks.git; \
+  git clone --depth 1 --branch foreman_2_3 https://github.com/theforeman/foreman_rh_cloud.git; \
```
laviro, line 25: can you update those to the latest branches?

PR #646, opened 2021-09-19
Switch to host_params method

ShimShtein:


lib/foreman_inventory_upload/generators/tags.rb
```diff
@@ -19,8 +19,7 @@ def generate
       def generate_parameters
         return [] unless Setting[:include_parameter_tags]
 
-        (@host.host_inherited_params_objects || [])
-          .map { |item| [item.name, item.value] }
+        (@host.host_params || {})
```
laviro, line 22: Looks better 👍🏼 
btw, on my end, I see that with `host_params` there are additional new fields that didn't exist before,
though it includes all of the previous params pairs so it seems ok

laviro (approved):

ShimShtein:
Thanks for the ACK, merging.

PR #649, opened 2021-09-19
Move all jobs to foreman tasks

ShimShtein:


lib/foreman_inventory_upload/async/generate_all_reports_job.rb
```diff
@@ -16,17 +18,19 @@ def perform
...
       end
 
-      def self.singleton_job_name
-        name
+      def rescue_strategy_for_self
+        Dynflow::Action::Rescue::Fail
+      end
+
+      def plan_generate_report(folder, organization)
+        ForemanTasks.async_task(ForemanInventoryUpload::Async::GenerateReportJob, folder, organization.id)
```
adamruzicka, line 33: Is there a reason for going with a separate execution plan for each organization instead of having this just as actions with different inputs within the same execution plan?

laviro (approved):
Overall looks good, thanks @ShimShtein 👍🏼 
any other thoughts @adamruzicka ?

adamruzicka (approved):
Nope, LGTM

laviro:
Thanks @ShimShtein @adamruzicka !

PR #655, opened 2021-10-19
update settings code to match latest foreman

ShimShtein:


ShimShtein:
It didn't fail till now, so we can assume 2.5 is quite safe from this bug

laviro (approved):
Thanks @ezr-ondrej @ShimShtein !

PR #661, opened 2021-11-16
Add total risk chart

Ron-Lavi:
https://user-images.githubusercontent.com/26363699/143744638-f48cdffb-6925-489c-b24b-08e0535d51a1.mp4

### DIFF
--- a/webpack/InsightsHostDetailsTab/InsightsTab.scss
+++ b/webpack/InsightsHostDetailsTab/InsightsTab.scss
@@ -88,3 +88,29 @@
 #new_host_details_insights_tab {
   padding: 24px;
 }
+
+#rh-cloud-total-risk-card {
+  svg {
+    tspan:nth-child(2) {
+      fill: black !important;
+    }
+
+    g:nth-child(2) {
+      path {
+        cursor: pointer;
+      }
+    }
+
+    g:nth-child(3) {
+      a {
+        text {
+          tspan {
+            fill: var(--pf-global--link--Color) !important;
+            text-decoration: underline;
+            text-decoration-color: var(--pf-global--link--Color) !important;
+          }
+        }
+      }
+    }
+  }
+}
--- a/webpack/InsightsHostDetailsTab/InsightsTab.scss
+++ b/webpack/InsightsHostDetailsTab/InsightsTab.scss
@@ -88,29 +88,3 @@
 #new_host_details_insights_tab {
   padding: 24px;
 }
-
-#rh-cloud-total-risk-card {
-  svg {
-    tspan:nth-child(2) {
-      fill: black !important;
-    }
-
-    g:nth-child(2) {
-      path {
-        cursor: pointer;
-      }
-    }
-
-    g:nth-child(3) {
-      a {
-        text {
-          tspan {
-            fill: var(--pf-global--link--Color) !important;
-            text-decoration: underline;
-            text-decoration-color: var(--pf-global--link--Color) !important;
-          }
-        }
-      }
-    }
-  }
-}
--- a/webpack/InsightsHostDetailsTab/InsightsTotalRiskChart.js
+++ b/webpack/InsightsHostDetailsTab/InsightsTotalRiskChart.js
@@ -1,0 +1,156 @@
+import React, { useEffect, useState, useMemo } from 'react';
+import PropTypes from 'prop-types';
+import { useDispatch } from 'react-redux';
+import { push } from 'connected-react-router';
+import { useHistory } from 'react-router-dom';
+import { DropdownItem, Bullseye, Title } from '@patternfly/react-core';
+import { ChartDonut, ChartLegend, ChartLabel } from '@patternfly/react-charts';
+import { STATUS } from 'foremanReact/constants';
+import { useAPI } from 'foremanReact/common/hooks/API/APIHooks';
+import CardTemplate from 'foremanReact/components/HostDetails/Templates/CardItem/CardTemplate';
+import { translate as __ } from 'foremanReact/common/I18n';
+import SkeletonLoader from 'foremanReact/components/common/SkeletonLoader';
+import { insightsCloudUrl } from '../InsightsCloudSync/InsightsCloudSyncHelpers';
+import { getInitialRisks } from './InsightsTabConstants';
+
+const InsightsTotalRiskCard = ({ hostDetails: { id } }) => {
+  const [totalRisks, setTotalRisks] = useState(getInitialRisks());
+  const hashHistory = useHistory();
+  const dispatch = useDispatch();
+  const API_KEY = `HOST_${id}_RECOMMENDATIONS`;
+  const API_OPTIONS = useMemo(() => ({ key: API_KEY }), [API_KEY]);
+  const {
+    status = STATUS.PENDING,
+    response: { hits = [] },
+  } = useAPI('get', insightsCloudUrl(`hits/${id}`), API_OPTIONS);
+
+  useEffect(() => {
+    if (status !== STATUS.PENDING) {
+      const risks = getInitialRisks();
+      hits.forEach(({ total_risk: risk }) => {
+        risks[risk].value += 1;
+      });
+      risks.total = hits.length;
+      setTotalRisks(risks);
+    }
+  }, [hits, status]);
+
+  const onChartClick = (evt, { index }) => {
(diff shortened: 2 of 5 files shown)


Ron-Lavi:
@ShimShtein can you take a look?

Ron-Lavi:
@ShimShtein I updated the chart's legend to links that also redirect to the insights tab
gnome-shell-screenshot-3P7ZC1
:

webpack/InsightsHostDetailsTab/InsightsTotalRiskChart.js
```diff
@@ -0,0 +1,155 @@
...
+import ErrorState from '../ForemanInventoryUpload/Components/AccountList/Components/ErrorState/ErrorState';
+
+const InsightsTotalRiskCard = ({ hostDetails: { id } }) => {
+  const initialRisks = {
+    1: { value: 0, title: __('Low') },
+    2: { value: 0, title: __('Moderate') },
+    3: { value: 0, title: __('Important') },
+    4: { value: 0, title: __('Critical') },
+    total: 0,
+  };
```
amirfefer, line 27: looks like it a constant object, could you extract it out from the component?

webpack/InsightsHostDetailsTab/InsightsTotalRiskChart.js
```diff
@@ -0,0 +1,155 @@
...
+
+  useEffect(() => {
+    id &&
+      dispatch(
+        get({
+          key: API_KEY,
+          url: insightsCloudUrl(`hits/${id}`),
+        })
+      );
+  }, [API_KEY, dispatch, id]);
```
amirfefer, line 44: nit: how about using `useAPI` instead?

webpack/InsightsHostDetailsTab/InsightsTotalRiskChart.js
```diff
@@ -0,0 +1,155 @@
...
+          eventHandlers: {
+            onClick: onChartClick,
+          },
+        },
+      ]}
+    />
+  );
+
+  if (status === STATUS.pending) {
+    cardBody = <Spinner isSVG />;
```
amirfefer, line 130: do we want to keep the skeleton loading consistency on this page?
we can reuse `SkeletonLoader`

webpack/InsightsHostDetailsTab/InsightsTotalRiskChart.js
```diff
@@ -0,0 +1,155 @@
...
+  }
+
+  if (status === STATUS.ERROR) {
+    cardBody = <ErrorState error={error} />;
+  }
```
amirfefer, line 135: do we want to keep the skeleton loading consistency on this page?
we can reuse `SkeletonLoader` component for this behavior

webpack/InsightsHostDetailsTab/InsightsTotalRiskChart.js
```diff
@@ -0,0 +1,155 @@
...
+        get({
+          key: API_KEY,
+          url: insightsCloudUrl(`hits/${id}`),
+        })
+      );
+  }, [API_KEY, dispatch, id]);
+
+  // eslint-disable-next-line consistent-return
+  useEffect(() => {
+    if (!hits || !hits.length) return setTotalRisks(initialRisks);
```
amirfefer, line 48: do we need to set the total risk  to the  initial values if it has already been set (in line 28)

webpack/InsightsHostDetailsTab/InsightsTotalRiskChart.js
```diff
@@ -0,0 +1,155 @@
...
+    const risks = { ...initialRisks };
+    // eslint-disable-next-line camelcase
+    hits.forEach(({ total_risk }) => {
```
amirfefer, line 51: nit: you can rename the prop while object destructing for avoiding the lint issue

webpack/InsightsHostDetailsTab/InsightsTotalRiskChart.js
```diff
@@ -0,0 +1,155 @@
...
+      risks[total_risk].value += 1;
+    });
+    risks.total = hits.length;
+    setTotalRisks(risks);
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [hits.length]);
```
amirfefer, line 57: In general, using the length as a dependency might create unexpected bugs, i.e the array can be at the same size but with different values, that's why eslint's `exhaustive-deps`  is angry :)

Instead, we can use useSelector's `shallowEqual`
```js
import { useSelector, shallowEqual } from 'react-redux';

const { hits = [] } = useSelector(state => selectAPIResponse(state, API_KEY), shallowEqual);
```

amirfefer (changes requested):
Thanks @Ron-Lavi ! Looks great 👍 

I left a few inline comments

Ron-Lavi:
An issue is remaining with the useEffect being called too often and the number of total risks calculated is wrong

In addition need to add proper loading skeleton and error state

Ron-Lavi:
Now works as expected,
@ShimShtein @amirfefer the PR is ready for another review

https://user-images.githubusercontent.com/26363699/143743769-90b697ef-fee3-4f48-b10b-a13a7e1d8bd4.mp4

ShimShtein (changes requested):

webpack/InsightsHostDetailsTab/InsightsTab.scss
```diff
@@ -88,3 +88,75 @@
 #new_host_details_insights_tab {
   padding: 24px;
 }
+
+$first-child-fill: var(--pf-global--palette--blue-100) !important;
```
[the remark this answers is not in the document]
Ron-Lavi: I am not sure the CSS rule can be more specific than https://github.com/theforeman/foreman_rh_cloud/pull/661/files#diff-2fe4f4562fe10e9b91389c934af9b648897009d41173e49bcd56793bc774927cR148 

I think it's somehow related to the nature of SVGs and the way Victory.js loads it that without `!important` didn't work.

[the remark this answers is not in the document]
Ron-Lavi: yep, Victory.js puts a style tag on the element itself, so no other way to override it, e.g:
```html
<path d="M 227.128, 151.777
      h9.744000000000028
      v-9.744000000000028
      h-9.744000000000028
      z" role="presentation" shape-rendering="auto" style="fill: rgb(81, 157, 233);"></path>
```

Ron-Lavi:
@ShimShtein I added the theme props directly to the chart and chart's legend, can you take another look?

Ron-Lavi:
@ShimShtein there is an issue in the test with foreman_ansible: `foreman_ansible plugin requires Foreman >= 3.0 but current is 2.5.4` should we create a new branch for foreman_3_1 and pin foreman_ansible version there?

Ron-Lavi:
@ShimShtein, just rebased, thanks for your tests setup 👍🏼

ShimShtein:
When there is only one item to show, the checkboxes are displayed incorrectly:

image

Ron-Lavi:
Thanks @ShimShtein, it's a known issue in PF when no rows have a checkbox the main checkbox is half-baked,
I guess that's the reason by design they want to move the main checkbox to the toolbar above

Ron-Lavi:
Thanks @ShimShtein, it's a known issue in PF when no rows have a checkbox the main checkbox is half-baked,
I guess that's the reason by design they want to move the main checkbox to the toolbar above

ShimShtein:
I would really like to get rid of the direct css and use `theme` property instead.
Besides that - it looks great!

webpack/InsightsHostDetailsTab/NewHostDetailsTab.js
```diff
@@ -20,7 +20,6 @@ import {
   selectSearch,
   selectHits,
 } from '../InsightsCloudSync/Components/InsightsTable/InsightsTableSelectors';
-import './InsightsTab.scss';
```
ShimShtein, line 23: Should we also remove the scss file itself?

Ron-Lavi: it's still needed for the insights tab in the old host page, when it'll be gone we can remove these files.

ShimShtein:
Merged, thanks @Ron-Lavi!

PR #673, opened 2021-12-14
Add permissions test and fix missing permissions

ShimShtein:


config/routes.rb
```diff
@@ -18,7 +18,7 @@
   namespace :insights_cloud do
     resources :tasks, only: [:create]
     resource :settings, only: [:show, :update]
-    resources :hits, except: %i[show] do
+    resources :hits, except: [:show, :edit, :destroy, :update, :create, :new] do
```
Ron-Lavi, line 21: should we switch to `only` ?

ShimShtein: Didn't try `only` with more definitions inside

ShimShtein:
@Ron-Lavi ready for another round

config/routes.rb
```diff
@@ -18,7 +18,7 @@
   namespace :insights_cloud do
     resources :tasks, only: [:create]
     resource :settings, only: [:show, :update]
-    resources :hits, except: %i[show] do
+    resources :hits, only: [:index, :show] do
```
Ron-Lavi, line 21:
```suggestion
    resources :hits, only: [:index] do
```

with `show` included there is an issue in the old host page: `ActionController::ParameterMissing in InsightsCloud::HitsController#show`

ShimShtein:
@Ron-Lavi yup, missed it. Should be OK now

Ron-Lavi (approved):
Thanks @ShimShtein !

PR #676, opened 2021-12-22
Add RHC endpoint

ShimShtein:
This PR adds an endpoint that will be triggered by the `rhc-cloud-connector-worker` in order to execute remediation playbook.

the message format for these requests is:
``` json
{
    "type": "data",
    "message_id": "a6a7d866-7de0-409a-84e0-3c56c4171bb7",
    "version": 1,
    "sent": "2021-01-12T15:30:08+00:00",
    "directive": "playbook-sat",
    "metadata": {
        "operation": "run",
        "return_url": "https://cloud.redhat.com/api/v1/ingres/upload",
        "correlation_id": "6684b9dd-0d16-42c1-b13a-9f45be59e3b6",
        "playbook_run_name": "Human-readable playbook run name",
        "playbook_run_url": "https://console.redhat.com/insights/remediations/1234",
        "sat_id": "aa3b1faa-56f3-4d14-8258-615d11e20060",
        "sat_org_id": "12345",
        "initiator_user_id": "4efca34c6d9ae05ef7c3d7a7424e6370d198159a841ae005084888a9a4529e27",
        "hosts": "01.example.com,02.example.com",
        "response_interval": "30",
        "response_full": "false"
    },
    "content": "https://cloud.redhat.com/api/v1/remediations/1234/playbook"
}
```

It adds a new route: `POST api/v2/rh_cloud/cloud_request`
This request would be processed by the cloud request controller. Since the message is generic, a dispatching mechanism would decide which controller method is responsible for handling the specific message according to the `directive` field.

Currently only one directive is implemented: `playbook-sat` which is mapped to "Download and execute a playbook n specific hosts" use case.

The handler creates a REX job with the playbook URL extracted from the message as a parameter. This job will invoke the `rh_cloud_download_playbook` template. The template is using a newly introduced helper that is responsible for retrieving the playbook according to the passed URI and returning plain playbook text.
Under the hood, the template calls a service class `UrlRemediationsRetriever < RemediationRetriever`. The base class is responsible to get a playbook according to the parameters set in `playbook_url`, `headers`, `payload` and `method` and validate it. The derived class is responsible to set up the relevant request information and pass it to the base class.


ShimShtein:
@ehelms This is the counterpart that works with https://github.com/ShimShtein/rhc-cloud-connector-worker/pull/2

app/views/job_templates/rh_cloud_download_playbook.erb
```diff
@@ -0,0 +1,14 @@
...
+description_format: 'Playbook generated by RH Cloud'
+feature: rh_cloud_connector_run_playbook
+template_inputs:
+- name: playbook_url
+  description: URL of the playbook to run
+  input_type: user
+  required: true
+provider_type: Ansible
+%>
+<%= download_rh_playbook(input('playbook_url')) %>
```
ehelms, line 14: When and where does this mean the download of the playbook happens?

app/services/foreman_rh_cloud/template_renderer_helper.rb
```diff
@@ -15,7 +15,16 @@ module TemplateRendererHelper
...
-      retriever = ForemanRhCloud::RemediationsRetriever.new(hit_remediation_pairs, logger: template_logger)
+      retriever = ForemanRhCloud::HitRemediationsRetriever.new(hit_remediation_pairs, logger: template_logger)
+      retriever.create_playbook
+    end
+
+    apipie :method, 'Returns a Red Hat remediation playbook compiled on console.redhat.com' do
+      required :remediation_path, String, desc: ''
+      returns String, desc: 'Playbook downloaded from the cloud'
+    end
+    def download_rh_playbook(playbook_url)
```
ehelms, line 26: Where do you think you would do playbook signature validation?

app/services/foreman_rh_cloud/url_remediations_retriever.rb
```diff
@@ -0,0 +1,33 @@
...
+    attr_reader :url, :payload, :headers
+
+    def initialize(url:, payload: '', headers: {}, logger: Logger.new(IO::NULL))
+      super(logger: logger)
+
+      @url = url
+      @payload = payload
+      @headers = headers
+
+      logger.debug("Querying playbook at: #{url} with payload: #{payload} and headers: #{headers}")
```
ehelms, line 12: Does this method actually perform the query? If not, I would think of moving this closer to the actual retrieval call.

app/services/foreman_rh_cloud/template_renderer_helper.rb
```diff
@@ -15,7 +15,16 @@ module TemplateRendererHelper
...
+      retriever = ForemanRhCloud::UrlRemediationsRetriever.new(url: playbook_url, logger: template_logger)
```
ehelms, line 27: If there is an error querying for the playbook -- where and how does that get logged?

ehelms:
Could you add a comment that shows the rough high level workflow? e.g. the API that is being hit by the worker and the payload it is passing?

app/views/job_templates/rh_cloud_download_playbook.erb
ShimShtein: the `download_rh_playbook` method will do the downloading once invoked.

app/services/foreman_rh_cloud/template_renderer_helper.rb
ShimShtein: Inside this method. Either directly, or through a concern (if I want to reuse the logic in other places too)

app/services/foreman_rh_cloud/template_renderer_helper.rb
ShimShtein: It would be logged through the `template_logger` that is part of template rendering mechanism

ShimShtein:
@adamruzicka, Could you please review the progress reporting part? It's in https://github.com/theforeman/foreman_rh_cloud/pull/676/commits/5ed7d92b6426d04a47649969deb964aae58777ee

adamruzicka (commented):
Also there are some things in the specification document which are not yet implemented here, should I point them out specifically?

app/services/foreman_rh_cloud/template_renderer_helper.rb
adamruzicka, line 26: I'd suggest an early return if the render is done as a preview. Also it would be a good idea to cache the playbook to avoid re-downloading the same playbook for each host separately.

We already did that for the first iteration of cloud integration https://github.com/theforeman/foreman_ansible/blob/master/app/services/foreman_ansible/renderer_methods.rb#L20-L22

ShimShtein:
@adamruzicka  ready for another round

ShimShtein:
@adamruzicka @ehelms finally this is ready for review. Would you mind giving it another round?

app/controllers/api/v2/rh_cloud/cloud_request_controller.rb
```diff
@@ -0,0 +1,88 @@
...
+      render json: {
+        :message => "Handled #{directive} by #{handler}",
+      }
+    end
+
+    def server_features
+      render json: {
+        features: {},
+      }
+    end
```
adamruzicka, line 31: Where is this used?

app/controllers/api/v2/rh_cloud/cloud_request_controller.rb
```diff
@@ -0,0 +1,78 @@
...
+
+    private
+
+    def metadata
+      params['metadata']
+    end
+
+    def content
+      # the content received as base 64 of the string in double quotes
+      Base64.decode64(params['content']).tr('"', '')
```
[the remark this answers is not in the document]
adamruzicka: Have you checked it against an actual message?

app/controllers/api/v2/rh_cloud/cloud_request_controller.rb
```diff
@@ -0,0 +1,88 @@
...
+      logger.error("Reporting URL is not valid: #{metadata['return_url']}") && return unless valid_url?(metadata['return_url'])
+
+      hosts = metadata['hosts'].split(',')
+      host_ids = host_ids(hosts)
+
+      logger.warning("Some hosts were not found. Looked for: #{hosts}, found ids: #{host_ids}") unless host_ids.length == hosts.length
+
+      logger.error("sat_org_id is not present in the metadata") && return unless metadata['sat_org_id']
+      org_id = metadata['sat_org_id'].to_i
+      organization = Organization.find(org_id)
```
adamruzicka, line 59: Shouldn't the org be taken into account when resolving hosts? Just in case

app/models/setting/rh_cloud.rb
```diff
@@ -20,6 +20,7 @@ def self.default_settings
       set('rh_cloud_token', N_('Authentication token to Red Hat cloud services. Used to authenticate requests to cloud APIs'), nil, N_('Red Hat Cloud token'), nil, encrypted: true),
       set('exclude_installed_packages', N_('Exclude installed packages from being uploaded to the Red Hat cloud'), false, N_("Exclude installed Packages")),
       set('include_parameter_tags', N_('Should import include parameter tags from Foreman?'), false, N_('Include parameters in insights-client reports')),
+      set('rhc_instance_id', N_('RHC daemon id'), nil, N_('ID of the RHC(Yiggdrasil) daemon')),
```
adamruzicka, line 23: It is just `Yggdrasil`. How should I understand this? Will there be multiple instances of RHC running on the Satellite?

lib/insights_cloud/async/connector_playbook_execution_reporter_task.rb
```diff
@@ -0,0 +1,192 @@
...
+      end
+
+      def self.connector_feature_id
+        @connector_feature_id ||= RemoteExecutionFeature.feature!(:rh_cloud_connector_run_playbook).id
+      end
+
+      def plan(job_invocation)
+        return unless connector_playbook_job?(job_invocation)
+
+        @job_invocation = job_invocation
```
adamruzicka, line 18: Why though?

lib/insights_cloud/async/connector_playbook_execution_reporter_task.rb
```diff
@@ -0,0 +1,192 @@
...
+      end
+
+      def task_status(host_task, host_name)
+        unless host_task
+          return { 'state' => 'unknown' }
+        end
+
+        {
+          'state' => host_task.state,
+          'output' => host_task.main_action.live_output.map { |line| line['output'] }.join("\n"),
```
adamruzicka, line 129: To be honest I'm not sure if the live output is sorted by default or not

lib/insights_cloud/generators/playbook_progress_generator.rb
```diff
@@ -0,0 +1,49 @@
...
+        }
+      end
+
+      def host_finished_message(host_name, exit_code)
+        @messages << {
+          "type": "playbook_run_finished",
+          "version": 3,
+          "correlation_id": correlation_id,
+          "host": host_name,
+          "status":  exit_code == 0 ? 'success' : 'failure',
```
adamruzicka, line 27:
```suggestion
          "status": exit_code == 0 ? 'success' : 'failure',
```

lib/tasks/insights.rake
```diff
@@ -19,4 +19,20 @@ namespace :rh_cloud_insights do
...
+  desc "Re-announce all organizations into Sources on RH cloud."
+  task announce_to_sources: [:environment] do
+    logger = Logging::Logger.new(STDOUT)
+    Organization.unscoped.each do |org|
+      presence = ForemanRhCloud::CloudPresence.new(org, logger)
+      presence.announce_to_sources
+    rescue StandardError => ex
+      require 'pry'
+      binding.pry
+
```
adamruzicka, line 32: This most likely shouldn't be here

lib/insights_cloud/async/connector_playbook_execution_reporter_task.rb
ShimShtein: I think it is: https://github.com/theforeman/foreman_remote_execution/blob/0292f908250bb10275138e021e879267885d220d/app/lib/actions/remote_execution/run_host_job.rb#L105

app/controllers/api/v2/rh_cloud/cloud_request_controller.rb
ShimShtein: It's part of the contract that all hosts belong to a single org.

app/models/setting/rh_cloud.rb
ShimShtein: It's a single instance that runs on the host.

lib/insights_cloud/async/connector_playbook_execution_reporter_task.rb
ShimShtein: To cache the job invocation instance. Otherwise, the code will re-fetch the instance that I already have.

adamruzicka:
At least some of the test failures seem related

ShimShtein:
Finally the tests are green! Who wants to push the button?

app/controllers/api/v2/rh_cloud/cloud_request_controller.rb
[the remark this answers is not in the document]
ShimShtein: Yes, I have tried end to end with this algo.

app/models/setting/rh_cloud.rb
ShimShtein: We don't know its identifier. It's an ID derived from certificates used for the authentication.
We need this id to pass it on to Sources.

adamruzicka (approved):
Looks good to me. @ehelms could we get your blessing as well?

ehelms (approved):

app/controllers/api/v2/rh_cloud/cloud_request_controller.rb
```diff
@@ -0,0 +1,82 @@
...
+    end
+
+    def handle_run_playbook_request
+      logger.error("Playbook URL is not valid: #{content}") && return unless valid_url?(content)
+      logger.error("Reporting URL is not valid: #{metadata['return_url']}") && return unless valid_url?(metadata['return_url'])
+
+      hosts = metadata['hosts'].split(',')
+      host_ids = host_ids(hosts)
+
+      logger.warning("Some hosts were not found. Looked for: #{hosts}, found ids: #{host_ids}") unless host_ids.length == hosts.length
```
Ron-Lavi, line 49:
```suggestion
      logger.warn("Some hosts were not found. Looked for: #{hosts}, found ids: #{host_ids}") unless host_ids.length == hosts.length
```

app/controllers/api/v2/rh_cloud/cloud_request_controller.rb
```diff
@@ -0,0 +1,82 @@
...
+
+      logger.error("sat_org_id is not present in the metadata") && return unless metadata['sat_org_id']
+      org_id = metadata['sat_org_id'].to_i
+      organization = Organization.find(org_id)
+
+      composer = nil
+      Organization.as_org(organization) do
+        composer = ::JobInvocationComposer.for_feature(
+          :rh_cloud_connector_run_playbook,
```
Ron-Lavi, line 58: probably I missed something in my setup,
I am getting the error: `"ERF42-5962 [Foreman::Exception]: No template mapped to feature Run RH Cloud playbook"`

when debugging,I see the REX feature is: 
```rb
21:08:46 rails.1   |  id: 25,
21:08:46 rails.1   |  label: "rh_cloud_connector_run_playbook",
21:08:46 rails.1   |  name: "Run RH Cloud playbook",
21:08:46 rails.1   |  description: "Run playbook genrated by Red Hat remediations app",
21:08:46 rails.1   |  provided_inputs: "playbook_url,report_url,correlation_id,report_interval",
21:08:46 rails.1   |  job_template_id: nil,
21:08:46 rails.1   |  host_action_button: false,
21:08:46 rails.1   |  notification_builder: nil,
21:08:46 rails.1   |  proxy_selector_override: nil>
```
but since `job_template` is `nil` it Immediately fails.
REX code for reference: https://github.com/theforeman/foreman_remote_execution/blob/master/app/models/job_invocation_composer.rb#L353

Ron-Lavi: the `seed` worked and now I have `job_template_id`
there was an old version of another plugin that prevented the seed.

app/services/foreman_rh_cloud/cloud_presence.rb
```diff
@@ -0,0 +1,123 @@
...
+          method: :post,
+          url: create_rhc_connections_url,
+          headers: {
+            content_type: :json,
+          },
+          ssl_client_cert: OpenSSL::X509::Certificate.new(certs[:cert]),
+          ssl_client_key: OpenSSL::PKey::RSA.new(certs[:key]),
+          payload: {
+            source_id: source_id,
+            rhc_id: Setting[:rhc_instance_id],
```
Ron-Lavi, line 89: should we return some error to the user if `Setting[:rhc_instance_id]` is `nil` ?

Ron-Lavi (commented):
Thanks @ShimShtein !
tested and the flow works nicely, great work!
left a few inline comments.

ShimShtein:
@Ron-Lavi I think I have addressed all your comments. Can you give it another spin?

Ron-Lavi (approved):
Thanks @ShimShtein !!

PR #677, opened 2022-01-05
Build container action

ShimShtein:
@amirfefer @sharvit @Ron-Lavi 
This introduces and builds a container that we can use for testing plugins.
Any comments? 
You don't have to build your container locally, since it's already published here: https://github.com/ShimShtein/foreman_rh_cloud/pkgs/container/tfm_plugin_test


Ron-Lavi (approved):
LGTM @ShimShtein, thanks!!

PR #680, opened 2022-01-13
[CP 2.3] Truncate tags to 250 chars

ShimShtein:


ShimShtein:
Built the container locally and tested it. The tests passed.

PR #689, opened 2022-01-25
fix infinite loop on total risk chart error

Ron-Lavi:
found that when there's a network error or host without recommendation, it may cause an infinite loop in the use-effect.


ShimShtein (approved):
ACK, works as expected

PR #696, opened 2022-02-21
BZ #1979092 - Use infrastructure facet to determine foreman host

adamruzicka:


adamruzicka:
I admit this is not the most optimal way, but the whole "find host by name, alternatively by something completely different which can also raise an exception and then fall back to name again" was a bit unwieldy, but I can go that way if you want.

PR #697, opened 2022-03-01
recommendation link to correct host page version

Ron-Lavi:
the `host_details_ui` setting should determine the correct host path to redirect.


ShimShtein:
Merged, thanks @Ron-Lavi!

PR #699, opened 2022-03-08
Fix redhat-cloud-services/frontend-components scss import

Ron-Lavi:
@redhat-cloud-services/frontend-components/index.css will be exported to Foreman via https://github.com/theforeman/foreman-js/pull/387


ShimShtein:
Merged, thanks @Ron-Lavi

Ron-Lavi:
I think we need to revert the merge until this is added to foreman-js

PR #710, opened 2022-03-30
BZ2069640 - prevent hits#show network errors

Ron-Lavi:
When accessing the overview tab on the new host page,
there are two network error responses with a 500 status.
the first one which tries the endpoint `/insights_cloud/hits/3` is because that specific host doesn't have recommendations.

and the second which comes right after that is failing due to the path ends with `/insights_cloud/hits/undefined`

this doesn't cause errors to be seen on the UI,
but rather an API call that could be avoided.

### DIFF
--- a/app/controllers/insights_cloud/hits_controller.rb
+++ b/app/controllers/insights_cloud/hits_controller.rb
@@ -14,9 +14,16 @@ module InsightsCloud
 
     def show
       host = Host.where(id: host_id_param).first
+      hits = host.insights&.hits
+
+      unless hits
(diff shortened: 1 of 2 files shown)


ShimShtein:
I have fixed the tests, could you please rebase?

Ron-Lavi:
updated, thanks

ShimShtein:
@Ron-Lavi Another rebase please? This time I hope it will work :)

Ron-Lavi:
it's green now :)

PR #716, opened 2022-04-11
Properly treat caching headers

ShimShtein:

PR #720, opened 2022-04-13
test: add test for missing etag header

subpop:
Adds a test to ensure the ETag header is in the response when the
request contains the If-None-Match header.


subpop:
Note: this test will fail until #716 is merged.

ShimShtein:
@subpop , can you rebase please? #716 is already merged

PR #722, opened 2022-04-22
Fix sentence case for Total risks card

jeremylenz:
Change card title from "Total Risks" to "Total risks"

that's all!


ShimShtein:
Merged, thanks @jeremylenz !

PR #732, opened 2022-05-11
Set cloud connector http_proxy if globally configured

ehelms:


app/services/foreman_rh_cloud/cloud_connector.rb
```diff
@@ -12,11 +12,20 @@ def install
       target_host = foreman_host
       composer = nil
 
+      input = {
+        :satellite_cloud_connector_user => service_user.login,
+        :satellite_cloud_connector_password => token_value,
+      }
+
+      if (http_proxy = Setting[:http_proxy])
```
ShimShtein, line 20: You should use [`ForemanRhCloud.proxy_setting(logger: logger)`](https://github.com/theforeman/foreman_rh_cloud/blob/2519e2fff3dd08b4e87663cb9a465ed98d6856f8/lib/foreman_rh_cloud.rb#L40) to conform with proxy selection logic for the rest of rh_cloud plugin.

ShimShtein (approved):

ShimShtein (changes requested):
Sorry, hit the wrong button. We need to change the proxy string source.

app/services/foreman_rh_cloud/cloud_connector.rb
```diff
@@ -12,11 +12,20 @@ def install
...
+        input[:satellite_cloud_connector_http_proxy] = ForemanRhCloud.proxy_setting(logger: Foreman::Logging.logger('app'))
```
ShimShtein, line 21:
```suggestion
      if (http_proxy = ForemanRhCloud.proxy_setting(logger: Foreman::Logging.logger('app')))
        input[:satellite_cloud_connector_http_proxy] = http_proxy
```

We should use the same data source for both the condition and the assignment. Otherwise you will have edge cases where the condition would be falsy, but the proxy would be required.

app/views/job_templates/cloud_connector.erb
```diff
@@ -14,6 +14,12 @@ template_inputs:
   advanced: false
   value_type: plain
   hidden_value: true
+- name: http_proxy
```
ShimShtein, line 17: @adamruzicka, is it enough to add a parameter to the template to make it available to the underlying playbook?

ShimShtein (changes requested):

app/views/job_templates/cloud_connector.erb
adamruzicka: It should be, as long as the role looks for a variable with exactly the same name

ShimShtein:
Merged, thanks @ehelms !

PR #738, opened 2022-06-03
Adjust card weight so order is correct

jeremylenz:
Requires https://github.com/theforeman/foreman/pull/9251
To test, must rebase this on https://github.com/theforeman/foreman_rh_cloud/pull/727

See testing instructions on https://github.com/theforeman/foreman/pull/9251


webpack/ForemanRhCloudFills.js
```diff
@@ -18,10 +18,10 @@ const fills = [
     weight: 400,
   },
   {
-    slot: 'details-cards',
+    slot: 'host-overview-cards',
```
ShimShtein, line 21: Which version of Foreman changed the slot name?

jeremylenz: @ShimShtein The Foreman change https://github.com/theforeman/foreman/pull/9189 is linked to https://projects.theforeman.org/issues/33010 which is "Fixed in Releases" 3.3.0.

jeremylenz: btw, if you use the old slot name, the card weight won't work properly and the card will always be last. That's why both changes are here.

chris1984 (approved):
ACK https://nimb.ws/beGcrI

Ron-Lavi (approved):
LGTM, thanks @jeremylenz !
Once we will move forward with the master branch, it will be good to merge this.

jeremylenz:
@ShimShtein oh hi! Been a while.

rebased :slightly_smiling_face:

ShimShtein:
Merged, thanks @jeremylenz !

PR #742, opened 2022-06-09
Move sources announcement after configuration job

ShimShtein:
@adamruzicka Can you take a look?

### DIFF
--- a/lib/insights_cloud/async/cloud_connector_announce_task.rb
+++ b/lib/insights_cloud/async/cloud_connector_announce_task.rb
@@ -15,10 +15,7 @@ module InsightsCloud
         plan_self
       end
 
-      def run(event = nil)
-        # Handle skip events
-        return if event == Dynflow::Action::Skip
-
+      def finalize(*_args)
         Organization.unscoped.each do |org|
           presence = ForemanRhCloud::CloudPresence.new(org, logger)
           presence.announce_to_sources
--- a/lib/insights_cloud/async/cloud_connector_announce_task.rb
+++ b/lib/insights_cloud/async/cloud_connector_announce_task.rb
@@ -1,0 +1,47 @@
+module InsightsCloud
+  module Async
+    class CloudConnectorAnnounceTask < ::Actions::EntryAction
+      def self.subscribe
+        Actions::RemoteExecution::RunHostsJob
+      end
+
+      def self.connector_feature_id
+        @connector_feature_id ||= RemoteExecutionFeature.feature!(ForemanRhCloud::CloudConnector::CLOUD_CONNECTOR_FEATURE).id
+      end
+
+      def plan(job_invocation)
+        return unless connector_playbook_job?(job_invocation)
+
+        plan_self
+      end
+
+      def run(event = nil)
+        # Handle skip events
+        return if event == Dynflow::Action::Skip
+
+        Organization.unscoped.each do |org|
+          presence = ForemanRhCloud::CloudPresence.new(org, logger)
+          presence.announce_to_sources
+        rescue StandardError => ex
+          logger.warn(ex)
+        end
+      end
+
+      def rescue_strategy_for_self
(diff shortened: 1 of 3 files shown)


lib/insights_cloud/async/cloud_connector_announce_task.rb
```diff
@@ -0,0 +1,47 @@
...
+        @connector_feature_id ||= RemoteExecutionFeature.feature!(ForemanRhCloud::CloudConnector::CLOUD_CONNECTOR_FEATURE).id
+      end
+
+      def plan(job_invocation)
+        return unless connector_playbook_job?(job_invocation)
+
+        plan_self
+      end
+
+      def run(event = nil)
```
adamruzicka, line 18: To be honest I'm not sure how subscribed actions behave. Wouldn't this run in parallel to `RunHostsJob` that triggered this one?

lib/insights_cloud/async/cloud_connector_announce_task.rb
```diff
@@ -0,0 +1,47 @@
...
+        # Handle skip events
+        return if event == Dynflow::Action::Skip
```
adamruzicka, line 20: You could make `#run` accept no arguments and then you could drop this line and it should behave exactly the same way

ShimShtein:
@adamruzicka Now it should be better

lib/insights_cloud/async/cloud_connector_announce_task.rb
```diff
@@ -0,0 +1,44 @@
+module InsightsCloud
+  module Async
+    class CloudConnectorAnnounceTask < ::Actions::EntryAction
+      def self.subscribe
+        Actions::RemoteExecution::RunHostsJob
+      end
+
+      def self.connector_feature_id
+        @connector_feature_id ||= RemoteExecutionFeature.feature!(ForemanRhCloud::CloudConnector::CLOUD_CONNECTOR_FEATURE).id
```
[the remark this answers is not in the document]
ShimShtein: It would be an error state, since I add this feature in the [engine](https://github.com/theforeman/foreman_rh_cloud/blob/65660aaa793d8c64d331095d7b825db06bfd6d3b/lib/foreman_rh_cloud/engine.rb#L163)

lib/insights_cloud/async/cloud_connector_announce_task.rb
```diff
@@ -0,0 +1,44 @@
...
+      end
+
+      def plan(job_invocation)
+        return unless connector_playbook_job?(job_invocation)
+
+        plan_self
+      end
+
+      def finalize(*_args)
```
adamruzicka, line 18: `#finalize` never receives any arguments

lib/insights_cloud/async/cloud_connector_announce_task.rb
```diff
@@ -0,0 +1,44 @@
...
+          logger.warn(ex)
+        end
+      end
+
+      def rescue_strategy_for_self
+        Dynflow::Action::Rescue::Skip
+      end
+
+      def connector_playbook_job?(job_invocation)
+        job_invocation&.remote_execution_feature_id == connector_feature_id
```
adamruzicka, line 32: This would not work if someone went and ran the template directly. Is that acceptable? Otherwise we'd have to go through rex features assigned to the job's template.

adamruzicka (commented):
One nit and one question, otherwise lgtm

lib/insights_cloud/async/cloud_connector_announce_task.rb
ShimShtein: Yes, it's acceptable. If someone is configuring cloud connector manually, we can assume he knows what he's doing and will run the announce rake task later on.

adamruzicka (approved):
Disclaimer: I don't have a dev env with RH cloud

But I'd say ship it :shipit:

PR #745, opened 2022-06-10
job_finished_message should indicate overall playbook status

ehelms:
consoleDot expects the `playbook_run_completed` status to indicate if the playbook succeeded for all hosts or indicate failure if at least one host failed.

@tahmidefaz @aleccohan -- could you verify this looks as you are expecting?


tahmidefaz (approved):
Looks good to me! Seems like you are flipping the success/failure status corretly. :+1:

Ichimonji10:
Bug tracked in [https://issues.redhat.com/browse/SAT-10934](SAT-10934).

PR #748, opened 2022-06-13
use a space after `>=` for better parsing

evgeni:
rpm doesn't like the modifier not to be separated from the version without a space


Ron-Lavi (approved):

Ron-Lavi:
thanks @evgeni!

PR #750, opened 2022-07-06
Add support for LEAPP data downloads

ShimShtein:
LEAPP upgrade tool needs to download files from the cloud, so they need support for forwarding their requests through the server similar to insights-client.

### DIFF
--- a/app/services/foreman_rh_cloud/cloud_request_forwarder.rb
+++ b/app/services/foreman_rh_cloud/cloud_request_forwarder.rb
@@ -71,6 +71,12 @@ module ForemanRhCloud
           ssl_client_cert: OpenSSL::X509::Certificate.new(certs[:cert]),
           ssl_client_key: OpenSSL::PKey::RSA.new(certs[:key]),
         }
+      when pes_request?
+        {
+          url: ForemanRhCloud.cert_base_url + request_path,
+          ssl_client_cert: OpenSSL::X509::Certificate.new(certs[:cert]),
+          ssl_client_key: OpenSSL::PKey::RSA.new(certs[:key]),
+        }
       else # Legacy insights API
         {
           url: ForemanRhCloud.legacy_insights_url + request_path.sub('/redhat_access/r/insights', '/r/insights'),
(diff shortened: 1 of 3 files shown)


ehelms:
Thinking a bit about this from a dependency stand point, I tend to expect either:

 a) this to live in foreman_leapp with a dependency on foreman_rh_cloud
 b) foreman_rh_cloud to have a dependency on, or at least check that foreman_leapp is present to enable this set of functionality
 c) foreman_leapp to have a dependency on foreman_rh_cloud

It won't be obvious in the current form that there is some leapp related functionality over here in foreman_rh_cloud when trying to debug this or understand the workflow.

app/services/foreman_rh_cloud/cloud_request_forwarder.rb
```diff
@@ -99,6 +105,10 @@ def connection_test_request?
       ->(request_path) { request_path =~ /redhat_access\/r\/insights\/?$/ }
     end
 
+    def pes_request?
+      ->(request_path) { request_path =~ /api\/pes\// }
```
ekohl, line 109: I see it was already there before, but have you considered the `%r{}` notation?
```suggestion
      ->(request_path) { request_path =~ %r{api/pes/} }
```

Also note you only perform partial matching, so `/something/elseapi/pes/` will also match.

ekohl (commented):
> a) this to live in foreman_leapp with a dependency on foreman_rh_cloud

foreman_leapp lives in plugins, foreman_rh_cloud lives in and depends on katello. You can't have that dependency. Or it needs to be a soft dependency where foreman_leapp detects the presence of foreman_rh_cloud at runtime and enables functionality.

> c) foreman_leapp to have a dependency on foreman_rh_cloud

For the same reason, this will break repoclosure.

app/services/foreman_rh_cloud/cloud_request_forwarder.rb
ShimShtein: It comes with the routes filter, so `something/else/api/pes` wouldn't get here anyway.

evgeni:
I don't think `foreman_leapp` should be a trigger.

People can use Leapp w/o `foreman_leapp` on systems that are behind a Katello.

ShimShtein:
Converting this to draft, as the team asked us to wait with this feature

ShimShtein:
Closing, since it is not needed for too long

PR #751, opened 2022-07-06
Use generic openssl read key method

ShimShtein:
This helps supporting keys from all types, not only RSA


ShimShtein:
@evgeni Would you mind to review?

evgeni (approved):
With all my extensive* Ruby::OpenSSL knowledge: ACK

*: non-existant

PR #767, opened 2022-09-05
Set autoupload when cloud connector is requested

ShimShtein:


ShimShtein:
@Ron-Lavi can you take a look?

Ron-Lavi:
@ShimShtein does it need to be cherry-picked?

ShimShtein:
No, it's a new feature, we don't have to.

Ron-Lavi:
Well, we have the confirm modal in Foreman since Foreman 3.1 so it would be better to use that,
here's an example usage: https://github.com/theforeman/foreman/blob/122f5378a5fbdc6f1946f72c47069be59012681b/webpack/assets/javascripts/react_app/components/HostDetails/Status/StatusTable.js#L17

so on button click, you will:
```js
dispatch(
      openConfirmModal({
        title: 'some title',
        message: 'some message',
        isWarning: true, // better in our case
        onConfirm: () => {
          dispatch(configureCloudConnector());
        },
      })
    );
```

ShimShtein:
@Ron-Lavi what about now?

Ron-Lavi (commented):
The code looks good @ShimShtein 👍🏼 
We need to add a mock for the `openConfirmModal` in this directory: https://github.com/theforeman/foreman_rh_cloud/tree/master/webpack/__mocks__/foremanReact/components

Ron-Lavi (approved):
Waiting on the tests 🍏

PR #782, opened 2022-11-22
Use polling action to create exponential backoff

ShimShtein:


ShimShtein:
@adamruzicka I think it's finally ready for review

lib/foreman_inventory_upload/async/shell_process.rb
```diff
@@ -4,13 +4,27 @@ module ForemanInventoryUpload
...
+        poll_external_task_with_rescue
+        # supress unexpected task output serialization
+        {}
+      end
+
+      def poll_external_task
+        action
+        # supress unexpected task output serialization
+        {}
+      end
```
adamruzicka, line 25: This gets repeated all over the place. Any reason for being this explicit and not having it inside `::ForemanRhCloud::Async::ExponentialBackoff`?

ShimShtein:
@adamruzicka, removed the duplication and created a test class.

lib/insights_cloud/async/insights_full_sync.rb
```diff
@@ -28,7 +29,7 @@ def plan(organizations)
         end
       end
 
-      def run
+      def action
```
adamruzicka, line 32: While I'm going on about naming, this feels a bit unfortunate too. I mean a method called `action` on an action class does not really convey the intent

adamruzicka (commented):
The point I made earlier about me not being able to properly test this end to end sadly still holds. Code wise looks good, I don't expect this would cause any surprises. 

That being said, I could not resist leaving a note or two about naming, but as I made those with my pedantic hat on, I can just take the hat off and we can move forward.

ShimShtein:
@Ron-Lavi Can you take this one for a spin?

Ron-Lavi (approved):
Thanks @ShimShtein !!

I tested the inventory upload exponential calls by changing the `red_hat_inventory/uploads/uploader.sh` DEST link to something wrong and it worked well

I also, tested the insights full sync by setting the `ENV['SATELLITE_CERT_RH_CLOUD_URL']` to another URL that doesn't work and also so the exponential calls.

LGTM :)

PR #788, opened 2022-12-06
Bump container to foreman 3.5

ShimShtein:


adamruzicka:
Would be nice it it actually run the tests with the newly built container. Anyway, lgtm

adamruzicka:
Ah, even better, I didn't expect it there.

PR #789, opened 2022-12-07
Remove direct references to frontend-components

ShimShtein:


ShimShtein:
@chris1984 can you review this one? I had to reference `frontend_components` directly because foreman-js was not providing this package. Now that the package is part of foreman-js, I can remove the direct references.

chris1984 (approved):
Makes sense, looks good to me. ACK

PR #790, opened 2022-12-09
Add host searching based on insights statuses

ares:
This is adding missing search definitions which were forgotten when the statuses were introduced.


adamruzicka (approved):
I don't have a machine with rh cloud data, but the generated sql queries look correct. The `no_report` value feels a bit misleading, but there's not much we can do about it without

ACK

ShimShtein:
Merged, thanks @ares !

PR #791, opened 2023-01-04
Ensure Configure Cloud Connector runs as root

ekohl:
When a user has set the REX user as non-root, this task fails. Setting become to true means it will attempt to use sudo, which allows it to work in more cases.

Complete untested, just as an example of what I think could help.

### DIFF
--- a/app/views/job_templates/cloud_connector.erb
+++ b/app/views/job_templates/cloud_connector.erb
@@ -30,6 +30,7 @@ feature: ansible_configure_cloud_connector
 
 ---
 - hosts: all
+  become: true
   vars:
     satellite_cloud_connector_url: "<%= foreman_server_url %>"
   roles:


ShimShtein:
@adamruzicka this looks good to me, but does it need some REX configuration in order to work?

ekohl:
My reasoning was that there are 3 cases:
* Runs as root: currently safe, in my version also safe
* Runs as non-root, but has passwordless sudo: currently broken, fixed with this version
* Runs as non-root without sudo: currently broken, also broken in this version but at least with a clear description. It should also fail at the very start instead of possibly somewhere in between

So it should all stay at least the same, or be better for the user. That said: this came up in a team discussion and I thought this would clearly show what I meant rather than trying to describe it. I haven't looked at the actual source code and what it really does.

ShimShtein:
@ekohl can you please change the target branch to foreman_3_9? I'll merge it for the next foreman version.

ekohl:
I'd strongly advise against the current model of changing default branches. It's really disruptive to contributors. Git doesn't update the `HEAD` ref when you pull so tools like [hub](https://github.com/github/hub) aren't aware of it and continue to submit PRs against an older default branch. Changing the base branch on PRs is also very noisy and generally confuses GitHub Actions. So I'm closing this and opening a new one: https://github.com/theforeman/foreman_rh_cloud/pull/848

PR #792, opened 2023-01-04
Remove icon from toolbar buttons on inventory page

chris1984:
@ShimShtein let me know how this looks, I was able to get a devel box installed with the plugin and I don't see the icon anymore.

I switched the logic up based on this:

https://stackoverflow.com/a/70337425


ShimShtein:
Merged, thanks @chris1984 !

PR #798, opened 2023-01-30
Update report metadata to include system info

chris1984:
@ShimShtein Let me know how this looks?

### DIFF
--- a/lib/foreman_inventory_upload/generators/fact_helpers.rb
+++ b/lib/foreman_inventory_upload/generators/fact_helpers.rb
@@ -109,6 +109,18 @@ module ForemanInventoryUpload
         IPAddr.new(max_obfuscated + 1, Socket::AF_INET).to_s
       end
 
+      def hostname_match
+        bash_hostname = `uname -n`.chomp
+        foreman_hostname = ForemanRhCloud.foreman_host&.name
+        if bash_hostname == foreman_hostname
+          fqdn(foreman_hostname)
+        elsif Setting[:obfuscate_inventory_hostnames]
+          obfuscate_fqdn(bash_hostname)
+        else
+          bash_hostname
+        end
+      end
+
       def bios_uuid(host)
         value = fact_value(host, 'dmi::system::uuid') || ''
         uuid_value(value)
--- a/lib/foreman_inventory_upload/generators/metadata.rb
+++ b/lib/foreman_inventory_upload/generators/metadata.rb
@@ -1,6 +1,7 @@
 module ForemanInventoryUpload
   module Generators
     class Metadata
+      include FactHelpers
       def initialize(output = [])
         @stream = JsonStream.new(output)
       end
@@ -28,6 +29,9 @@ module ForemanInventoryUpload
           @stream.simple_field('report_id', Foreman.uuid)
           @stream.simple_field('host_inventory_api_version', '1.0')
           @stream.simple_field('source', 'Satellite')
+          @stream.simple_field('reporting_host_name', hostname_match)
+          @stream.simple_field('reporting_host_ips', host_ips(ForemanRhCloud.foreman_host))
+          @stream.simple_field('reporting_host_bios_uuid', bios_uuid(ForemanRhCloud.foreman_host))
           @stream.simple_field('source_metadata', metadata)
           @stream.object_field('report_slices', :last) do
             yield(self)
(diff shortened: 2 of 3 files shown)


lib/foreman_inventory_upload/generators/metadata.rb
```diff
@@ -27,7 +27,9 @@ def render_report(metadata)
         @stream.object do
           @stream.simple_field('report_id', Foreman.uuid)
           @stream.simple_field('host_inventory_api_version', '1.0')
-          @stream.simple_field('source', 'Satellite')
```
ShimShtein, line 30: Let's leave the `source` field as is. It can indicate whether rh_cloud is generating this report, or other services.

lib/foreman_inventory_upload/generators/metadata.rb
```diff
@@ -27,7 +27,9 @@ def render_report(metadata)
...
+          Setting[:obfuscate_inventory_hostnames] ? @stream.simple_field('source', 'Satellite') : @stream.simple_field('source', ForemanRhCloud.foreman_host.name)
```
ShimShtein, line 30: We can use our own helpers here:
``` ruby
include FactHelpers
# ...
@stream.simple_field('reporting_host_name', fqdn(ForemanRhCloud.foreman_host))
```

lib/foreman_inventory_upload/generators/metadata.rb
```diff
@@ -27,7 +27,9 @@ def render_report(metadata)
...
+          @stream.simple_field('IP address', ForemanRhCloud.foreman_host.facts['network::ipv4_address']) unless Setting[:obfuscate_inventory_ips]
```
ShimShtein, line 31: We can use our own helpers here:
``` ruby
include FactHelpers
# ...
@stream.simple_field('reporting_host_ips', host_ips(ForemanRhCloud.foreman_host).values.join(','))
```

lib/foreman_inventory_upload/generators/metadata.rb
```diff
@@ -27,7 +27,9 @@ def render_report(metadata)
...
+          @stream.simple_field('bios UUID', ForemanRhCloud.foreman_host.facts['dmi::system::uuid'])
```
ShimShtein, line 32:
```suggestion
          @stream.simple_field('reporting_host_bios_uuid', bios_uuid(ForemanRhCloud.foreman_host))
```

ShimShtein (changes requested):

chris1984:
@ShimShtein updated, let me know how this looks.

lib/foreman_inventory_upload/generators/metadata.rb
```diff
@@ -28,6 +29,9 @@ def render_report(metadata)
           @stream.simple_field('report_id', Foreman.uuid)
           @stream.simple_field('host_inventory_api_version', '1.0')
           @stream.simple_field('source', 'Satellite')
+          @stream.simple_field('reporting_host_name', fqdn(ForemanRhCloud.foreman_host)) unless Setting[:obfuscate_inventory_hostnames]
+          @stream.simple_field('reporting_host_ips', host_ips(ForemanRhCloud.foreman_host).values.join(',')) unless Setting[:obfuscate_inventory_ips]
```
ShimShtein, line 33: Both `fqdn` and `host_ips` methods are aware of the obfuscation. And since the obfuscation algorithm is known, there is actual value even if we send the obfuscated data.
We could ask the user to run obfuscation for all his host names and match to the one that is in this report, so the user will be able to locate the host.

ShimShtein (changes requested):
Almost there, small changes needed though

lib/foreman_inventory_upload/generators/metadata.rb
chris1984: @ShimShtein I am not sure what that change would look like, can you point to an example?

chris1984:
@ShimShtein how does this look?

lib/foreman_inventory_upload/generators/fact_helpers.rb
```diff
@@ -109,6 +109,18 @@ def obfuscate_ip(ip, ips_dict)
         IPAddr.new(max_obfuscated + 1, Socket::AF_INET).to_s
       end
 
+      def hostname_match
+        bash_hostname = `uname -n`.chomp
+        foreman_hostname = ForemanRhCloud.foreman_host
```
ShimShtein, line 114:
```suggestion
        foreman_hostname = ForemanRhCloud.foreman_host&.name
```
Otherwise the next `if` statement would always return `false`

ShimShtein (changes requested):

ShimShtein (approved):

ShimShtein:
Merged, thanks @chris1984!

PR #801, opened 2023-02-22
Csv export definition

adamruzicka:
Alternatively if we were ok with the label being 'Insights Hits - Count', then we could use the simplified variant using `export_key`. 

Requires:
- [ ] https://github.com/theforeman/foreman/pull/9618


ShimShtein:
I am waiting for 3.5 cherry-pick of the https://github.com/theforeman/foreman/pull/9618. Once it's cherry-picked, I will merge this PR. Otherwise it will cause the test suite to fail.

adamruzicka:
Could we get this in?

ShimShtein:
Merged. Thanks @adamruzicka!

PR #815, opened 2023-03-23
Update cloud_connector.erb

### LINKED ISSUE #814: Cloud Connector Configuration fails
I am trying to Configure the Cloud Connector using **Configure** / **Inventory Upload** / **Configure Cloud Connector** which invokes an Ansible Job run. This fails due to a missing role 'redhat.satellite_operations.cloud_connector'.
```
ERROR! the role 'redhat.satellite_operations.cloud_connector' was not found in /tmp/d20230323-802194-1r0ilbc/project/roles:/etc/ansible/roles:/usr/share/ansible/roles:/tmp/d20230323-802194-1r0ilbc/project
The error appears to be in '/tmp/d20230323-802194-1r0ilbc/project/playbook.yml': line 6, column 7, but may
be elsewhere in the file depending on the exact syntax problem.
The offending line appears to be:
  roles:
      - redhat.satellite_operations.cloud_connector
      ^ here
Exit status: 1
```

LiohMoeller:
Use theforeman.operations for Upstream

### DIFF
--- a/app/views/job_templates/cloud_connector.erb
+++ b/app/views/job_templates/cloud_connector.erb
@@ -33,4 +33,4 @@ feature: ansible_configure_cloud_connector
   vars:
     satellite_cloud_connector_url: "<%= foreman_server_url %>"
   roles:
-    - redhat.satellite_operations.cloud_connector
+    - theforeman.operations.cloud_connector
--- a/app/views/job_templates/cloud_connector.erb
+++ b/app/views/job_templates/cloud_connector.erb
@@ -31,6 +31,6 @@ feature: ansible_configure_cloud_connector
 ---
 - hosts: all
   vars:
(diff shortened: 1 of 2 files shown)


app/views/job_templates/cloud_connector.erb
```diff
@@ -33,4 +33,4 @@ feature: ansible_configure_cloud_connector
   vars:
     satellite_cloud_connector_url: "<%= foreman_server_url %>"
```
evgeni: you'll also have to update this to `foreman_cloud_connector_url`

see the readme at https://github.com/theforeman/foreman-operations-collection/tree/develop/roles/cloud_connector

app/views/job_templates/cloud_connector.erb
ekohl, line 1: The various template inputs are also branded, so they probably wouldn't work either.

PR #819, opened 2023-05-15
Add Migration for new missing hosts table

chris1984:
@ShimShtein here is the pr for the table create migration

### DIFF
--- a/app/models/insights_missing_hosts
+++ b/app/models/insights_missing_hosts
@@ -1,0 +1,2 @@
+class InsightsMissingHosts < ApplicationRecord
+end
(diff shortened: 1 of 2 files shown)


ShimShtein (changes requested):
Can you please rename the file to have `.foreman_rh_cloud.rb` suffix? It should enable to roll forward or backward only plugin's migrations.
Upd: I suppose a model class should be in this PR too.

ShimShtein:
We are missing an `organization_id` field. The host should belong to the org that queried the cloud inventory.

chris1984:
@ShimShtein updated.

ShimShtein (approved):

PR #821, opened 2023-05-19
Add backend changes for missing hosts

chris1984:
* Added route
* Added GET controller method for API
* Added new controller for UI with GET route
* Added permissions for routes
* Updated model with validations
* Fix model file name
* Fix migration (was not adding the `organization_id` column for some reason, this way it does)

Getting data back on get method :)
```ruby
19:21:52 rails.1   | 2023-05-19T19:21:52 [I|app|9d49580f] Started GET "/api/v2/organizations/3/rh_cloud/missing_hosts" for 192.168.196.1 at 2023-05-19 19:21:52 -0400
19:21:52 rails.1   | 2023-05-19T19:21:52 [I|app|9d49580f] Processing by Api::V2::RhCloud::InventoryController#get_hosts as JSON
19:21:52 rails.1   | 2023-05-19T19:21:52 [I|app|9d49580f]   Parameters: {"apiv"=>"v2", "organization_id"=>"3"}
19:21:52 rails.1   | 2023-05-19T19:21:52 [D|app|9d49580f] Body: {"id":1,"name":"foo.example.com","organization_id":3,"insights_id":"111-222","rhsm_id":"111-222","ip_address":"192.168.196.1"}
```

### DIFF
--- a/app/controllers/api/v2/rh_cloud/inventory_controller.rb
+++ b/app/controllers/api/v2/rh_cloud/inventory_controller.rb
@@ -28,6 +28,29 @@ module Api
           }, status: :ok
         end
 
+        api :GET, "/organizations/:organization_id/rh_cloud/missing_hosts", N_("Grab hosts that are missing in RH Cloud")
+        param :organization_id, Integer, required: true, desc: N_("Set the current organization context for the request")
+        def get_hosts
+          organization_id = params[:organization_id]
+          payload = InsightsMissingHosts.where(organization_id: organization_id)
+
+          render :json => payload
+        end
+
+        api :POST, "/organizations/:organization_id/rh_cloud/missing_hosts", N_("Grab hosts that are missing in RH Cloud")
+        param :organization_id, Integer, required: true, desc: N_("Set the current organization context for the request")
+        param :search_term, String, required: true, desc: N_("Scoped search string for host removal")
+        def remove_hosts
+          organization_id = params[:organization_id]
+          search_term = params[:search_term]
+
+          task = ForemanTasks.async_task(ForemanInventoryUpload::Async::RemoveInsightsHostsJob, search_term, organization_id)
+
+          render json: {
+            task: task,
+          }, status: :ok
+        end
+
         api :POST, "/organizations/:organization_id/rh_cloud/inventory_sync", N_("Start inventory synchronization")
         param :organization_id, Integer, required: true, desc: N_("Set the current organization context for the request")
         def sync_inventory_status
--- a/app/controllers/foreman_inventory_upload/missing_hosts_controller.rb
+++ b/app/controllers/foreman_inventory_upload/missing_hosts_controller.rb
@@ -1,0 +1,22 @@
+module ForemanInventoryUpload
+  class MissingHostsController < ::ApplicationController
+    def index
+      organizations = Organization.current || User.current.my_organizations
+      organization_id = organizations.pluck(:id)
+      payload = InsightsMissingHosts.where(organization_id: organization_id)
+
+      render :json => payload
+    end
+
+    def remove_hosts
+      organization_id = params[:organization_id]
+      search_term = params[:search_term]
+
+      task = ForemanTasks.async_task(ForemanInventoryUpload::Async::RemoveInsightsHostsJob, search_term, organization_id)
+
+      render json: {
+        task: task,
+      }, status: :ok
+    end
+  end
+end
--- a/lib/foreman_rh_cloud/engine.rb
+++ b/lib/foreman_rh_cloud/engine.rb
@@ -52,9 +52,10 @@ module ForemanRhCloud
             :generate_foreman_rh_cloud,
             'foreman_inventory_upload/reports': [:generate],
             'foreman_inventory_upload/tasks': [:create],
-            'api/v2/rh_cloud/inventory': [:sync_inventory_status, :download_file, :generate_report, :enable_cloud_connector],
+            'api/v2/rh_cloud/inventory': [:get_hosts, :remove_hosts, :sync_inventory_status, :download_file, :generate_report, :enable_cloud_connector],
             'foreman_inventory_upload/uploads': [:enable_cloud_connector],
             'foreman_inventory_upload/uploads_settings': [:set_advanced_setting],
+            'foreman_inventory_upload/missing_hosts': [:remove_hosts],
             'insights_cloud/settings': [:update],
             'insights_cloud/tasks': [:create]
           )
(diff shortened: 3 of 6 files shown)


app/models/insights_missing_hosts.rb
```diff
@@ -0,0 +1,4 @@
+class InsightsMissingHosts < ApplicationRecord
```
ShimShtein:
```suggestion
class InsightsMissingHost < ApplicationRecord
```
Models should be singular.

app/controllers/api/v2/rh_cloud/inventory_controller.rb
```diff
@@ -28,6 +28,15 @@ def generate_report
           }, status: :ok
         end
 
+        api :GET, "/organizations/:organization_id/rh_cloud/missing_hosts", N_("Grab hosts that are missing in RH Cloud")
```
ShimShtein, line 31: Unlike Katello, we have separated UI and API requests. I would suggest a new controller `MissingHostsController.rb` under the [`foreman_inventory_upload`](https://github.com/theforeman/foreman_rh_cloud/tree/foreman_3_7/app/controllers/foreman_inventory_upload) folder.

ShimShtein (changes requested):

app/controllers/api/v2/rh_cloud/inventory_controller.rb
chris1984: Is the new controller in addition to the new route I added here, or are you saying get rid of this one?

I'm guessing new one for the ui and this one stay for the api.

ShimShtein: Yup, this one will stay for the API, but for the UI we will need another one. 
If you feel that a lot of work would be duplicated, you can create a concern/class to encapsulate it, and then consume it from both API and UI

chris1984:
@ShimShtein updated, want me to start working on tests for the get/index controller methods or do you have something else in mind for the next steps?

chris1984:
Do you have a suggestion what I should add for the permissions, the tests are failing with
```
Failure:
RhCloudPermissionsTest#test_0021_route api/v2/rh_cloud/inventory/get_hosts should have a permission that grants access [/projects/foreman/test/unit/shared/access_permissions_test_base.rb:33]
Minitest::Assertion: permission for api/v2/rh_cloud/inventory/get_hosts not found, check access_permissions.rb.
Expected [] to not be empty.
``` I thought I covered them in the `engine.rb` file but I guess i'm doing it wrong?

**Comment by chris1984** [comment]
@ShimShtein Added the `POST` methods and got the permissions fixed up, let me know how this looks. I am sure it looks like 💩 🙃

File: app/controllers/foreman_inventory_upload/missing_hosts_controller.rb [hunk 3]
```diff
@@ -0,0 +1,24 @@
...
+      payload = InsightsMissingHosts.where(organization_id: organization_id)
+
+      render :json => payload
+    end
+
+    def remove_hosts
+      organization_id = params[:organization_id]
+      search_term = params[:search_term]
+
+      task = ForemanTasks.sync_task(ForemanInventoryUpload::Async::RemoveInsightsHostsJob, search_term, organization_id)
```
ShimShtein:
```suggestion
      task = ForemanTasks.async_task(ForemanInventoryUpload::Async::RemoveInsightsHostsJob, search_term, organization_id)
```
The user doesn't have to wait for the action to finish.

app/controllers/foreman_inventory_upload/missing_hosts_controller.rb
```diff
@@ -0,0 +1,24 @@
...
+
+      render json: {
+        task: task,
+      }, status: :ok
+    rescue ForemanInventoryUpload::Async::RemoveInsightsHostsJob => error
```
ShimShtein: I think the exception will be caught by the general foreman exception mechanism, so no need to handle it directly here.

app/controllers/foreman_inventory_upload/missing_hosts_controller.rb
```diff
@@ -0,0 +1,24 @@
+module ForemanInventoryUpload
+  class MissingHostsController < ::ApplicationController
+    def index
+      organizations = User.current.my_organizations
```
ShimShtein:
```suggestion
      organizations = Organization.current || User.current.my_organizations
```
If the user has scoped the UI to a specific org, I think we should respect his choice here too.

app/controllers/api/v2/rh_cloud/inventory_controller.rb
```diff
@@ -28,6 +28,31 @@ def generate_report
...
+        end
+
+        api :POST, "/organizations/:organization_id/rh_cloud/missing_hosts", N_("Grab hosts that are missing in RH Cloud")
+        param :organization_id, Integer, required: true, desc: N_("Set the current organization context for the request")
+        param :search_term, String, required: true, desc: N_("Scoped search string for host removal")
+        def remove_hosts
+          organization_id = params[:organization_id]
+          search_term = params[:search_term]
+
+          task = ForemanTasks.sync_task(ForemanInventoryUpload::Async::RemoveInsightsHostsJob, search_term, organization_id)
```
ShimShtein:
```suggestion
          task = ForemanTasks.async_task(ForemanInventoryUpload::Async::RemoveInsightsHostsJob, search_term, organization_id)
```
We don't want the thread to wait for the action to finish.

lib/foreman_rh_cloud/engine.rb
```diff
@@ -52,7 +52,7 @@ def self.register_scheduled_task(task_class, cronline)
             :generate_foreman_rh_cloud,
             'foreman_inventory_upload/reports': [:generate],
             'foreman_inventory_upload/tasks': [:create],
-            'api/v2/rh_cloud/inventory': [:sync_inventory_status, :download_file, :generate_report, :enable_cloud_connector],
+            'api/v2/rh_cloud/inventory': [:get_hosts, :remove_hosts, :sync_inventory_status, :download_file, :generate_report, :enable_cloud_connector],
```
ShimShtein, line 55: I suppose we also need to add 
``` ruby
            'foreman_inventory_upload/missing_hosts': [:remove_hosts],
```
to the "generate" role. (It should be called the admin role, as opposed to the viewer role, but I don't want to change it for backwards compatibility)

ShimShtein (changes requested):
Looks quite good actually. A couple of smaller comments and we're there.

lib/foreman_rh_cloud/engine.rb
chris1984: +1

chris1984:
> Looks quite good actually. A couple of smaller comments and we're there.

@shim Thanks 😃 also removed the rescue from the `POST` in the inventory controller as well. Marked it ready for review and removed it out of DRAFT status.

ShimShtein:
@chris1984 , could you please rebase this one?

chris1984:
@ShimShtein updated :)

chris1984:
@ShimShtein can this get a merge?

ShimShtein (approved):

PR #824, opened 2023-06-20
don't depend on theforeman/stories

evgeni:
this was dropped from core in https://github.com/theforeman/foreman/commit/5e7343ecdad7632a94efa26fcba84889e66b06b4

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,7 +9,7 @@
     "test": "tfm-test --plugin",
     "test:watch": "tfm-test --plugin --watchAll",
     "test:current": "tfm-test --plugin --watch",
-    "publish-coverage": "tfm-publish-coverage",
+    "publish-coverage": "tfm-publish-coverage"
   },
   "repository": {
     "type": "git",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -10,8 +10,6 @@
     "test:watch": "tfm-test --plugin --watchAll",
     "test:current": "tfm-test --plugin --watch",
     "publish-coverage": "tfm-publish-coverage",
-    "stories": "tfm-stories --plugin",
-    "stories:build": "tfm-build-stories --plugin",
     "stories:deploy": "surge --project .storybook-dist"
```
evgeni:
```suggestion
```

ShimShtein:
@MariaAga what am I missing here? What should be updated?

MariaAga:
looks like the error that https://github.com/theforeman/foreman-js/pull/469 solved, but not sure why it is happening here, might worth to check also if there is some custom eslint config in the project

PR #841, opened 2023-09-21
always report katello agent as not running, as it's removed

evgeni:


lib/foreman_inventory_upload/generators/slice.rb
```diff
@@ -127,7 +127,7 @@ def report_system_profile(host, host_ips_cache)
         @stream.simple_field('os_kernel_version', fact_value(host, 'uname::release'))
         @stream.simple_field('arch', host.architecture&.name)
         @stream.simple_field('subscription_status', host.subscription_status_label)
-        @stream.simple_field('katello_agent_running', host.content_facet&.katello_agent_installed?)
+        @stream.simple_field('katello_agent_running', false)
```
evgeni, line 130: the alternative is to drop this field completely, but I am not sure nothing depends on it being present

evgeni:
included in #847 for 3.9+ and not required for 3.7, closing

PR #848, opened 2023-10-17
Ensure Configure Cloud Connector runs as root

ekohl:
When a user has set the REX user as non-root, this task fails. Setting become to true means it will attempt to use sudo, which allows it to work in more cases.

Replacement of https://github.com/theforeman/foreman_rh_cloud/pull/791


chris1984 (approved):
ACK

PR #850, opened 2023-10-23
Add ability to generate report without uploading

chris1984:
@ShimShtein here is the updated PR

### DIFF
--- a/app/controllers/concerns/inventory_upload/report_actions.rb
+++ b/app/controllers/concerns/inventory_upload/report_actions.rb
@@ -10,8 +10,8 @@ module InventoryUpload
       end
     end
 
-    def start_report_generation(organization_id)
-      ForemanTasks.async_task(ForemanInventoryUpload::Async::GenerateReportJob, ForemanInventoryUpload.generated_reports_folder, organization_id)
+    def start_report_generation(organization_id, disconnected)
+      ForemanTasks.async_task(ForemanInventoryUpload::Async::GenerateReportJob, ForemanInventoryUpload.generated_reports_folder, organization_id, disconnected)
     end
 
     def report_file(organization_id)
--- a/lib/foreman_inventory_upload/async/generate_report_job.rb
+++ b/lib/foreman_inventory_upload/async/generate_report_job.rb
@@ -5,7 +5,7 @@ module ForemanInventoryUpload
         "report_for_#{label}"
       end
 
-      def plan(base_folder, organization_id)
+      def plan(base_folder, organization_id, disconnected)
         sequence do
           super(
             GenerateReportJob.output_label(organization_id),
@@ -17,7 +17,8 @@ module ForemanInventoryUpload
             QueueForUploadJob,
             base_folder,
             ForemanInventoryUpload.facts_archive_name(organization_id),
-            organization_id
+            organization_id,
+            disconnected
           )
         end
       end
--- a/app/controllers/api/v2/rh_cloud/inventory_controller.rb
+++ b/app/controllers/api/v2/rh_cloud/inventory_controller.rb
@@ -18,10 +18,12 @@ module Api
 
         api :POST, "/organizations/:organization_id/rh_cloud/report", N_("Start report generation")
         param :organization_id, Integer, required: true, desc: N_("Set the current organization context for the request")
+        param :disconnected, :bool, required: false, desc: N_('Generate the report, but do not upload')
         def generate_report
           organization_id = params[:organization_id]
+          disconnected = params[:disconnected]
 
-          start_report_generation(organization_id)
+          start_report_generation(organization_id, disconnected)
 
           render json: {
             action_status: 'success',
--- a/app/controllers/foreman_inventory_upload/reports_controller.rb
+++ b/app/controllers/foreman_inventory_upload/reports_controller.rb
@@ -21,8 +21,9 @@ module ForemanInventoryUpload
 
     def generate
       organization_id = params[:organization_id]
+      disconnected = params[:disconnected]
 
-      start_report_generation(organization_id)
+      start_report_generation(organization_id, disconnected)
 
       render json: {
         action_status: 'success',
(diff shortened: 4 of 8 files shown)


lib/foreman_inventory_upload/async/generate_report_job.rb
```diff
@@ -18,7 +18,7 @@ def plan(base_folder, organization_id)
             base_folder,
             ForemanInventoryUpload.facts_archive_name(organization_id),
             organization_id
-          )
+          ) if !disconnected
```
chris1984, line 21: Rubocop was not happy with `unless disconnected` so changed it

chris1984:
Are the tests failing because of my new param and the conditional I put on the job?

Looking at `log/test.log` I see this
```ruby
2023-10-23T16:55:07 [I|app|] InventoryUpload::Api::InventoryControllerTest: test_0001_Starts report generation
2023-10-23T16:55:07 [I|app|] ---------------------------------------------------------------------------------
2023-10-23T16:55:07 [I|app|] Processing by Api::V2::RhCloud::InventoryController#generate_report as JSON
2023-10-23T16:55:07 [I|app|]   Parameters: {"organization_id"=>"447626459", "inventory"=>{}}
2023-10-23T16:55:07 [D|tax|] Current location set to none
2023-10-23T16:55:07 [D|tax|] Current organization set to org11
2023-10-23T16:55:07 [I|app|] Completed 500 Internal Server Error in 4ms (ActiveRecord: 0.7ms | Allocations: 1881)
2023-10-23T16:55:07 [D|tax|] Current location set to none
2023-10-23T16:55:07 [D|tax|] Current organization set to none
```

app/controllers/concerns/inventory_upload/report_actions.rb
```diff
@@ -10,7 +10,7 @@ def initialize(**params)
       end
     end
 
-    def start_report_generation(organization_id)
+    def start_report_generation(organization_id, disconnected)
       ForemanTasks.async_task(ForemanInventoryUpload::Async::GenerateReportJob, ForemanInventoryUpload.generated_reports_folder, organization_id)
```
ShimShtein:
```suggestion
      ForemanTasks.async_task(ForemanInventoryUpload::Async::GenerateReportJob, ForemanInventoryUpload.generated_reports_folder, organization_id, disconnected)
```

lib/foreman_inventory_upload/async/generate_report_job.rb
```diff
@@ -18,8 +19,9 @@ def plan(base_folder, organization_id)
             base_folder,
             ForemanInventoryUpload.facts_archive_name(organization_id),
             organization_id
-          )
+          ) unless disconnected
```
ShimShtein, line 22: This means that the action will not be executed at all if `disconnected` is set to `false`

ShimShtein (changes requested):

ShimShtein (changes requested):
You have missed a reference in https://github.com/theforeman/foreman_rh_cloud/blob/foreman_3_9/lib/foreman_inventory_upload/async/generate_all_reports_job.rb#L36 and https://github.com/theforeman/foreman_rh_cloud/blob/foreman_3_9/lib/tasks/rh_cloud_inventory.rake#L50
Besides that - looks good.

chris1984:
> You have missed a reference in https://github.com/theforeman/foreman_rh_cloud/blob/foreman_3_9/lib/foreman_inventory_upload/async/generate_all_reports_job.rb#L36 and https://github.com/theforeman/foreman_rh_cloud/blob/foreman_3_9/lib/tasks/rh_cloud_inventory.rake#L50 Besides that - looks good.

Added the reference to disconnected on the `genereate_all_reports_job.rb`, the rake file already has the change you requested:

https://github.com/theforeman/foreman_rh_cloud/pull/850/files#diff-9ca8d337c7041b111f4dee13b8c59910e8ad72ca97ca82b0a3b845f68893555aR51

chris1984:
@jameerpathan111 Can you test this change with CURL? Going to work on the hammer PR while awaiting your tests

chris1984:
@jameerpathan111 

It looks like you can use hammer if you want instead of CURL, it picked up the new param with this pr checked out
```bash
[vagrant@hammer ~]$ hammer rh_report generate -h
Usage:
    hammer rh_report generate [OPTIONS]

Options:
 --disconnected BOOLEAN                  Generate the report, but do not upload
 --location[-id|-title] VALUE/NUMBER     Set the current location context for the request
 --organization[-id|-title] VALUE/NUMBER Set the current organization context for the request
 -h, --help                              Print help
```

ShimShtein:
We have missed a reference: https://github.com/theforeman/foreman_rh_cloud/blob/5745be80c865dad235c6c6057118ac8b1d26a8ab/lib/foreman_inventory_upload/async/generate_report_job.rb#L16C3-L17C31 should have the `disconnected` parameter passed too

chris1984:
> We have missed a reference: https://github.com/theforeman/foreman_rh_cloud/blob/5745be80c865dad235c6c6057118ac8b1d26a8ab/lib/foreman_inventory_upload/async/generate_report_job.rb#L16C3-L17C31 should have the `disconnected` parameter passed too

@ShimShtein updated

jameerpathan111:
@chris1984 @ShimShtein I was able to use api to generate a new report and also performed basic sanity checks on generated report. But I am not seeing `rh_report` subcommand in hammer, not sure why. I tried `foreman-rake apipie:cache` but it didn't help either.

ShimShtein:
Thanks @jameerpathan111 and @chris1984 !
Merged.

chris1984:
> @chris1984 @ShimShtein I was able to use api to generate a new report and also performed basic sanity checks on generated report. But I am not seeing `rh_report` subcommand in hammer, not sure why. I tried `foreman-rake apipie:cache` but it didn't help either.

When this goes into a snap, we can test it with hammer and see if we see it. I spun up a centos7-hammer-devel box from forklift and it saw it right away, so maybe something was up with hammer on your box

PR #866, opened 2024-01-18
use shared github workflow to run tests

evgeni:


evgeni:
cold cache: 13m 24s (https://github.com/theforeman/foreman_rh_cloud/actions/runs/7567480704/job/20606728804)
warn cache:  11m 18s (https://github.com/theforeman/foreman_rh_cloud/actions/runs/7567634004/job/20607196990)
old action: 7m 50s (https://github.com/theforeman/foreman_rh_cloud/actions/runs/7531259398/job/20499435028)

so, while slower, certainly not *bad*

ekohl:
Should this also drop the "build container" workflows, since they're no longer needed?

ekohl:
This fails in a migration, probably because the plugin can't be installed after Foreman has already been installed. I suppose that's another difference in migrations. Perhaps our reusable action should test both approaches.

evgeni:
> This fails in a migration, probably because the plugin can't be installed after Foreman has already been installed. I suppose that's another difference in migrations. Perhaps our reusable action should test both approaches.

Correct.
* I've opened https://projects.theforeman.org/issues/37076 to track this in Katello
* I've opened https://github.com/theforeman/actions/pull/32 to allow us to skip this while :point_up: is not fixed
* The reusable action does test `db:seed` in the "normal" flow (install foreman and plugin in the same run) since https://github.com/theforeman/actions/commit/590eefb499d35a190f851b1cb44459bb7f94abbd, so no action needed

evgeni:
I've split out some of the changes here into #867 and #868 as they are only indirectly related to the whole GHA change

foreman_rh_cloud.gemspec
```diff
@@ -18,12 +18,12 @@ Gem::Specification.new do |s|
   s.test_files = Dir['test/**/*']
   s.test_files -= Dir['test/**/*.orig']
 
+  s.required_ruby_version = '>= 2.7', '< 4'
```
evgeni: @ekohl interestingly it does not fail

.github/workflows/ruby_tests.yml
```diff
@@ -2,63 +2,23 @@ name: Ruby tests
...
-          /usr/bin/entrypoint.sh pnpm config set store-dir /root/.local/share/pnpm/store/v3
-          cd /projects/foreman
-          /usr/bin/entrypoint.sh git config --global user.name $GIT_COMMITTER_NAME
-          /usr/bin/entrypoint.sh git config --global user.email $GIT_COMMITTER_EMAIL
-      - name: Run tests suite
-        run: |
-          cd /projects/foreman
-          /usr/bin/entrypoint.sh /usr/bin/run_tests.sh
+  test:
+    name: Ruby
```
evgeni, line 19: now that rubocop is passing, this should get a `needs: rubocop`

.github/workflows/ruby_tests.yml
```diff
@@ -2,63 +2,23 @@ name: Ruby tests
...
+    uses: theforeman/actions/.github/workflows/foreman_plugin.yml@existing-database
```
evgeni: https://github.com/theforeman/actions/pull/32

foreman_rh_cloud.gemspec
ekohl: That is quite weird. I saw it fail before. I swear I'm not crazy ;)

evgeni: So did I 🤷‍♀️

ekohl: Perhaps the use of theforeman-rubocop actually downgraded it to a version that doesn't have the cop

ekohl (approved):

ShimShtein:
@evgeni , can you define the version of Foreman you want to use for the tests? I see the action uses the develop branch by default. It won't be good enough for stable branches.

evgeni:
Of course, which ones would you like? 3.9 and develop?

evgeni:
> Of course, which ones would you like? 3.9 and develop?

I've implemented 3.9 an develop now, what do you think? :)

ekohl:
@ShimShtein could you have a look?

ShimShtein (approved):

ShimShtein:
Merged, thanks @evgeni !

Do you expect we will need to cherry-pick this one to older branches of the plugin?

evgeni:
Nah, as long as it's in the one for 3.10, we're good

ShimShtein:
Which version of Katello are we using? I will probably need to freeze Katello and REX versions. Is there an option to do that, or do I need to freeze them in the gemspec?

evgeni:
Whichever the gemspec allows is used, yeah.

PR #867, opened 2024-01-22
define required_ruby_version

evgeni:


foreman_rh_cloud.gemspec
```diff
@@ -18,6 +18,8 @@ Gem::Specification.new do |s|
   s.test_files = Dir['test/**/*']
   s.test_files -= Dir['test/**/*.orig']
 
+  s.required_ruby_version = '>= 2.7'
```
ekohl: I like an upper bound
```suggestion
  s.required_ruby_version = '>= 2.7', '< 4'
```

ekohl (changes requested):

ekohl:
Also, the target Ruby version needs to match. Looks like the config is pinned to Foreman 2.3 even though this branch is for Foreman 3.9.
https://github.com/theforeman/foreman_rh_cloud/blob/ebcbeed9d6707fa2760c14490ee8824e25ef2d7b/.rubocop.yml

evgeni:
> Also, the target Ruby version needs to match. Looks like the config is pinned to Foreman 2.3 even though this branch is for Foreman 3.9. https://github.com/theforeman/foreman_rh_cloud/blob/ebcbeed9d6707fa2760c14490ee8824e25ef2d7b/.rubocop.yml

I wanted to tackle that once the first round of auto-fixes is in ;-)

ekohl:
I mostly mention it because https://github.com/theforeman/foreman_rh_cloud/pull/866 fails on it.

evgeni:
hmm, given it takes the TargetRubyVersion from theforeman-rubocop… shall we set it to 2.5+ for now, or add it to rubocop_todo to fix properly?

ekohl (approved):

chris1984 (approved):

PR #868, opened 2024-01-22
rubocop --auto-correct

evgeni:


evgeni:
no idea why the current CI didn't flag those rubocop offenses, but #866 did, so here we are.

config/routes.rb
```diff
@@ -41,7 +41,7 @@
     end
 
     scope '/r/insights' do
-      match '(/*path)(/)', :constraints => lambda { |req| !req.path.include?('view/api') }, to: 'machine_telemetries#forward_request', via: [:get, :post, :delete,:put, :patch]
+      match '(/*path)(/)', :constraints => ->(req) { !req.path.include?('view/api') }, :to => 'machine_telemetries#forward_request', :via => [:get, :post, :delete, :put, :patch]
```
ekohl: Wouldn't it be better to make it consistent this way:
```suggestion
      match '(/*path)(/)', constraints: ->(req) { !req.path.include?('view/api') },  to: 'machine_telemetries#forward_request', via: [:get, :post, :delete, :put, :patch]
```

lib/tasks/rh_cloud_inventory.rake
```diff
@@ -55,8 +55,8 @@ namespace :rh_cloud_inventory do
 
   desc "Synchronize Hosts inventory"
   task sync: [:environment, 'dynflow:client'] do
-    if ! ENV['organization_id'].nil?
-      organizations = [ Organization.where(:id => ENV['organization_id']).first ]
+    if !ENV['organization_id'].nil?
```
ekohl, line 58: Funny how the `generate_upload` is close, but not really the same in various subtle ways.

ekohl (changes requested):

config/routes.rb
evgeni: consistent to what? the rest of the file uses both styles, so it's hard to say ;-)

(I don't mind either way, this is just what the :policeman: generated)

ekohl: I think my suggestion is a smaller diff and also more consistent within the file.

ekohl (approved):

lib/tasks/rh_cloud_inventory.rake
ShimShtein: We definitely can refactor both of them into using a helper method that will receive the action as lambda. Something along the lines of
``` ruby
with_organization_scope do |organization|
  ForemanTasks.async_task(InventorySync::Async::InventoryFullSync, organization)
end
```

Indeed it would be much prettier.

PR #873, opened 2024-03-21
Make sure to forward raw_post for patch commands

ShimShtein:


chris1984 (approved):
ACK, works well:

Tested register/unregister to make sure there were not any regressions
```bash
[root@dell-r730-018 ~]# insights-client --ansible-host=foo.example.com
Ansible hostname updated to foo.example.com.
```

Issue #874, opened 2024-03-22
Scheduled task for uploading report fails


knoppi:
System:
  - foreman 3.9.1
  - foreman_rh_cloud 9.0.53

The regular upload of the insights reports fails with the exception ArgumentError: wrong number of arguments (given 2, expected 3).

![grafik](https://github.com/theforeman/foreman_rh_cloud/assets/10814054/37522dc6-009e-4bf8-983d-4ab085783b37)

I think this was introduced to #850. The method `ForemanInventoryUpload:Async:GenerateAllReportsJob:plan_generate_report` obtained a new argument `disconnected` but the call in line 23 in `lib/foreman_inventory_upload/async/generate_all_reports_job.rb` still only passes 2 arguments.

chris1984:
Fixed in https://github.com/theforeman/foreman_rh_cloud/pull/875

chris1984:
Will get a new version of 9.x out with the backported fix

PR #875, opened 2024-03-22
Fix scheduled task for uploading report failed (#874)

sbernhard:


chris1984 (approved):
@sbernhard thanks for catching this from the discourse post. ACK

PR #877, opened 2024-03-22
[CP] Cherrypick for Foreman3_9 and version bump to 9.0.56

chris1984:


chris1984:
@ShimShtein I am getting this:
```ruby
Foreman::PluginRequirementError: ERF72-1682 [Foreman::PluginRequirementError]: foreman_remote_execution plugin requires Foreman >= 3.10 but current is 3.9.1
```

Am I correct in that I need to make a commit before the version bump to pin the version of `foreman_remote_execution` that uses Foreman 3.9?

PR #878, opened 2024-03-28
Pin remote_execution, katello and foreman_ansible to Foreman 3.9 versions

chris1984:


.github/workflows/ruby_tests.yml
```diff
@@ -29,3 +29,4 @@ jobs:
       postgresql_container: ghcr.io/theforeman/postgresql-evr
       test_existing_database: false
       foreman_version: ${{ matrix.foreman }}
+      matrix_include: '[{"ruby": "2.7", "node": "16"}]'
```
ekohl: Why include this? We don't run this combination anywhere.

gemfile.d/lock.rb
```diff
@@ -0,0 +1 @@
+gem "foreman_remote_execution", "<= 12.0.5"
```
ekohl: What's wrong with 12.0.6? Wouldn't semver suggest something like `< 13` would be more appropriate?

.github/workflows/ruby_tests.yml
chris1984: I'll remove this, this was me trying to figure out the JavaScript stuff until we found out it's in foreman.

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "foreman_rh_cloud",
-  "version": "9.0.51",
+  "version": "9.0.55",
```
ekohl: This looks unrelated.

gemfile.d/lock.rb
```diff
@@ -0,0 +1,2 @@
+gem "foreman_remote_execution", "< 13"
+gem "foreman_ansible", "< 14"
```
ekohl: For CI you can also provide an environment variable like https://github.com/dm-drogeriemarkt/foreman_wreckingball/blob/861f54cab044dafddefffe9e0402b9aeb330ba80/.github/workflows/ci.yml#L34-L35

Then you can use `ENV['FOREMAN_VERSION']` in this file. That can be useful to be more future compatible: when you start to support newer versions then you don't want to forget about this.

I'm using https://github.com/dm-drogeriemarkt/foreman_wreckingball/pull/116 to verify this theory. If so, I'll properly document it.

chris1984:
@ekohl is there anything extra I need to do to get the JS passing? I see the prs are merged in Foreman and Foreman-JS. Katello prs are passing now too.

chris1984:
@ekohl Do you know how I can pin Katello 4.11 ?

gemfile.d/lock.rb
```diff
@@ -0,0 +1,6 @@
+case ENV['FOREMAN_VERSION']
+when '3.9-stable'
+  gem "foreman_remote_execution", "< 13"
+  gem "foreman_ansible", "< 14"
+  gem "katello", "< 12"
```
ekohl: I think you want
```suggestion
  gem "katello", "~> 4.11.0"
```

gemfile.d/lock.rb
```diff
@@ -0,0 +1,6 @@
+case ENV['FOREMAN_VERSION']
+when '3.9-stable'
+  gem "foreman_remote_execution", "< 13"
```
ekohl, line 3: Once https://github.com/Katello/katello/commit/4a91ec50461a5bbdac6929f7f4cb42eddb3a7afb is released this shouldn't be needed anymore, but for now this is ok.

ekohl (changes requested):
Last thing is the Katello version.

ekohl (approved):

PR #885, opened 2024-05-28
Add safe navigation if ip is nil on hosts sync

chris1984:
### What are the changes introduced in this pull request?
* Adding a skip if `ip_addresses` is `nil`

### What are the testing steps for this pull request?
* Register a host, run inventory sync
* Unregister the host to run the missing hosts method
* Add in `ip_addresses = nil` at https://github.com/theforeman/foreman_rh_cloud/blob/v10/lib/inventory_sync/async/inventory_hosts_sync.rb#L70
* Run `foreman-rake rh_cloud_inventory:sync` and watch the `production.log` fail with `undefined method take_while for nil:NilClass (NoMethodError)`
* Apply PR and restart services
* Try the task and see that it completes without a traceback

### DIFF
--- a/lib/inventory_sync/async/inventory_hosts_sync.rb
+++ b/lib/inventory_sync/async/inventory_hosts_sync.rb
@@ -68,7 +68,7 @@ module InventorySync
 
       def to_ip_address_string(ip_addresses)
         string_size = 0
-        ip_addresses.take_while { |address| (string_size += address.length) <= MAX_IP_STRING_SIZE }
+        ip_addresses&.take_while { |address| (string_size += address.length) <= MAX_IP_STRING_SIZE }
       end
 
       def plan_self_host_sync
--- a/test/jobs/inventory_hosts_sync_test.rb
+++ b/test/jobs/inventory_hosts_sync_test.rb
@@ -44,11 +44,59 @@ class InventoryHostsSyncTest < ActiveSupport::TestCase
 
     inventory_json = <<-INVENTORY_JSON
     {
-      "total": 3,
-      "count": 3,
+      "total": 4,
+      "count": 4,
       "page": 1,
-      "per_page": 3,
+      "per_page": 4,
       "results": [
+        {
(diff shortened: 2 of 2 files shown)


ShimShtein (changes requested):
I know it's a small issue, but can you please add a test case to https://github.com/theforeman/foreman_rh_cloud/blob/c74cfb41a6f3a398f04dc5bb049b17d5da0e95d9/test/jobs/inventory_hosts_sync_test.rb ?

Specifically we can add a mock record similar to the one we have on [line 58](https://github.com/theforeman/foreman_rh_cloud/blob/c74cfb41a6f3a398f04dc5bb049b17d5da0e95d9/test/jobs/inventory_hosts_sync_test.rb#L58), but without the `ip_addresses` field and modify a the [create records test](https://github.com/theforeman/foreman_rh_cloud/blob/c74cfb41a6f3a398f04dc5bb049b17d5da0e95d9/test/jobs/inventory_hosts_sync_test.rb#L298-L311) to make sure it creates a missing record for the new "missing" host too.

chris1984:
@ShimShtein let me know what i'm missing with the test, it's still asserting 2 instead of 3. or we can talk about it tomorrow when we sync up.

ShimShtein:
@chris1984 you are not missing anything, actually you are adding too much here: https://github.com/theforeman/foreman_rh_cloud/pull/885/files#diff-1db2fc1b0779ffd29ad0788780a64b71490bad277d33887ef72e1283465b09d5R42

If you create a record in foreman, and create it in the Insights request mock, it means that it's not a missing host. This is why you have only two missing hosts, even though you have added a new one to the Insights request.
If you want to test for creation of a new missing host record, you need to make sure it exists only in Insights, and not in Foreman.

chris1984:
@ShimShtein updated, thanks for the pointer

ShimShtein:
Merged, thanks @chris1984 !

PR #886, opened 2024-05-31
Update RuboCop configuration to inherit from theforeman-rubocop

archanaserver:
- Added theforeman-rubocop as a development dependency in the gemspec.
- Changed the RuboCop configuration to inherit from default.yml provided
by theforeman-rubocop for better style alignment.
- Removed redundant cops as they are already included in the inherited
default.yml or not needed anymore.

default.yml: https://github.com/theforeman/theforeman-rubocop/?tab=readme-ov-file#basic-style---default-performance-and-rails-cops

This is part of Rubocop standerdization, link for the reference: https://community.theforeman.org/t/standardizing-rubocop-with-theforeman-rubocop/37239


.rubocop.yml
```diff
@@ -1,2 +1,78 @@
-inherit_from:
-  - https://raw.githubusercontent.com/theforeman/foreman/2.3-stable/.rubocop.yml
+inherit_from: .rubocop_todo.yml
+
+# inherit_from: .rubocop_todo.yml
```
qcjames53: This line can be safely removed.

.rubocop.yml
qcjames53, line 1: I noticed there are a few styles included in the [old rubocop config file](https://raw.githubusercontent.com/theforeman/foreman/2.3-stable/.rubocop.yml) that didn't make the jump over to this file (Style/CollectionMethods, Style/Documentation, Style/EmptyAssignment, etc). Is that intentional? The two new inclusions at the bottom (RaiseArgs and SkipsModelValidations) make sense to me.

I know this hardly matters but I think it would be great if this file were alphabetized for readability like the previous version. Would that be ok?

db/migrate/20210720000001_remove_old_insights_statuses.foreman_rh_cloud.rb
```diff
@@ -1,6 +1,6 @@
 class RemoveOldInsightsStatuses < ActiveRecord::Migration[5.2]
   def up
-    InsightsClientReportStatus.where(status: 2).update_all(status: InsightsClientReportStatus::NO_REPORT)
-    InsightsClientReportStatus.where(status: 3).update_all(status: InsightsClientReportStatus::REPORTING)
+    InsightsClientReportStatus.where(status: 2).update(status: InsightsClientReportStatus::NO_REPORT)
```
qcjames53: This might be embarrassing if I'm incorrect here but I don't believe we can call `update` on an ActiveRecord::Relation as is done here. I asked ChatGPT for backup and it agreed with me :)

It seems like this will need to be a `.each do |record| record.update(...) end` kind of thing if we're looking to have validations run? Please let me know if I'm wrong here.

db/migrate/20221102110254_fix_rh_cloud_settings_category_to_dsl.rb
```diff
@@ -2,6 +2,6 @@
 
 class FixRhCloudSettingsCategoryToDsl < ActiveRecord::Migration[6.0]
   def up
-    Setting.where(category: 'Setting::RhCloud').update_all(category: 'Setting') if column_exists?(:settings, :category)
+    Setting.where(category: 'Setting::RhCloud').update(category: 'Setting') if column_exists?(:settings, :category)
```
qcjames53: Same comment as above with the `update` thing.

lib/insights_cloud/async/insights_client_status_aging.rb
```diff
@@ -7,7 +7,7 @@ def run
         host_ids = InsightsClientReportStatus.stale.reporting.pluck(:host_id)
 
         # update all stale records
-        InsightsClientReportStatus.where(host_id: host_ids).update_all(status: InsightsClientReportStatus::NO_REPORT)
+        InsightsClientReportStatus.where(host_id: host_ids).update(status: InsightsClientReportStatus::NO_REPORT)
```
qcjames53: Same as above.

lib/insights_cloud/async/insights_full_sync.rb
```diff
@@ -100,7 +100,7 @@ def host_id(uuid)
       def replace_hits_data(hits, organization)
         InsightsHit.transaction do
           # Reset hit counters to 0, they will be recreated later
-          InsightsFacet.for_organizations(organization.id).update_all(hits_count: 0)
+          InsightsFacet.for_organizations(organization.id).update(hits_count: 0)
```
qcjames53: Same as above.

test/unit/foreman_rh_cloud_self_host_test.rb
```diff
@@ -18,7 +18,7 @@ class ForemanRhCloudSelfHostTest < ActiveSupport::TestCase
 
   test 'finds host by shortname' do
     @host = FactoryBot.create(:host, :managed)
-    Host.where(name: @host.name).update_all(name: @host.shortname)
+    Host.where(name: @host.name).update(name: @host.shortname)
```
qcjames53: Same as above.

qcjames53 (changes requested):
Hi Archana!

I'm sorry it took me so long to review this PR, I wanted to triple check everything in the config file since this will end up being pretty important for devs and QE. I had a few questions and suggestions there but all in all everything looks great.

I expanded on this below but I don't *think* we can call `.update` on an `ActiveRecord::Relation`. It's not in any documentation I could find. I asked ChatGPT if the changes were ok and it said NO 3 times and YES once 🤷‍♂️. I'd put everything in a `.each` loop if validations are important and would keep the lines as `.update_all` if not.

archanaserver:
> I expanded on this below but I don't _think_ we can call `.update` on an `ActiveRecord::Relation`. 

Yes you are right, I had similar conversation few weeks back in one of my another PR on update and update_all, I misunderstood few things here in the code, but I was following the RuboCop doc: https://docs.rubocop.org/rubocop-rails/cops_rails.html#railsskipsmodelvalidations So I did this to few other places as well that is disabling this cop because it is not safe to use.

.rubocop.yml
archanaserver: > I noticed there are a few styles included in the [old rubocop config file](https://raw.githubusercontent.com/theforeman/foreman/2.3-stable/.rubocop.yml) that didn't make the jump over to this file (Style/CollectionMethods, Style/Documentation, Style/EmptyAssignment, etc). Is that intentional? The two new inclusions at the bottom (RaiseArgs and SkipsModelValidations) make sense to me.

What do you mean by this? Would you mind expanding a little more.


> I know this hardly matters but I think it would be great if this file were alphabetized for readability like the previous version. Would that be ok?

I don’t think we have specifically addressed alphabetizing RuboCop rules in our changes so far, but I can look into whether there is a cop for this and update it accordingly. Thanks for bringing this up; I believe it would be a good improvement for readability

qcjames53: Hi Archana! For the old config file stuff, I looked through the file linked above and found these styles that used to be in the rubocop definition file but are now no longer included:
```
Style/Alias:
Style/AndOr:
Style/CollectionMethods:
Style/Documentation:
Style/EmptyMethod:
Style/FrozenStringLiteralComment:
Style/InverseMethods:
Style/MultipleComparison:
Style/ParallelAssignment:
Style/PreferredHashMethods:
Style/TernaryParentheses:
Style/TrailingCommaInArrayLiteral:
Style/TrailingCommaInHashLiteral:
```

We import a good amount of these styles through default.yml in theforeman-rubocop. Here is what's left:
```
Style/CollectionMethods:
Style/FrozenStringLiteralComment:
Style/InverseMethods:
Style/MultipleComparison:
Style/ParallelAssignment:
Style/PreferredHashMethods:
```

I didn't find these styles included anywhere I looked; is it OK to drop these requirements?

qcjames53 (commented):
The changes look great to me, thank you! I'm totally ok with keeping `update_all` instead of refactoring into an each loop; I can see several ways that could end up a big headache. I just had one question about the included styles (posted elsewhere) but I'm happy to ACK once that is resolved.

.rubocop.yml
archanaserver: > For the old config file stuff, I looked through the file linked above and found these styles that used to 
> be in the rubocop definition file but are now no longer included:
> 
> ```
> Style/Alias:
> Style/AndOr:
> Style/CollectionMethods:
> Style/Documentation:
> Style/EmptyMethod:
> Style/FrozenStringLiteralComment:
> Style/InverseMethods:
> Style/MultipleComparison:
> Style/ParallelAssignment:
> Style/PreferredHashMethods:
> Style/TernaryParentheses:
> Style/TrailingCommaInArrayLiteral:
> Style/TrailingCommaInHashLiteral:
> ```
@qcjames53 so style cops which are not included in the config because we already have those in [theforeman-rubocop](https://github.com/theforeman/theforeman-rubocop), and we are inheriting this in rubocop config, so I believe we don't need to mention it again.

> Style/CollectionMethods:
> Style/FrozenStringLiteralComment:
> Style/InverseMethods:
> Style/MultipleComparison:
> Style/ParallelAssignment:
> Style/PreferredHashMethods:
> 
> I didn't find these styles included anywhere I looked; is it OK to drop these requirements?

Let's talk about each cops I have dropped here:

`Style/CollectionMethods`: I have dropped this one because I found that we don't need it, even after enabling I found no offenses, please correct me if I'm wrong or missing something. If adding this is good, I can add it here but looking at the old config, it was mentioned to disable it and that make no sense to add this in the config.
Same for `Style/InverseMethods`, `Style/MultipleComparison`, `Style/ParallelAssignment` and `Style/PreferredHashMethods`

`Style/FrozenStringLiteralComment`: We have this already, but enabled it to false, as mentioned in the old config.

archanaserver (commented):
Dropping those cops from here because we already have it in theforeman-rubocop.

.rubocop.yml
qcjames53: Thank you for the great explanation with this! That all sounds good to me. ACK'ing

qcjames53 (approved):

chris1984 (approved):
Thanks @archanaserver for your contribution and @qcjames53 for reviewing this, looks good from my end too.

archanaserver:
Thanks @chris1984 and @qcjames53 :)

PR #888, opened 2024-06-13
SAT-19621  Scope Insights clean task as anonymous_admin

chris1984:
@ShimShtein here is the PR. I was unable to reproduce the issue locally but still made it based on our talks.


ShimShtein:
Merged, thanks @chris1984 !

PR #889, opened 2024-06-17
SAT-19796 - Change Google to gcp to match RH Cloud expectations

chris1984:
@ShimShtein here is the Google one


ShimShtein:
Merged. Thanks @chris1984!

PR #890, opened 2024-06-20
SAT-22626 - Log rule response on insights sync and dont fail

chris1984:


ShimShtein:
Merged, thanks @chris1984 !

PR #897, opened 2024-08-28
Replace "Satellite" with "Foreman" in user-facing strings.

Thorben-D:
This PR provides a starting point for un-branding this plugin.

Only user-facing strings are un-branded right now. Un-branding environment variables or hash-keys would either break existing usages or create a bit of a mess. I am not sure that would be worth the trouble...

Using `foreman_theme_satellite` in conjunction with this plugin re-applies the branding.

### DIFF
--- a/webpack/ForemanInventoryUpload/Components/PageHeader/components/PageDescription/PageDescription.js
+++ b/webpack/ForemanInventoryUpload/Components/PageHeader/components/PageDescription/PageDescription.js
@@ -8,19 +8,19 @@ export const PageDescription = () => (
   <div id="inventory_page_description">
     <Text>
       {__(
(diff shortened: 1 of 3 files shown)


chris1984 (commented):
Looks fine, can you fix the failing snapshot JS test?

Thorben-D:
Tests should be working now!

ShimShtein:
Let's wait for the branching, and then push the PR. We are after the strings were sent to translations, so changing the templates will result in breakage.

sbernhard:
Is the branching already done @ShimShtein ? 
maybe have a look at https://github.com/theforeman/foreman_rh_cloud/pull/901, too

chris1984:
> Is the branching already done @ShimShtein ? maybe have a look at #901, too

Branching will happen after we go GA downstream. There is discussion still about the branch name for the next version too.

Thorben-D:
Rebased to "develop"

chris1984 (approved):
Looks good, Thanks @Thorben-D

PR #901, opened 2024-09-17
Use theforeman upstream cloud_connector role

sbernhard:


app/views/job_templates/cloud_connector.erb
```diff
@@ -34,4 +34,4 @@ feature: ansible_configure_cloud_connector
   vars:
     satellite_cloud_connector_url: "<%= foreman_server_url %>"
   roles:
-    - redhat.satellite_operations.cloud_connector
+  - theforeman.operations.cloud_connector
```
maximiliankolb:
```suggestion
    - theforeman.operations.cloud_connector
```

maximiliankolb (approved):
I was expecting this to show up on Ansible Galaxy: https://docs.ansible.com/ansible/latest/collections/theforeman/index.html Unsure why because I found https://github.com/theforeman/foreman-operations-collection/blob/develop/galaxy.yml#L2

Diff LGTM from a branding perspective.

chris1984:
> I was expecting this to show up on Ansible Galaxy: https://docs.ansible.com/ansible/latest/collections/theforeman/index.html Unsure why because I found https://github.com/theforeman/foreman-operations-collection/blob/develop/galaxy.yml#L2
> 
> Diff LGTM from a branding perspective.

Downstream I guess we will have to branch this instead of pulling the gem in and changing this back to Satellite Operations. Does that sound correct @ShimShtein If so that could cause some overhead for the delivery team.

app/views/job_templates/cloud_connector.erb
```diff
@@ -34,4 +34,4 @@ feature: ansible_configure_cloud_connector
   vars:
     satellite_cloud_connector_url: "<%= foreman_server_url %>"
   roles:
-    - redhat.satellite_operations.cloud_connector
+    - theforeman.operations.cloud_connector
```
ehelms, line 40: This alone won't work either -- all of the variables in this template are expecting the `redhat.satellite_operations` variables. I think the path forward is to create two templates and not try to play a game of branding.

sbernhard: Understand. Thanks for your feedback. 
I have not test system right now for this change but, what do you think about changing the vars in that way, that they don't have foreman/satellite in their name. The template can then set foreman_cloud_connector_* or satellite_cloud_connector_* accordingly.

ShimShtein: I think we can start with renaming the vars to something that is project-neutral, but Eric's idea of replacing the template looks like a way forward if we want to properly brand this.

@adamruzicka , do we have a how-to guide for branding `for_feature` calls?

adamruzicka: What would you like to brand there?

PR #904, opened 2024-09-23
correctly register REX features when booting the app

evgeni:


evgeni:
In https://github.com/theforeman/foreman_rh_cloud/commit/3e9932693ed6cfd232e486596662414c7c70bbc0 the plugin registration part was moved into `app.reloader.to_prepare`, but not the REX features part.
Given the same (registering REX features) works for Katello, and there it was moved into `reloader.to_prepare` in https://github.com/Katello/katello/commit/c90a1dc46dcf0991127f1f369f83d00e07b524d5, I did the same here.

The `:after => 'foreman_remote_execution.register_plugin'` part wasn't necessary in my tests, but seemed more correct to explicitly specify.

lib/foreman_rh_cloud/engine.rb
```diff
@@ -157,37 +157,43 @@ def self.register_scheduled_task(task_class, cronline)
...
-        # skip object creation when admin user is not present, for example in test DB
-        if User.unscoped.find_by_login(User::ANONYMOUS_ADMIN).present?
-          ::ForemanTasks.dynflow.config.on_init(false) do |world|
-            ForemanRhCloud::Engine.register_scheduled_task(ForemanInventoryUpload::Async::GenerateAllReportsJob, '0 0 * * *')
-            ForemanRhCloud::Engine.register_scheduled_task(InventorySync::Async::InventoryScheduledSync, '0 0 * * *')
-            ForemanRhCloud::Engine.register_scheduled_task(InsightsCloud::Async::InsightsScheduledSync, '0 0 * * *')
-            ForemanRhCloud::Engine.register_scheduled_task(InsightsCloud::Async::InsightsClientStatusAging, '0 0 * * *')
-          end
+      # skip object creation when admin user is not present, for example in test DB
+      if ActiveRecord::Base.connection.data_source_exists?(ForemanTasks::Task.table_name) && User.unscoped.find_by_login(User::ANONYMOUS_ADMIN).present?
```
evgeni, line 157: rubo :police_officer:  made me do this :(

lib/foreman_rh_cloud/engine.rb
```diff
@@ -157,37 +157,43 @@ def self.register_scheduled_task(task_class, cronline)
       end
     end
 
-    initializer 'foreman_rh_cloud.register_rex_features', :before => :finisher_hook do |_app|
+    initializer 'foreman_rh_cloud.register_rex_features', :before => :finisher_hook, :after => 'foreman_remote_execution.register_plugin' do |app|
+      app.reloader.to_prepare do
```
evgeni: The alternative seems to be to do this in a `config.to_prepare` block, and not in an initializer.

I have no idea what the differences are…

ShimShtein:
This looks much better, but still I can't understand why is it necessary. Also I am not sure what would be the impack on reloading in dev environment (`to_prepare` runs each time Rails reloads the app).

evgeni:
Impact in terms of performance? `.register` essentially does an `upsert` (https://github.com/theforeman/foreman_remote_execution/blob/d4abcf575b61211c3d0d970479431ff7bdbf4c3f/app/models/remote_execution_feature.rb#L30-L72) so as long as it's not reloading multiple times a second, it should be fine?

lib/foreman_rh_cloud/engine.rb
```diff
@@ -157,37 +157,37 @@ def self.register_scheduled_task(task_class, cronline)
       end
     end
 
-    initializer 'foreman_rh_cloud.register_rex_features', :before => :finisher_hook do |_app|
+    config.to_prepare do
```
ShimShtein, line 168: Small nitpick: can we have all initializers together and then the `to_prepare` block?

evgeni: Sure

ShimShtein:
Merged, thanks @evgeni!

evgeni:
I have verified this fixes Luna pipes by running:
```
ansible-playbook pipelines/install_pipeline.yml -e pipeline_os=centos9-stream -e pipeline_type=luna -e pipeline_version=nightly -e '{"packit_prs":["https://github.com/theforeman/foreman_rh_cloud/pull/904"]}'
```

PR #908, opened 2024-09-25
allow babel/core 7.x

evgeni:
We currently use 7.23 in Foreman and this works fine. Reflect it in package.json


chris1984 (approved):
Ship it

PR #911, opened 2024-10-02
Fixes #910 - Insights tab displayed for non-RHEL hosts

### LINKED ISSUE #910: The "Insights" tab is displayed for all operating systems
After installing the RH Cloud plugin, the Insights tab is displayed for each operating system on the host details page, although this is a RHEL feature.

image

Steps to Reproduce:
1. install the RH Cloud plugin
2. check the host details page under "Hosts-->All Hosts-->hostname"

Actual results:
The "Insights" tab is displayed for all operating systems

Expected results:
The "Insights" tab is displayed only for RHEL operating systems

The same applies to the total risk card on the overview tab.

Thorben-D:
Fixes #910 

## Changes made
- Add helper-function isNotRhelHost
- Hide insights tab based on host-OS by using "hideTab" fill-metadata field.
- Hide TotalRisk card based on host-OS by wrapping it.

The isNotRhelHost check is not the nicest but this is similar to how it is done in:
provisioning_templates/snippet/insights.erb:<% if @host.operatingsystem.name == 'RedHat' -%>

### DIFF
--- a/webpack/ForemanRhCloudHelpers.js
+++ b/webpack/ForemanRhCloudHelpers.js
@@ -4,3 +4,7 @@
  * should be imported once core moves it to the ReactApp folder.
  */
 export const foremanUrl = path => `${window.URL_PREFIX}${path}`;
+
+export const isNotRhelHost = ({ hostDetails }) =>
+  // eslint-disable-next-line camelcase
+  !new RegExp('red\\s?hat', 'i').test(hostDetails?.operatingsystem_name);
--- a/webpack/InsightsHostDetailsTab/InsightsTotalRiskChartWrapper.js
+++ b/webpack/InsightsHostDetailsTab/InsightsTotalRiskChartWrapper.js
@@ -1,0 +1,23 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import InsightsTotalRiskChart from './InsightsTotalRiskChart';
+import { isNotRhelHost } from '../ForemanRhCloudHelpers';
+
+export const InsightsTotalRiskChartWrapper = props => {
+  if (props.status === 'RESOLVED') {
+    return (
+      !isNotRhelHost(props.hostDetails) && <InsightsTotalRiskChart {...props} /> // check for RHEL hosts
+    );
+  }
+  return null;
+};
+
+InsightsTotalRiskChartWrapper.propTypes = {
+  status: PropTypes.string,
+  hostDetails: PropTypes.object,
+};
+
+InsightsTotalRiskChartWrapper.defaultProps = {
+  status: 'PENDING',
+  hostDetails: {},
+};
--- a/webpack/ForemanRhCloudFills.js
+++ b/webpack/ForemanRhCloudFills.js
@@ -2,7 +2,8 @@ import React from 'react';
 import { addGlobalFill } from 'foremanReact/components/common/Fill/GlobalFill';
 import InventoryAutoUploadSwitcher from './ForemanInventoryUpload/SubscriptionsPageExtension/InventoryAutoUpload';
 import NewHostDetailsTab from './InsightsHostDetailsTab/NewHostDetailsTab';
-import InsightsTotalRiskCard from './InsightsHostDetailsTab/InsightsTotalRiskChart';
+import { InsightsTotalRiskChartWrapper } from './InsightsHostDetailsTab/InsightsTotalRiskChartWrapper';
+import { isNotRhelHost } from './ForemanRhCloudHelpers';
 
 const fills = [
   {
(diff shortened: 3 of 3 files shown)


webpack/ForemanRhCloudHelpers.js
```diff
@@ -4,3 +4,7 @@
  * should be imported once core moves it to the ReactApp folder.
  */
 export const foremanUrl = path => `${window.URL_PREFIX}${path}`;
+
+export const isNotRhelHost = ({ hostDetails }) =>
+  // eslint-disable-next-line camelcase
+  !new RegExp('red\\s?hat', 'i').test(hostDetails?.operatingsystem_name);
```
nadjaheitmann, line 10: `operatingsystem_name` is not safe as this value can be changed by the user. As this plugin depends on Katello anyways, you can use `hostDetails?.operatingsystem_family`

webpack/ForemanRhCloudHelpers.js
```diff
@@ -4,3 +4,7 @@
  * should be imported once core moves it to the ReactApp folder.
  */
 export const foremanUrl = path => `${window.URL_PREFIX}${path}`;
+
+export const isNotRhelHost = ({ hostDetails }) =>
```
bastian-src, line 8: Might be nitpicking, but since the const is only called as `!isNotRhelHost`, I'd recommend changing it to `isRhelHost`.

webpack/ForemanRhCloudHelpers.js
nadjaheitmann: Well, you will get the OS type Redhat for Alma, Rocky, etc. as well, so this is tricky... But maybe you can use it for a sanity check.

webpack/InsightsHostDetailsTab/InsightsTotalRiskChartWrapper.js
```diff
@@ -0,0 +1,23 @@
...
+import InsightsTotalRiskChart from './InsightsTotalRiskChart';
+import { isNotRhelHost } from '../ForemanRhCloudHelpers';
+
+export const InsightsTotalRiskChartWrapper = props => {
+  if (props.status === 'RESOLVED') {
+    return (
+      !isNotRhelHost(props.hostDetails) && <InsightsTotalRiskChart {...props} /> // check for RHEL hosts
+    );
+  }
+  return null;
```
bastian-src, line 12: Is it better to return an empty element here? `</>`?

bastian-src (commented):
LGTM, adding the metadata field looks like a good solution to me.

webpack/ForemanRhCloudHelpers.js
Thorben-D: Notice that the `isNotRhelHost` function is also used in webpack/ForemanRhCloudFills.js, where it is passed to hide the tab.
It [needs to be a function](https://github.com/theforeman/foreman/blob/13bdb7f81a4fcc08eee82e726bd96c644aa23a05/webpack/assets/javascripts/react_app/components/HostDetails/index.js#L89-L92) and I preferred this over wrapping it in a lambda.

bastian-src: Missed that one, alright!

webpack/InsightsHostDetailsTab/InsightsTotalRiskChartWrapper.js
Thorben-D: Unsure, the [docs](https://legacy.reactjs.org/docs/conditional-rendering.html#preventing-component-from-rendering) recommend `null`, but I read somewhere that there is a case for returning `false` instead...

webpack/ForemanRhCloudHelpers.js
sbernhard: There is no better check for this right now. I would just go with this solution.

nadjaheitmann: What about checking at least for `RHEL` ?  I don't know how people are naming their OSes these days.

sbernhard: on Redhat, these OS are created automatically from the imported Manifest. 
Other areas will not work if OS name was changed, like https://github.com/theforeman/foreman/blob/develop/app/views/unattended/provisioning_templates/snippet/insights.erb#L10

sbernhard (approved):

chris1984:
Will take a look

chris1984:
ACK, works great. Thanks @Thorben-D I am not seeing the insights tab for my Ubuntu 24.04 box

Before PR:
Screenshot 2024-10-22 150531
Screenshot 2024-10-22 150555

After PR:
Screenshot 2024-10-22 150622

chris1984 (approved):

chris1984:
@Thorben-D I am not seeing the insights tab for a normal RHEL host now, is this something you are able to investigate?

chris1984:
Screenshot 2024-10-22 155635

Thorben-D:
@chris1984 Thanks for having a look! Does the OS name of your RHEL host not contain any permutation of "Red Hat"? 
Right now, that is the check that is done to determine whether the tab should be displayed or not. 
We can't use `os_family` as that also says "Red hat" for alma-/rockylinux. 
Maybe I should use the same check that determines the OS icon. That seems to work with "RHEL"...

nadjaheitmann:
https://github.com/theforeman/foreman_rh_cloud/pull/911#discussion_r1806166368 That is what is was talking about :)

chris1984:
@Thorben-D my host name is `ip-10-0-168-58.rhos-01.prod.psi.rdu2.redhat.com` with the OS `RHEL 9.4`

chris1984:
I wonder if you could use the distribution facts like here 
Screenshot 2024-10-23 083946

chris1984:
The Icon way, would work as well I think

chris1984:
@Thorben-D Do you think you can revisit this next week? Otherwise, I will prob have to revert the commit, so it does not affect our tests downstream when we do our weekly nightly/stream builds.

Thorben-D:
@chris1984 Not before wednesday I'm afraid. If that is too late, you'll have to revert the commit.
In that case, feel free to CC me in the PR and I'll give it a review.

PR #896, opened 2024-08-27
Add convert2rhel fact to host report

jeremylenz:
To test:

#### Setup

On a registered host, create a custom fact file that includes `conversions.env.CONVERT2RHEL_THROUGH_FOREMAN`:

```
[root@toledo8 ~]# cat /etc/rhsm/facts/convert2rhel.facts {
"conversions.version": "1",
"conversions.activity": "conversion",
"conversions.packages.0.nevra": "convert2rhel-0:2.0.0-1.el8.noarch",
"conversions.packages.0.signature": "RSA/SHA256, Thu May 30 13:31:33 2024, Key ID 199e2f91fd431d51",
"conversions.executed": "/usr/bin/convert2rhel",
"conversions.success": true,
"conversions.activity_started": "2024-07-11T17:28:54.281664Z",
"conversions.activity_ended": "2024-07-11T17:48:47.026664Z",
"conversions.source_os.id": "Cerulean Leopard",
"conversions.source_os.name": "AlmaLinux",
"conversions.source_os.version": "8.10",
"conversions.target_os.id": "Ootpa",
"conversions.target_os.name": "Red Hat Enterprise Linux",
"conversions.target_os.version": "8.10",
"conversions.env.CONVERT2RHEL_THROUGH_FOREMAN": "1",
"conversions.run_id": "null"
}
```

Update facts:
```
subscription-manager facts --update
```

Verify that the host's `subscription_facet_attributes` have `convert2rhel_through_foreman: 1`

#### Test

Ensure that the host's `host_registration_insights` parameter is set to true
On the Inventory Upload page, click 'Sync inventory status' and wait for the sync to complete

On the host, run
```
organization_id=1 target=/var/lib/foreman/red_hat_inventory/generated_reports/ bundle exec rake rh_cloud_inventory:report:generate
```

You should see, at the end of the output, something like 
```
Successfully generated /tmp/d20240827-17482-t377is/report_for_1.tar.xz for organization id 1
```

CD into that directory, then run
```
$ tar -vxf report_for_1.tar.xz
./
./metadata.json
./d5f5c08b-7903-4c58-8c08-53d1f37c753c.json
```

All files except `metadata.json` will be about individual hosts. Find your host and verify its json contains `convert2rhel_through_foreman`:

```
$ cat d5f5c08b-7903-4c58-8c08-53d1f37c753c.json | jq
{
  "report_slice_id": "d5f5c08b-7903-4c58-8c08-53d1f37c753c",
  "hosts": [
    {
[...]

PR #914, opened 2024-10-22
Add additional fields for convert2RHEL

jeremylenz:
This adds several convert2RHEL fields to the generated report:

```
"conversions": {
        "source_os": {
          "name": "AlmaLinux",
          "version": "8.10"
        },
        "target_os": {
          "name": "Red Hat Enterprise Linux",
          "version": "8.10"
        },
        "convert2rhel_through_foreman": 1,
        "activity": "conversion",
        "packages_0_nevra": "convert2rhel-0:2.0.0-1.el8.noarch",
        "packages_0_signature": "RSA/SHA256, Thu May 30 13:31:33 2024, Key ID 199e2f91fd431d51",
        "activity_started": "2024-07-11T17:28:54.281664Z",
        "activity_ended": "2024-07-11T17:48:47.026664Z",
        "success": "true"
      },
```

I had to learn several things the hard way here:

1. For some reason, RH Cloud seems to have _its own JSON generator_ in `lib/foreman_inventory_upload/generators/json_stream.rb`. !!??
2. In order to use the `simple_field` and `object_field` methods of this JSON generator properly, you must pass a second argument, `last`. If you don't pass `last`, and your field happens to be the last field in a object, the stream will include an extra comma and thus stop being valid JSON.
3. In order to be able to use the `fact_values` method, you have to add the fact names to `self.fact_names` in `lib/foreman_inventory_upload/generators/queries.rb`. If a fact name returns nil, `simple_field` will just return and skip that line. This causes problems with no. 2 above, because you never know for sure if a field will be the last field. (I'm guessing this is the reason for the `comma` method.

### Testing
see the instructions on https://github.com/theforeman/foreman_rh_cloud/pull/896
Make sure to test with different types of hosts (insights-enabled, insights-disabled)

Pipe your report output through `jq` (that's also a handy way to know if it's valid JSON) and make sure all the new values show up.

### DIFF
--- a/lib/foreman_inventory_upload/generators/queries.rb
+++ b/lib/foreman_inventory_upload/generators/queries.rb
@@ -28,6 +28,16 @@ module ForemanInventoryUpload
               'insights_client::hostname',
               'insights_client::ips',
               'insights_id',
+              'conversions::activity',
+              'conversions::packages::0::nevra',
+              'conversions::packages::0::signature',
+              'conversions::activity_started',
+              'conversions::activity_ended',
+              'conversions::success',
+              'conversions::source_os::name',
+              'conversions::source_os::version',
+              'conversions::target_os::name',
+              'conversions::target_os::version',
             ]).pluck(:name, :id)
           ]
       end
--- a/lib/foreman_inventory_upload/generators/slice.rb
+++ b/lib/foreman_inventory_upload/generators/slice.rb
@@ -40,6 +40,16 @@ module ForemanInventoryUpload
         end
       end
 
+      def report_conversions(host)
+        @stream.simple_field('convert2rhel_through_foreman', host.subscription_facet&.convert2rhel_through_foreman)
+        @stream.simple_field('activity', fact_value(host, 'conversions::activity'))
+        @stream.simple_field('packages_0_nevra', fact_value(host, 'conversions::packages::0::nevra'))
+        @stream.simple_field('packages_0_signature', fact_value(host, 'conversions::packages::0::signature'))
+        @stream.simple_field('activity_started', fact_value(host, 'conversions::activity_started'))
+        @stream.simple_field('activity_ended', fact_value(host, 'conversions::activity_ended'))
+        @stream.simple_field('success', fact_value(host, 'conversions::success') || false, :last)
+      end
+
       def report_host(host)
         host_ips_cache = host_ips(host)
         @stream.object do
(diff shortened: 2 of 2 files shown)


chris1984 (commented):
Looks good, sorry you had to go through that json mess. I did see one thing we have the `convert2rhel_through_foreman` in the report twice:

```json
  "hosts": [
    {
      "fqdn": "server.example.com",
      "account": "5235210",
      "subscription_manager_id": "2aa64a0c-a33e-4c7a-aae4-b8b6789a43f3",
      "satellite_id": "2aa64a0c-a33e-4c7a-aae4-b8b6789a43f3",
      "convert2rhel_through_foreman": 1,
      "conversions": {
        "source_os": {
          "name": "AlmaLinux",
          "version": "8.10"
        },
        "target_os": {
          "name": "Red Hat Enterprise Linux",
          "version": "8.10"
        },
        "convert2rhel_through_foreman": 1,
        "activity": "conversion",
        "packages_0_nevra": "convert2rhel-0:2.0.0-1.el8.noarch",
        "packages_0_signature": "RSA/SHA256, Thu May 30 13:31:33 2024, Key ID 199e2f91fd431d51",
        "activity_started": "2024-07-11T17:28:54.281664Z",
        "activity_ended": "2024-07-11T17:48:47.026664Z",
        "success": "true"
      },
```

Did you want to remove the top level one and just use yours?

jeremylenz:
> Did you want to remove the top level one and just use yours?

I left both in because I couldn't decide. 

@bocekm thoughts on the above data shape? We could remove the top-level `convert2rhel_through_foreman` if nothing relies on it yet.

ShimShtein:
@jeremylenz The reason we have our own implementation for JSON is because the standard one does not have a notion of streaming. If we use the regular JSON, it will create a ton of objects in Ruby's heap, which is not desired, especially if you are working with large JSONs. This is also the reason that we have the `comma` and `last` methods - we need to know if we should write the comma into the string, or is it the last object, and hence the comma is not needed.

If you have a better idea of last object detection and emission, I would really appreciate it.

My way to deal with the `last` parameter was to put the fields that we're not sure that they will appear before a field that is always written. This way we can be sure that the always written field is the last one. It's not bulletproof, but it worked till now.

jeremylenz:
> The reason we have our own implementation for JSON is because the standard one does not have a notion of streaming. If we use the regular JSON, it will create a ton of objects in Ruby's heap, which is not desired, especially if you are working with large JSONs. This is also the reason that we have the `comma` and `last` methods - we need to know if we should write the comma into the string, or is it the last object, and hence the comma is not needed.
> 
> If you have a better idea of last object detection and emission, I would really appreciate it.

Thanks for explaining :) My first thoughts were to use rabl, or to build a Ruby hash first and then convert it to json. But to be honest I haven't done much of a deep dive into how to minimize allocations etc., or if either of those would help in that context. We could add that to our list for RHCloud.

jeremylenz:
Added default values to a few `simple_field` calls to ensure that we don't generate invalid JSON. Verified that generating the report works both with and without those facts present.

jeremylenz:
Test failures unrelated, I think. That test passes locally; can we try running again?

bocekm:
> @bocekm thoughts on the above data shape? We could remove the top-level convert2rhel_through_foreman if nothing relies on it yet.

You can remove the top level one. Nothing relies on it so far, reading it in Tableau/Grokket is the next step (https://issues.redhat.com/browse/PNTBIZIN-12423) after all these conversion related changes to rh_cloud are shipped and available in Satellite.

jeremylenz:
Removed the top-level `convert2rhel_through_foreman` node.

chris1984 (approved):
Looks good to me now that the top level convert2rhel is removed.

chris1984:
@ColeHiggins2 can you test this one please :)

ColeHiggins2:
@jeremylenz @chris1984 Verified and tested with Packit

ColeHiggins2 (approved):

bocekm:
It seems to me that the "success" field should be returing true/false without quotes. With quotes it's a string and not boolean, isn't it?

PR #915, opened 2024-10-24
Fix uninitialized constant ForemanRhCloud::VERSION

stejskalleos:
Workflows:
- insights client registration
- rh_cloud_inventory:report:generate task

### DIFF
--- a/config/initializers/zeitwerk.rb
+++ b/config/initializers/zeitwerk.rb
@@ -1,3 +1,5 @@
+require "#{ForemanRhCloud::Engine.root}/lib/foreman_rh_cloud/version"
+
 Rails.autoloaders.main.ignore(
   ForemanRhCloud::Engine.root.join('lib/foreman_rh_cloud/version.rb')
 )


stejskalleos:
@ofedoren

ofedoren:
Thanks, @stejskalleos, that could do it. We could also try to do it future-proof: since we already remove that constant from Zeitwerk loader, we can explicitly require it in the same initializer, so we don't need to require it in each file that potentially uses that constant.
In https://github.com/theforeman/foreman_rh_cloud/blob/develop/config/initializers/zeitwerk.rb:
```ruby
require 'foreman_rh_cloud/version'

Rails.autoloaders.main.ignore(
  ForemanRhCloud::Engine.root.join('lib/foreman_rh_cloud/version.rb')
)
```

chris1984 (commented):
I like @ofedoren method so we don't have to include it in each file, @stejskalleos can you update the pr and we can test

jeremylenz:
I've also been getting some problems in foreman_theme_satellite with `Rails.autoloaders.main.ignore` - apparently `main` doesn't exist sometimes?

ofedoren:
> I've also been getting some problems in foreman_theme_satellite with `Rails.autoloaders.main.ignore` - apparently `main` doesn't exist sometimes?

AFAIU, `.main` comes with Zeitwerk. Before that I think it was `.autoloaders.each` for Rails' ones?

ofedoren (approved):
Thanks, @stejskalleos, seems to fix the issue

chris1984 (approved):
Thanks @stejskalleos

PR #921, opened 2024-11-08
Add Global params for host obfuscation and migrate off settings

chris1984:
* Created a global parameter for `obfuscate_inventory_hostnames` and `obfuscate_inventory_ips` and set them to `false` which is the default value for `insights-client`
* Created a DB migration to set the values of the global parameters to the value of the settings for `obfuscate_inventory_hostnames` and `obfuscate_inventory_ips` and deletes the settings from the plugin and DB
* Removed the UI toggles
* Updated UI and JS tests to use the new parameters
* These parameters will be used for the REX job template for existing hosts to configure host obfuscation and for global registration.

### DIFF
--- a/db/migrate/20241108165827_convert_obfuscation_settings_to_params.rb
+++ b/db/migrate/20241108165827_convert_obfuscation_settings_to_params.rb
@@ -1,0 +1,22 @@
+class ConvertObfuscationSettingsToParams < ActiveRecord::Migration[6.1]
+  def change
+    # Create the common parameters if they don't exist and set the default values to false
+    CommonParameter.find_or_create_by!(name: 'obfuscate_inventory_hostnames', value: false)
+    CommonParameter.find_or_create_by!(name: 'obfuscate_inventory_ips', value: false)
+
+    # Copy the settings to the common parameters and remove the settings from the database
+    # rubocop:disable Style/GuardClause
+    if Setting.find_by(name: 'obfuscate_inventory_hostnames')&.value
+      hostname_setting = CommonParameter.find_by(name: 'obfuscate_inventory_hostnames')
+      hostname_setting.update!(value: Setting.find_by(name: 'obfuscate_inventory_hostnames')&.value)
+      Setting.find_by(name: 'obfuscate_inventory_hostnames').destroy_all
+    end
+
+    if Setting.find_by(name: 'obfuscate_inventory_ips')&.value
+      ip_setting = CommonParameter.find_by(name: 'obfuscate_inventory_ips')
+      ip_setting.update!(value: Setting.find_by(name: 'obfuscate_inventory_ips')&.value)
+      Setting.find_by(name: 'obfuscate_inventory_ips').destroy_all
+    end
+    # rubocop:enable Style/GuardClause
+  end
+end
--- a/lib/foreman_inventory_upload/generators/fact_helpers.rb
+++ b/lib/foreman_inventory_upload/generators/fact_helpers.rb
@@ -61,7 +61,7 @@ module ForemanInventoryUpload
         insights_client_setting = ActiveModel::Type::Boolean.new.cast(insights_client_setting)
         return insights_client_setting unless insights_client_setting.nil?
 
-        Setting[:obfuscate_inventory_hostnames]
+        CommonParameter.find_by(name: 'obfuscate_inventory_hostnames')&.value
       end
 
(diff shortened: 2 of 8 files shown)


db/migrate/20241108165827_convert_obfuscation_settings_to_params.rb
```diff
@@ -0,0 +1,22 @@
+class ConvertObfuscationSettingsToParams < ActiveRecord::Migration[6.1]
+  def change
+    # Create the common parameters if they don't exist and set the default values to false
+    CommonParameter.find_or_create_by!(name: 'obfuscate_inventory_hostnames', value: false)
+    CommonParameter.find_or_create_by!(name: 'obfuscate_inventory_ips', value: false)
```
ShimShtein, line 5:
```suggestion
    hostname_setting = CommonParameter.find_or_create_by!(name: 'obfuscate_inventory_hostnames', value: false)
    ip_setting = CommonParameter.find_or_create_by!(name: 'obfuscate_inventory_ips', value: false)
```

You can skip the `find_by`  later in the code

lib/foreman_inventory_upload/generators/fact_helpers.rb
```diff
@@ -61,7 +61,7 @@ def obfuscate_hostname?(host)
         insights_client_setting = ActiveModel::Type::Boolean.new.cast(insights_client_setting)
         return insights_client_setting unless insights_client_setting.nil?
 
-        Setting[:obfuscate_inventory_hostnames]
+        CommonParameter.find_by(name: 'obfuscate_inventory_hostnames')&.value
```
ShimShtein, line 64: You should look for a value on the host, not just the common parameter, similar to https://github.com/theforeman/foreman_rh_cloud/blob/48f9d74ba8a671c74cec3a79b2646c38172a42a3/app/controllers/insights_cloud/api/machine_telemetries_controller.rb#L87

lib/foreman_inventory_upload/generators/fact_helpers.rb
```diff
@@ -79,7 +79,7 @@ def obfuscate_ips?(host)
         insights_client_setting = ActiveModel::Type::Boolean.new.cast(insights_client_setting)
         return insights_client_setting unless insights_client_setting.nil?
 
-        Setting[:obfuscate_inventory_ips]
+        CommonParameter.find_by(name: 'obfuscate_inventory_ips')&.value
```
ShimShtein, line 82: If it's a parameter, you have to look up the value on the host object itself: https://github.com/theforeman/foreman_rh_cloud/blob/48f9d74ba8a671c74cec3a79b2646c38172a42a3/app/controllers/insights_cloud/api/machine_telemetries_controller.rb#L87

lib/foreman_inventory_upload/generators/fact_helpers.rb
```diff
@@ -114,7 +114,7 @@ def hostname_match
         foreman_hostname = ForemanRhCloud.foreman_host&.name
         if bash_hostname == foreman_hostname
           fqdn(ForemanRhCloud.foreman_host)
-        elsif Setting[:obfuscate_inventory_hostnames]
+        elsif CommonParameter.find_by(name: 'obfuscate_inventory_hostnames')&.value
```
ShimShtein, line 117: If it's a parameter, you have to look up the value on the host object itself: https://github.com/theforeman/foreman_rh_cloud/blob/48f9d74ba8a671c74cec3a79b2646c38172a42a3/app/controllers/insights_cloud/api/machine_telemetries_controller.rb#L87

test/unit/metadata_generator_test.rb
```diff
@@ -23,8 +23,8 @@ class MetadataGeneratorTest < ActiveSupport::TestCase
   end
 
   test 'generates an empty report with hidden hostname and ip' do
-    Setting[:obfuscate_inventory_hostnames] = true
-    Setting[:obfuscate_inventory_ips] = true
+    CommonParameter.find_or_create_by!(name: 'obfuscate_inventory_ips', value: true)
```
ShimShtein, line 26: If the parameter exists, I think this method will not reset its value.

ShimShtein (commented):
I think we will need to add a UI message on the inventory page about the ability to obfuscate using the parameter. Otherwise it will feel to the user that the setting just disappeared without replacement.

PR #927, opened 2024-12-03
Fix About subscription watch link on RH Cloud Inventory page

Odilhao:


chris1984 (approved):

foreman_rh_cloud.gemspec
```diff
@@ -24,6 +24,7 @@ Gem::Specification.new do |s|
   s.add_dependency 'foreman-tasks'
   s.add_runtime_dependency 'katello', '>= 4.14.0.rc1.1'
 
+  s.add_development_dependency 'minitest', '5.25.1'
```
evgeni: plugin dev deps are ignored during testing (you'd need a `gemfile.d/test.rb` like [katello has](https://github.com/Katello/katello/blob/master/gemfile.d/test.rb) for this to work), so drop this line please

evgeni:
It fails on Zeitwerk stuff, because the tests run against develop but this branch doesn't support Zeitwerk loader.

https://github.com/theforeman/foreman_rh_cloud/pull/930 should fix that

chris1984 (approved):

PR #932, opened 2024-12-16
Update main nav and UI for local Advisor Engine

jeremylenz:
Introduce a new SETTING to tell RH Cloud if you're using Insights on Prem. In response to this setting,
- hide the Inventory Upload page from the main nav
- hide the "Sync automatically" toggle from the Recommendations page (formerly the Insights page)
- hide the 'View in Red Hat Insights' links from various locations on the Recommendations page (formerly the Insights page)

goes with https://github.com/theforeman/foreman_rh_cloud/pull/939

#### Before
image

#### After
image

### Testing

In your Foreman `settings.yaml`, include a new section
```
:foreman_rh_cloud:
  :use_local_advisor_engine: true
```


parthaa:
Can you also hide both these entries from the insights recommendations page. 

image

Same for 
image


Looks good otherwise.

parthaa:
Also recommended patch. This redirects the remediations + certs  to use localhost . May make sense in a separate PR though

```patch
diff --git a/lib/foreman_rh_cloud.rb b/lib/foreman_rh_cloud.rb
index f9e6430..406703f 100644
--- a/lib/foreman_rh_cloud.rb
+++ b/lib/foreman_rh_cloud.rb
@@ -3,17 +3,23 @@ require 'cgi'
 require 'uri'
 
 module ForemanRhCloud
+  def self.on_premise_url
+    return unless ForemanRhCloud.with_insights_on_premise?
+    port = ENV['INSIGHTS_ADVISOR_PORT'] || "8000"
+    "http://#{ForemanRhCloud.foreman_host.fqdn || 'localhost'}:#{port}"
+  end
+
   def self.base_url
     # for testing set ENV to 'https://ci.cloud.redhat.com'
-    @base_url ||= ENV['SATELLITE_RH_CLOUD_URL'] || 'https://cloud.redhat.com'
+    @base_url ||= on_premise_url || ENV['SATELLITE_RH_CLOUD_URL'] || 'https://cloud.redhat.com'
   end
 
   def self.cert_base_url
-    @cert_base_url ||= ENV['SATELLITE_CERT_RH_CLOUD_URL'] || 'https://cert.cloud.redhat.com'
+    @cert_base_url ||= on_premise_url || ENV['SATELLITE_CERT_RH_CLOUD_URL'] || 'https://cert.cloud.redhat.com'
   end
 
   def self.legacy_insights_url
-    @legacy_insights_url ||= ENV['SATELLITE_LEGACY_INSIGHTS_URL'] || 'https://cert-api.access.redhat.com'
+    @legacy_insights_url ||= on_premise_url || ENV['SATELLITE_LEGACY_INSIGHTS_URL'] || 'https://cert-api.access.redhat.com'
   end
 
   def self.verify_ssl_method
```

jeremylenz:
> Also recommended patch. This redirects the remediations + certs to use localhost . May make sense in a separate PR though

I agree, I think this makes more sense as a separate PR. Opened https://github.com/theforeman/foreman_rh_cloud/pull/936

lib/foreman_rh_cloud/engine.rb
```diff
@@ -193,4 +199,8 @@ def self.register_scheduled_task(task_class, cronline)
       end
     end
   end
+
+  def self.with_insights_on_premise?
+    SETTINGS&.[](:insights)&.[](:use_insights_on_premise) || false
```
ShimShtein: Assuming we push the pot definition PR, can we rely on that ENV variable instead of introducing yet another setting?

lib/insights_cloud/async/insights_full_sync.rb
```diff
@@ -121,6 +121,7 @@ def to_model_hash(hit_hash)
           likelihood: hit_hash['likelihood'].to_i,
           results_url: hit_hash['results_url'],
           rule_id: to_rule_id(hit_hash['results_url']),
+          is_local_insights_advisor: ::ForemanRhCloud.use_insights_on_premise?,
```
ShimShtein:
```suggestion
          is_local_insights_advisor: ::ForemanRhCloud.with_insights_on_premise?,
```

lib/foreman_rh_cloud/engine.rb
jeremylenz: I think @ehelms and Insights folks already decided on having an installer option, which is what this reflects. right eric?

jeremylenz: side note: The name of the setting (and application) is currently in the bikeshedding phase; I will update this PR when it's decided.

ehelms: We once tried to make it so that we could handle SETTINGS through env variables,  but I think it's currently restrictive and doesn't handle plugins, see https://github.com/theforeman/foreman/blob/develop/config/settings.rb#L11

ehelms: The biggest issue I see with relying purely on environment variables is that we are ending up with two different paradigms for how we configure Foreman and plugins. We have the standard model of deploying a plugin specific configuration file and then we have this environment variable model that results in deploying systemd override files. I generally just don't like splitting the design because it creates mental overhead.

As I understand things, today, foreman_rh_cloud does not have a config file like other plugins and instead only relies on environment variables.

@ekohl @evgeni -- do you have an opinion on this design split?

evgeni: Do we need to make this selection org-aware? ENV vars aren't ;)

parthaa: ENV vars are definitely a hack. But isn't this PR using SETTINGS ?

ehelms: > Do we need to make this selection org-aware? ENV vars aren't ;)

IIRC, @parthaa got agreement to start with either all local or all connected, and not per-org and to revisit that later.

parthaa: Yes. All Connected or All Local. We may make it org-based in the future.

jeremylenz:
Updated to use `local_advisor_engine` naming per offline discussion.

ehelms:
You could consider splitting out the first part, navigation rename, to a stand-alone PR as the two changes are independent.

lib/foreman_rh_cloud/engine.rb
ekohl: > Do we need to make this selection org-aware? ENV vars aren't ;)

`SETTINGS` isn't org aware either.

Also, it can be written using `dig`.
```ruby
SETTINGS.dig(:insights, :use_insights_on_premise) || false
```

app/controllers/insights_cloud/settings_controller.rb
```diff
@@ -1,6 +1,12 @@
 module InsightsCloud
   class SettingsController < ::ApplicationController
     def show
+      if ForemanRhCloud.with_local_advisor_engine?
+        render json: {
+          error: _('This Foreman is configured to use Insights on Premise. Syncing to Insights Cloud is disabled.'),
+        }, status: :unprocessable_entity
+        return
+      end
       render_setting(:insightsSyncEnabled, :allow_auto_insights_sync)
```
ekohl: Nit: is this easier to follow? I generally try to avoid `return` and rely on the visual indentation to follow the code.

```suggestion
      else
        render_setting(:insightsSyncEnabled, :allow_auto_insights_sync)
      end
```

Though taking a step back, from an API perspective I wonder why this is even here in the first place. We already have an API to show settings: `/api/settings/:id` that supports both [GET](https://apidocs.theforeman.org/katello/latest/apidoc/v2/settings/show.html) and [PUT](https://apidocs.theforeman.org/katello/latest/apidoc/v2/settings/update.html).

Are there any permission checks on this controller to prevent unauthenticated users to change the setting or can any user change this for any organization?

lib/foreman_rh_cloud/engine.rb
```diff
@@ -193,4 +199,8 @@ def self.register_scheduled_task(task_class, cronline)
       end
     end
   end
+
+  def self.with_local_advisor_engine?
+    SETTINGS&.[](:foreman_rh_cloud)&.[](:use_local_advisor_engine) || false
```
ekohl: You can use [dig](https://ruby-doc.org/3.4.1/Hash.html#method-i-dig) on a hash and I think it's safe to assume `SETTINGS` is a hash (so `&` is redundant):
```suggestion
    SETTINGS.dig(:foreman_rh_cloud, :use_local_advisor_engine) || false
```

lib/foreman_rh_cloud/engine.rb
evgeni: Right. `SETTINGS`, `Settings`, `settings`…

webpack/InsightsCloudSync/Components/InsightsSettings/InsightsSettingsActions.js
```diff
@@ -19,7 +19,11 @@ export const getInsightsSyncSettings = () => async dispatch => {
         },
       },
     });
-  } catch ({ message }) {
+  } catch (err) {
+    const { message } = err;
+    if (err?.response?.status === 422) {
```
ekohl: Should this catch any non-2xx code? Any network connection can fail so it could return 5xx.

ekohl (commented):
> You could consider splitting out the first part, navigation rename, to a stand-alone PR as the two changes are independent.

I think that's also a wise thing.

As for the specific setting: I wonder if it should be part of [ForemanContext](https://github.com/theforeman/foreman/blob/develop/webpack/assets/javascripts/react_app/Root/Context/ForemanContext.js) though we may not have a good way to extend that from plugins.

Design wise I don't like that you can change a global setting from regular workflows. That's not how we do things in the application because it completely ignores the multi tenancy model.

webpack/InsightsCloudSync/Components/InsightsSettings/InsightsSettingsActions.js
jeremylenz: Since I send myself the 422 specifically for this check, I thought it made sense to check for it specifically as well. But I may end up re-architecting this whole thing anyway, stand by..

jeremylenz:
> As for the specific setting: I wonder if it should be part of [ForemanContext](https://github.com/theforeman/foreman/blob/develop/webpack/assets/javascripts/react_app/Root/Context/ForemanContext.js) though we may not have a good way to extend that from plugins.

This was one of the first things I tried; it would make a lot of sense! It was apparent after an hour or so that adding it to ForemanContext was going to be slow and painful, because plugins were not considered at all there and I hit some dead ends trying to overwrite Ruby private methods from a plugin. I'm sure it's possible somehow but I've moved on to other solutions now. I think I may just add a small API endpoint.

jeremylenz:
Tasos found that I can't use React context for InsightsTable because the component is reused on the host details page, and thus breaks. So stand by for more updates.

webpack/InsightsCloudSync/Components/InsightsSettings/InsightsSettingsActions.js
ekohl: In that case, perhaps it should raise a more specific exception than `Error`?

parthaa:
image


Please hide sync recommendations

lib/foreman_rh_cloud/engine.rb
```diff
@@ -74,6 +74,7 @@ def self.register_scheduled_task(task_class, cronline)
               'foreman_inventory_upload/cloud_status': [:index],
               'foreman_inventory_upload/uploads_settings': [:index],
               'foreman_inventory_upload/missing_hosts': [:index],
+              'api/v2/rh_cloud/advisor_engine_config': [:show],
```
jeremylenz, line 77: This makes the endpoint require `:view_foreman_rh_cloud` permission. that ok with you @parthaa ?

app/controllers/insights_cloud/settings_controller.rb
jeremylenz: > Are there any permission checks on this controller to prevent unauthenticated users to change the setting or can any user change this for any organization?

It looks like settings endpoints need the `:generate_foreman_rh_cloud` permission:

https://github.com/theforeman/foreman_rh_cloud/blob/dc80828e3949cd9967121f0a251639561ec2124a/lib/foreman_rh_cloud/engine.rb#L63

jeremylenz:
:green_apple:

lib/foreman_rh_cloud/engine.rb
parthaa: I am good with that

parthaa (approved):
Works well

PR #935, opened 2024-12-19
SAT-15137 - Rename inventory sync and restart button

chris1984:
* Renamed restart to Generate and upload report
* Renamed Sync inventory status to Sync all Inventory Status
* Renamed Configure Cloud Connector to Configure cloud connector

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,7 +15,7 @@ for how to install Foreman plugins
 
 #### Inventory upload
 
-In UI: Configure -> Inventory Upload -> Restart
+In UI: Configure -> Inventory Upload -> Generate and upload report 
 
 From command-line:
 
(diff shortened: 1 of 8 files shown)


lfu:
Before
b4

After
after

Do you think ```Configure Could Connector``` should be fixed with `Configure cloud connector`?
Other than this, it looks good.

chris1984:
> Before b4
> 
> After after
> 
> Do you think `Configure Could Connector` should be fixed with `Configure cloud connector`? Other than this, it looks good.

Yeah I can change that. Let me push an update up

chris1984:
@lfu updated

lfu (approved):
LGTM :+1:  Nice work!

PR #936, opened 2024-12-23
Allow ENV variables for urls

jeremylenz:
This is @parthaa's recommended patch from https://github.com/theforeman/foreman_rh_cloud/pull/932#issuecomment-2558014648

Requires https://github.com/theforeman/foreman_rh_cloud/pull/932

### DIFF
--- a/lib/foreman_rh_cloud.rb
+++ b/lib/foreman_rh_cloud.rb
@@ -3,17 +3,27 @@ require 'cgi'
 require 'uri'
 
 module ForemanRhCloud
+  def self.on_premise_url
+    return unless ForemanRhCloud.with_local_advisor_engine?
+    port = ENV['ADVISOR_ENGINE_PORT'] || "8000"
+    ENV['ADVISOR_ENGINE_URL'] || "http://#{ForemanRhCloud.foreman_host.fqdn || 'localhost'}:#{port}"
+  end
+
+  def self.env_or_on_premise_url(env_var_name)
+    ENV[env_var_name] || on_premise_url
+  end
+
   def self.base_url
     # for testing set ENV to 'https://ci.cloud.redhat.com'
-    @base_url ||= ENV['SATELLITE_RH_CLOUD_URL'] || 'https://cloud.redhat.com'
+    @base_url ||= env_or_on_premise_url('SATELLITE_RH_CLOUD_URL') || 'https://cloud.redhat.com'
   end
 
   def self.cert_base_url
-    @cert_base_url ||= ENV['SATELLITE_CERT_RH_CLOUD_URL'] || 'https://cert.cloud.redhat.com'
+    @cert_base_url ||= env_or_on_premise_url('SATELLITE_CERT_RH_CLOUD_URL') || 'https://cert.cloud.redhat.com'
   end
 
   def self.legacy_insights_url
-    @legacy_insights_url ||= ENV['SATELLITE_LEGACY_INSIGHTS_URL'] || 'https://cert-api.access.redhat.com'
+    @legacy_insights_url ||= env_or_on_premise_url('SATELLITE_LEGACY_INSIGHTS_URL') || 'https://cert-api.access.redhat.com'
   end
 
   def self.verify_ssl_method


lib/foreman_rh_cloud.rb
```diff
@@ -3,17 +3,23 @@
...
+  def self.on_premise_url
+    return unless ForemanRhCloud.with_insights_on_premise?
+    port = ENV['INSIGHTS_ADVISOR_PORT'] || "8000"
+    "http://#{ForemanRhCloud.foreman_host.fqdn || 'localhost'}:#{port}"
+  end
+
   def self.base_url
     # for testing set ENV to 'https://ci.cloud.redhat.com'
-    @base_url ||= ENV['SATELLITE_RH_CLOUD_URL'] || 'https://cloud.redhat.com'
+    @base_url ||= on_premise_url || ENV['SATELLITE_RH_CLOUD_URL'] || 'https://cloud.redhat.com'
```
ShimShtein:
```suggestion
    @base_url ||= ENV['SATELLITE_RH_CLOUD_URL'] || on_premise_url || 'https://cloud.redhat.com'
```

I think we should leave the option of explicitly defining the URL even if the port is defined

jeremylenz: updated :+1:

chris1984 (dismissed):
Looks fine to me! @ShimShtein any other concerns?

jeremylenz:
Should merge [the other one](https://github.com/theforeman/foreman_rh_cloud/pull/932) first

lib/foreman_rh_cloud.rb
```diff
@@ -3,17 +3,23 @@
...
   end
 
   def self.cert_base_url
-    @cert_base_url ||= ENV['SATELLITE_CERT_RH_CLOUD_URL'] || 'https://cert.cloud.redhat.com'
+    @cert_base_url ||= on_premise_url || ENV['SATELLITE_CERT_RH_CLOUD_URL'] || 'https://cert.cloud.redhat.com'
   end
 
   def self.legacy_insights_url
-    @legacy_insights_url ||= ENV['SATELLITE_LEGACY_INSIGHTS_URL'] || 'https://cert-api.access.redhat.com'
+    @legacy_insights_url ||= on_premise_url || ENV['SATELLITE_LEGACY_INSIGHTS_URL'] || 'https://cert-api.access.redhat.com'
```
evgeni: the `ENV` vs `on_premise_url` priority seems inconsistent. I think `ENV` should "always" win?

jeremylenz:
@evgeni updated :+1:

lib/foreman_rh_cloud.rb
```diff
@@ -3,17 +3,27 @@
 require 'uri'
 
 module ForemanRhCloud
+  def self.on_premise_url
+    return unless ForemanRhCloud.with_insights_on_premise?
+    port = ENV['INSIGHTS_ADVISOR_PORT'] || "8000"
+    "http://#{ForemanRhCloud.foreman_host.fqdn || 'localhost'}:#{port}"
```
evgeni:
```suggestion
    "https://#{ForemanRhCloud.foreman_host.fqdn || 'localhost'}:#{port}"
```

evgeni: (or make it configurable… but we need SSL for real deployments)

jeremylenz: @parthaa :eyes:

lib/foreman_rh_cloud.rb
```diff
@@ -3,17 +3,27 @@
...
   end
 
   def self.cert_base_url
-    @cert_base_url ||= ENV['SATELLITE_CERT_RH_CLOUD_URL'] || 'https://cert.cloud.redhat.com'
+    @cert_base_url ||= env_or_on_premise_url('SATELLITE_CERT_RH_CLOUD_URL') || 'https://cert.cloud.redhat.com'
   end
 
   def self.legacy_insights_url
-    @legacy_insights_url ||= ENV['SATELLITE_LEGACY_INSIGHTS_URL'] || 'https://cert-api.access.redhat.com'
+    @legacy_insights_url ||= env_or_on_premise_url('SATELLITE_LEGACY_INSIGHTS_URL') || 'https://cert-api.access.redhat.com'
```
evgeni, line 26: For this one, we explicitly set a `ca_file`:
https://github.com/theforeman/foreman_rh_cloud/blob/d642c75dc2b2701ba9ab412e09dfefd9778f2c0e/app/services/foreman_rh_cloud/cloud_request_forwarder.rb#L81
https://github.com/theforeman/foreman_rh_cloud/blob/d642c75dc2b2701ba9ab412e09dfefd9778f2c0e/lib/foreman_rh_cloud.rb#L116-L118

Which will *not* validate our newly injected URL. Halp.

lib/foreman_rh_cloud.rb
jeremylenz: I could add another env variable for the protocol?

evgeni: or you just let people set `ADVISOR_URL`?

lib/foreman_rh_cloud.rb
jeremylenz: Not sure what change is needed here?

lib/foreman_rh_cloud.rb
jeremylenz: updated :+1:

lib/foreman_rh_cloud.rb
evgeni: I *think* (but don't know the code well enough), that we need to enhance https://github.com/theforeman/foreman_rh_cloud/blob/d642c75dc2b2701ba9ab412e09dfefd9778f2c0e/app/services/foreman_rh_cloud/cloud_request_forwarder.rb#L81 to only use the `legacy_insights_ca` if the  `legacy_insights_url` does not belong into `access.redhat.com`, or something like that?

evgeni: but then again I don't even know what the "legacy" in there means :D

lib/foreman_rh_cloud.rb
```diff
@@ -3,17 +3,27 @@
 require 'uri'
 
 module ForemanRhCloud
+  def self.on_premise_url
+    return unless ForemanRhCloud.with_local_advisor_engine?
+    port = ENV['ADVISOR_ENGINE_PORT'] || "8000"
+    ENV['ADVISOR_ENGINE_URL'] || "http://#{ForemanRhCloud.foreman_host.fqdn || 'localhost'}:#{port}"
```
parthaa, line 9: Tested it. Https works. So lets add that

```suggestion
    ENV['ADVISOR_ENGINE_URL'] || "https://#{ForemanRhCloud.foreman_host.fqdn || 'localhost'}:#{port}"
```

lib/foreman_rh_cloud.rb
parthaa: So the original LEGACY env vars were added because qe wanted to test with staging console url instead of the prod. But we are kinda leveraging that here. The way this code is written right now in this PR I think we dont need to set any of the LEGACY env vars if its on premise. If they are set they'll be preferred.

lib/foreman_rh_cloud.rb
evgeni: FWIW, the current deployment will only ever work for `localhost` with HTTPS, as that's what https://github.com/theforeman/puppet-certs/pull/474 generates.

lib/foreman_rh_cloud.rb
parthaa: > to only use the `legacy_insights_ca` if the `legacy_insights_url` does not belong into `access.redhat.com`, or something like that?

something like?

```patch
diff --git a/app/services/foreman_rh_cloud/cloud_request_forwarder.rb b/app/services/foreman_rh_cloud/cloud_request_forwarder.rb
index e44146c..30a860e 100644
--- a/app/services/foreman_rh_cloud/cloud_request_forwarder.rb
+++ b/app/services/foreman_rh_cloud/cloud_request_forwarder.rb
@@ -74,12 +74,15 @@ module ForemanRhCloud
           ssl_client_key: OpenSSL::PKey.read(certs[:key]),
         }
       else # Legacy insights API
-        {
+        legacy_params = {
           url: ForemanRhCloud.legacy_insights_url + request_path.sub('/redhat_access/r/insights', '/r/insights'),
           ssl_client_cert: OpenSSL::X509::Certificate.new(certs[:cert]),
           ssl_client_key: OpenSSL::PKey.read(certs[:key]),
-          ssl_ca_file: ForemanRhCloud.legacy_insights_ca,
         }
+        if ForemanRhCloud.using_console?
+          legacy_params[:ssl_ca_file] = ForemanRhCloud.legacy_insights_ca,
+        end
+        legacy_params
       end
     end
 
diff --git a/lib/foreman_rh_cloud.rb b/lib/foreman_rh_cloud.rb
index 3dd77e7..40cc3bc 100644
--- a/lib/foreman_rh_cloud.rb
+++ b/lib/foreman_rh_cloud.rb
@@ -3,6 +3,8 @@ require 'cgi'
 require 'uri'
 
 module ForemanRhCloud
+  CONSOLE_INSIGHTS_URL = "https://cert-api.access.redhat.com".freeze
+
   def self.on_premise_url
     return unless ForemanRhCloud.with_local_advisor_engine?
     port = ENV['ADVISOR_ENGINE_PORT'] || "8000"
@@ -23,7 +25,11 @@ module ForemanRhCloud
   end
 
   def self.legacy_insights_url
-    @legacy_insights_url ||= env_or_on_premise_url('SATELLITE_LEGACY_INSIGHTS_URL') || 'https://cert-api.access.redhat.com'
+    @legacy_insights_url ||= env_or_on_premise_url('SATELLITE_LEGACY_INSIGHTS_URL') || CONSOLE_INSIGHTS_URL
+  end
+
+  def self.using_console?
+    legacy_insights_url == CONSOLE_INSIGHTS_URL
   end
```

parthaa: Or may be patch 2 which adds a new `SATELLITE_LEGACY_INSIGHTS_CA_PATH` if they want to provide custom value. 

```patch
diff --git a/lib/foreman_rh_cloud.rb b/lib/foreman_rh_cloud.rb
index 3dd77e7..cbeabff 100644
--- a/lib/foreman_rh_cloud.rb
+++ b/lib/foreman_rh_cloud.rb
@@ -3,6 +3,8 @@ require 'cgi'
 require 'uri'
 
 module ForemanRhCloud
+  CONSOLE_INSIGHTS_URL = "https://cert-api.access.redhat.com".freeze
+
   def self.on_premise_url
     return unless ForemanRhCloud.with_local_advisor_engine?
     port = ENV['ADVISOR_ENGINE_PORT'] || "8000"
@@ -23,7 +25,7 @@ module ForemanRhCloud
   end
 
   def self.legacy_insights_url
-    @legacy_insights_url ||= env_or_on_premise_url('SATELLITE_LEGACY_INSIGHTS_URL') || 'https://cert-api.access.redhat.com'
+    @legacy_insights_url ||= env_or_on_premise_url('SATELLITE_LEGACY_INSIGHTS_URL') || CONSOLE_INSIGHTS_URL
   end
 
   def self.verify_ssl_method
@@ -124,7 +126,7 @@ module ForemanRhCloud
   end
 
   def self.legacy_insights_ca
-    "#{ForemanRhCloud::Engine.root}/config/rh_cert-api_chain.pem"
+    ENV['SATELLITE_LEGACY_INSIGHTS_CA_PATH'] || "#{ForemanRhCloud::Engine.root}/config/rh_cert-api_chain.pem"
   end
 
```

parthaa: I like 2 better. Any opinions @evgeni @jeremylenz ?

evgeni: The `CONSOLE_INSIGHTS_URL` constant is unused in 2?

lib/foreman_rh_cloud.rb
parthaa: hmm  both seem to give me the same response.

```bash
$ curl  https://buku.paji.example.com:8444//r/insights/v1/static/release/insights-core.egg
curl: (35) error:0A0000C6:SSL routines::packet length too long
[vagrant@buku ~]$ curl  https://localhost:8444//r/insights/v1/static/release/insights-core.egg
curl: (35) error:0A0000C6:SSL routines::packet length too long
```

evgeni: is your engine running with ssl enabled right now? (it only does so when you give it certs)

parthaa: while 
```bash
curl  http://localhost:8444/r/insights/v1/static/release/insights-core.egg > /dev/null
 % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100 1304k  100 1304k    0     0  42.4M      0 --:--:-- --:--:-- --:--:-- 41.0M
```
seems to work 

I think we need the mtls patch for the engine

lib/foreman_rh_cloud.rb
parthaa: used here
```patch
+    @legacy_insights_url ||= env_or_on_premise_url('SATELLITE_LEGACY_INSIGHTS_URL') || CONSOLE_INSIGHTS_URL

```

lib/foreman_rh_cloud.rb
evgeni: no, you need to boot your engine with certs provided, then it will only work with https :)

jeremylenz:
@parthaa are you taking this over? Or otherwise I need to know what to do with it..

lib/foreman_rh_cloud.rb
parthaa: Lets do this https change in another PR. I need more time to test with the installer. I think a separate PR is good idea. We'll leave it as http for now

parthaa (approved):
APJ works well

chris1984 (approved):
Acking based on parthas testing. Code looks good. I agree with Partha we should do https in a separate pr

jeremylenz:
@parthaa do you understand these test failures? I don't..

chris1984:
> @parthaa do you understand these test failures? I don't..

They look transient, kicked it off again. Sometimes you have to do it a few times with this repo. One day we should look at that.

chris1984:
@jeremylenz all 🍏

lib/foreman_rh_cloud.rb
jeremylenz: @parthaa Is this resolved? should I just merge and you can fix whatever this is?

parthaa: Hmm not resolved yet but we can add this as a part of https://issues.redhat.com/browse/SAT-30645

parthaa: The way you have it works for our setup. We can look in to Https in a the other issue.

PR #937, opened 2025-01-07
SAT-30227 - Set auto remediations sync setting as true.

chris1984:


lfu (approved):
LGTM :+1:

chris1984:
Got approval from both Rich Jerrido and Tiho that we can move forward with this.

PR #940, opened 2025-01-10
only return response from Exceptions that have it

evgeni:
Only RestClient::ExceptionWithResponse has .response, while a "normal"
RestClient::Exception does not.
But by catching RestClient::Exception we might swallow errors that would
be useful for users in the logs, like
RestClient::SSLCertificateNotVerified

### DIFF
--- a/app/services/foreman_rh_cloud/cloud_request_forwarder.rb
+++ b/app/services/foreman_rh_cloud/cloud_request_forwarder.rb
@@ -17,7 +17,7 @@ module ForemanRhCloud
       logger.debug("Sending request to: #{request_opts[:url]}")
 
       execute_cloud_request(request_opts)
-    rescue RestClient::Exception => error_response
+    rescue RestClient::ExceptionWithResponse => error_response
       error_response.response
     end
 
--- a/app/controllers/insights_cloud/api/machine_telemetries_controller.rb
+++ b/app/controllers/insights_cloud/api/machine_telemetries_controller.rb
@@ -16,7 +16,15 @@ module InsightsCloud::Api
     # The method that "proxies" requests over to Cloud
     def forward_request
       certs = candlepin_id_cert @organization
-      @cloud_response = ::ForemanRhCloud::CloudRequestForwarder.new.forward_request(request, controller_name, @branch_id, certs)
+      begin
+        @cloud_response = ::ForemanRhCloud::CloudRequestForwarder.new.forward_request(request, controller_name, @branch_id, certs)
+      rescue RestClient::Exception => e
+        logger.info("Forwarding request failed with exception: #{e}")
+        return render json: { error: e }, status: :bad_gateway
+      rescue RestClient::Timeout => e
+        logger.info("Forwarding request failed with timeout: #{e}")
+        return render json: { error: e }, status: :gateway_timeout
+      end
 
       if @cloud_response.code == 401
         return render json: {


evgeni:
compare to e.g. https://github.com/theforeman/foreman_rh_cloud/blob/d642c75dc2b2701ba9ab412e09dfefd9778f2c0e/lib/foreman_inventory_upload/async/remove_insights_hosts_job.rb#L45-L56 where this is done correctly

app/services/foreman_rh_cloud/cloud_request_forwarder.rb
```diff
@@ -17,7 +17,7 @@ def forward_request(original_request, controller_name, branch_id, certs)
       logger.debug("Sending request to: #{request_opts[:url]}")
 
       execute_cloud_request(request_opts)
-    rescue RestClient::Exception => error_response
+    rescue RestClient::ExceptionWithResponse => error_response
```
ShimShtein, line 20: Sounds good, but we need to take into account the proper reporting of errors without a response.

I suppose we will need to modify the reporting block: https://github.com/theforeman/foreman_rh_cloud/blob/d642c75dc2b2701ba9ab412e09dfefd9778f2c0e/app/controllers/insights_cloud/api/machine_telemetries_controller.rb#L28

evgeni: Is that so?

That block is the whole reason I came here, as when you get a `RestClient::SSLCertificateNotVerified`, that block still calls `.code` but on `nil` (as `error_response.response` that we return here is `nil`)

After the patch, it correctly raises the exception and makes it "someone elses problem" ;)

ShimShtein: It does, but the error becomes 500 which is not correct.

We need to relay the original failure to the client. That's why this block exists. 
The idea is that the method always returns a result and the forwarder relays the response to the client.

I think @chris1984 is working on this already.

evgeni: Oooh you mean catching that exception and reporting it to the user?
yeah, that'd improve the UX, sure.

(today, you too get a 500 error, for the undefined `.code` on `nil`, but with the patch at least the log is useful, which it wasn't before)

chris1984:
@evgeni does this fix this issue https://issues.redhat.com/browse/SAT-23465 as well?

evgeni:
> @evgeni does this fix this issue https://issues.redhat.com/browse/SAT-23465 as well?

No? But it will at least now log the correct error in the logs, which it before did not

chris1984 (approved):
@ShimShtein are your comments addressed?

ShimShtein:
While we are at it, let's do it properly:
Can you please wrap the calling [code](https://github.com/theforeman/foreman_rh_cloud/blob/54a97c4541c207fafaae7aa293d721324bc696ce/app/controllers/insights_cloud/api/machine_telemetries_controller.rb#L19) in a catch block and report the correct error:

``` ruby
      @cloud_response = ::ForemanRhCloud::CloudRequestForwarder.new.forward_request(request, controller_name, @branch_id, certs)
      rescue RestClient::Exception => general_exception
        logger.info("Forwarding request failed with exception: #{general_exception}")
        render json: {error: general_exception}, status: :bad_gateway
      end
      rescue RestClient::Timeout => timeout_exception
        logger.info("Forwarding request failed with timeout: #{timeout_exception}")
        render json: {error: timeout_exception}, status: :gateway_timeout
      end
```

How does this sound?

Edit:
Another option would be catching those exceptions in the forwarder and returning a dummy response object with the `code` set to 502 or 504 respectively and the message as the response, so the `if` statement down the line will catch it and generate a response for it.

evgeni:
I don't like faking response objects when there is none, so I went with the rescue case :)

PR #941, opened 2025-01-14
SAT-30393 - Skip recurring logic tasks if on prem is enabled

chris1984:
This PR adds a skip to inventory/insights/generate report scheduled sync tasks if on_prem is enabled. This includes a change @jeremylenz has in his pr, once that is merged, I will remove it from here.

### DIFF
--- a/lib/foreman_inventory_upload/async/generate_all_reports_job.rb
+++ b/lib/foreman_inventory_upload/async/generate_all_reports_job.rb
@@ -13,20 +13,28 @@ module ForemanInventoryUpload
           return
         end
 
-        after_delay do
-          organizations = Organization.unscoped.all
+        if ForemanRhCloud.with_local_advisor_engine?
+          plan_self # so that 'run' runs
+        else
+          after_delay do
+            organizations = Organization.unscoped.all
 
-          organizations.map do |organization|
-            total_hosts = ForemanInventoryUpload::Generators::Queries.for_org(organization.id, use_batches: false).count
+            organizations.map do |organization|
+              total_hosts = ForemanInventoryUpload::Generators::Queries.for_org(organization.id, use_batches: false).count
 
-            if total_hosts <= ForemanInventoryUpload.max_org_size
-              disconnected = false
-              plan_generate_report(ForemanInventoryUpload.generated_reports_folder, organization, disconnected)
-            else
-              logger.info("Skipping automatic uploads for organization #{organization.name}, too many hosts (#{total_hosts}/#{ForemanInventoryUpload.max_org_size})")
-            end
-          end.compact
+              if total_hosts <= ForemanInventoryUpload.max_org_size
+                disconnected = false
+                plan_generate_report(ForemanInventoryUpload.generated_reports_folder, organization, disconnected)
+              else
+                logger.info("Skipping automatic uploads for organization #{organization.name}, too many hosts (#{total_hosts}/#{ForemanInventoryUpload.max_org_size})")
+              end
+            end.compact
+          end
         end
+      end
+
+      def run
+        output[:status] = _('The scheduled process is disabled because this Foreman is configured with the use_local_advisor_engine option.') if ForemanRhCloud.with_local_advisor_engine?
       end
 
       def rescue_strategy_for_self
--- a/lib/insights_cloud/async/insights_scheduled_sync.rb
+++ b/lib/insights_cloud/async/insights_scheduled_sync.rb
@@ -13,9 +13,17 @@ module InsightsCloud
           return
         end
 
-        after_delay do
-          plan_full_sync
+        if ForemanRhCloud.with_local_advisor_engine?
+          plan_self
+        else
+          after_delay do
+            plan_full_sync # so that 'run' runs
+          end
         end
+      end
+
+      def run
+        output[:status] = _('The scheduled process is disabled because this Foreman is configured with the use_local_advisor_engine option.') if ForemanRhCloud.with_local_advisor_engine?
       end
 
       def plan_full_sync
--- a/test/jobs/inventory_scheduled_sync_test.rb
+++ b/test/jobs/inventory_scheduled_sync_test.rb
@@ -14,6 +14,16 @@ class InventoryScheduledSyncTest < ActiveSupport::TestCase
     ForemanTasks.sync_task(InventorySync::Async::InventoryScheduledSync)
   end
 
+  test 'Skips execution if with_local_advisor_engine? is true' do
+    ForemanRhCloud.stubs(:with_local_advisor_engine?).returns(true)
+
+    InventorySync::Async::InventoryScheduledSync.any_instance.expects(:plan_org_sync).never
+
+    task = ForemanTasks.sync_task(InventorySync::Async::InventoryScheduledSync)
+    status = task.output[:status].to_s
+    assert_match(/Foreman is configured with the use_local_advisor_engine option/, status)
+  end
+
   test 'Skips execution if auto upload is disabled' do
     Setting[:allow_auto_inventory_upload] = false
 
--- a/lib/inventory_sync/async/inventory_scheduled_sync.rb
+++ b/lib/inventory_sync/async/inventory_scheduled_sync.rb
@@ -13,13 +13,17 @@ module InventorySync
           return
         end
 
-        after_delay do
-          # perform a sequence of sync then delete in parallel for all organizations
-          concurrence do
-            Organization.unscoped.each do |org|
-              sequence do
-                plan_org_sync(org)
-                plan_remove_insights_hosts(org.id) if Setting[:allow_auto_insights_mismatch_delete]
+        if ForemanRhCloud.with_local_advisor_engine?
+          plan_self # so that 'run' runs
+        else
+          after_delay do
+            # perform a sequence of sync then delete in parallel for all organizations
+            concurrence do
+              Organization.unscoped.each do |org|
+                sequence do
+                  plan_org_sync(org)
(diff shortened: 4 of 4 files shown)


test/jobs/inventory_scheduled_sync_test.rb
```diff
@@ -5,6 +5,7 @@ class InventoryScheduledSyncTest < ActiveSupport::TestCase
   include ForemanTasks::TestHelpers::WithInThreadExecutor
 
   test 'Schedules an execution if auto upload is enabled' do
+    Setting[:use_local_advisor_engine] = false
```
jeremylenz: `use_local_advisor_engine` is a SETTING (set by installer options and/or settings.yaml), not a Setting (set by users via api/UI/Hammer). All of these `Setting[]` calls should not be needed. Instead the tests should probably mock `ForemanRhCloud.with_local_advisor_engine?` and make it return as needed.

test/jobs/inventory_scheduled_sync_test.rb
```diff
@@ -31,4 +31,13 @@ class InventoryScheduledSyncTest < ActiveSupport::TestCase
 
     ForemanTasks.sync_task(InventorySync::Async::InventoryScheduledSync)
   end
+
+  test 'Skips execution if local advisor engine is enabled' do
+    Setting[:allow_auto_inventory_upload] = false
+    ForemanRhCloud.with_local_advisor_engine?.returns(true)
```
jeremylenz:
```suggestion
    ForemanRhCloud.stubs(:with_local_advisor_engine?).returns(true)
```

chris1984: 🤦🏻

lib/foreman_inventory_upload/async/generate_all_reports_job.rb
```diff
@@ -5,6 +5,13 @@ class GenerateAllReportsJob < ::Actions::EntryAction
       include ForemanInventoryUpload::Async::DelayedStart
 
       def plan
+        if ForemanRhCloud.with_local_advisor_engine?
+          logger.debug(
+            'The scheduled process is disabled due to the "with_local_advisor_engine"
```
jeremylenz:
```suggestion
            'The scheduled process is disabled due to the "use_local_advisor_engine"
```

jeremylenz (commented):
I notice this approach just skips planning the scheduled sync tasks if we're using the local advisor engine. This should work fine, but you'd still see actions in Dynflow that are empty and successful.

Another option is to avoid scheduling them at all here --> https://github.com/theforeman/foreman_rh_cloud/blob/dc80828e3949cd9967121f0a251639561ec2124a/lib/foreman_rh_cloud/engine.rb#L168

But that may be a bit less safe because of where and when `ForemanRhCloud.with_local_advisor_engine?` is defined..? I'm a bit unclear on if that will work, but if it does it seems cleaner to me.

@ShimShtein @parthaa thoughts?

parthaa:
> I notice this approach just skips planning the scheduled sync tasks if we're using the local advisor engine. This should work fine, but you'd still see actions in Dynflow that are empty and successful.

While I agree you'd rather not have tasks scheduled if not needed. I am thinking in the future when we may want the hybrid option. I have a feeling we are better of scheduling but skipping.

parthaa:
Hmm how about a seed file that has this logic so that we `disable` the recurring task rather than just returning.

```ruby
if ForemanRhCloud.with_local_advisor_engine?
  ::ForemanTasks::RecurringLogic.joins(:tasks).where(tasks: { label: ["InventorySync::Async::InventoryScheduledSync", "ForemanInventoryUpload::Async::GenerateAllReportsJob", "InsightsCloud::Async::InsightsScheduledSync"] }, state: "active").distinct(:id).each {|logic| logic.update!(enabled: false)}
end
```

jeremylenz:
Do we support turning local off after it's on? if so, we may want to do it both ways (e.g. enable or disable the logic.)

chris1984:
> Do we support turning local off after it's on? if so, we may want to do it both ways (e.g. enable or disable the logic.)

We do support that with the installer flag setting it to false, we can't do this in a seed file then. We would have to do something in the engine rb right @parthaa

jeremylenz:
Side question: Why disable the insights scheduled sync one? Wouldn't that just go to the local iop-advisor-engine?

chris1984:
> Side question: Why disable the insights scheduled sync one? Wouldn't that just go to the local iop-advisor-engine?

Fair point, I was just doing it since the engine writes to the api and we modify the DB right then. There is no harm in having it sync to the engine. Will take that out of the pr.

jeremylenz:
> the engine writes to the api and we modify the DB right then. There is no harm in having it sync to the engine.

If it's also not doing anything useful, that'd be an argument for disabling it as well.

lib/foreman_inventory_upload/async/generate_all_reports_job.rb
```diff
@@ -5,6 +5,13 @@ class GenerateAllReportsJob < ::Actions::EntryAction
       include ForemanInventoryUpload::Async::DelayedStart
 
       def plan
+        if ForemanRhCloud.with_local_advisor_engine?
+          logger.debug(
+            'The scheduled process is disabled due to the "use_local_advisor_engine"
+            SETTING being set to true.'
```
jeremylenz: Instead of a debug log, may be more useful to put this in the task's output. like 
```
output[:status] = _('The scheduled process is disabled because this Foreman is configured with the use_local_advisor_engine option.')
```

jeremylenz: also please note my suggested wording ;)

parthaa (approved):
APJ. I thought about disabling tasks when advisor is on but feel its not worth the complexity and having to deal with enabling tasks when user decides to turn the advisor off.

lib/insights_cloud/async/insights_scheduled_sync.rb
```diff
@@ -13,9 +13,17 @@ def plan
...
+          after_delay do
+            plan_full_sync
+          end
         end
+        # rubocop:enable Style/GuardClause
+      end
+
+      def run
+        output[:status] = _('The scheduled process is disabled because this Foreman is configured with the use_local_advisor_engine option.') if ForemanRhCloud.with_local_advisor_engine?
       end
```
adamruzicka: As far as I remember, unless you call `plan_self` in `plan`, the `run` will never be executed, so why have it here? Or am I misremembering?

jeremylenz: I'm pretty sure you're not misremembering ;)

@chris1984 we can verify this during testing, but I'm pretty sure `plan_self` will be needed in `plan`, from what I remember.

chris1984:
Working correctly now 
screenshot-ip-10-0-167-73 rhos-01 prod psi rdu2 redhat com-2025 01 22-10_27_46

lib/insights_cloud/async/insights_scheduled_sync.rb
```diff
@@ -13,9 +13,16 @@ def plan
...
         end
 
-        after_delay do
-          plan_full_sync
+        unless ForemanRhCloud.with_local_advisor_engine?
+          after_delay do
+            plan_full_sync
+          end
         end
+        plan_self
```
parthaa: you dont want _plan_self_ to run if it is not _with_local_advisor_engine?_

chris1984: Fixed to be in an if/else

lib/foreman_inventory_upload/async/generate_all_reports_job.rb
```diff
@@ -13,22 +13,30 @@ def plan
...
-
-            if total_hosts <= ForemanInventoryUpload.max_org_size
-              disconnected = false
-              plan_generate_report(ForemanInventoryUpload.generated_reports_folder, organization, disconnected)
-            else
-              logger.info("Skipping automatic uploads for organization #{organization.name}, too many hosts (#{total_hosts}/#{ForemanInventoryUpload.max_org_size})")
-            end
-          end.compact
+        if ForemanRhCloud.with_local_advisor_engine?
+          plan_self
```
jeremylenz:
```suggestion
          plan_self # so that 'run' runs
```

chris1984: Added, will add it to the other tasks

parthaa:
Works well. ACK

jeremylenz (approved):

parthaa:
Recommended patch

```patch
diff --git a/test/jobs/inventory_scheduled_sync_test.rb b/test/jobs/inventory_scheduled_sync_test.rb
index 338d900..48a2f88 100644
--- a/test/jobs/inventory_scheduled_sync_test.rb
+++ b/test/jobs/inventory_scheduled_sync_test.rb
@@ -15,13 +15,12 @@ class InventoryScheduledSyncTest < ActiveSupport::TestCase
   end
 
   test 'Skips execution if with_local_advisor_engine? is true' do
-    Setting[:allow_auto_inventory_upload] = true
-
     ForemanRhCloud.stubs(:with_local_advisor_engine?).returns(true)
-
     InventorySync::Async::InventoryScheduledSync.any_instance.expects(:plan_org_sync).never
 
-    ForemanTasks.sync_task(InventorySync::Async::InventoryScheduledSync)
+    task = ForemanTasks.sync_task(InventorySync::Async::InventoryScheduledSync)
+    status = task.output[:status].to_s
+    assert_match(/Foreman is configured with the use_local_advisor_engine option/, status)
   end
 
   test 'Skips execution if auto upload is disabled' do
```

PR #943, opened 2025-01-27
update URL of local advisor engine as it is deployed by the installer

evgeni:
- **don't set legacy insights ca when local advisor engine is used**
- **update URL of local advisor engine as it is deployed by the installer**

### DIFF
--- a/lib/foreman_rh_cloud.rb
+++ b/lib/foreman_rh_cloud.rb
@@ -5,8 +5,8 @@ require 'uri'
 module ForemanRhCloud
   def self.on_premise_url
     return unless ForemanRhCloud.with_local_advisor_engine?
-    port = ENV['ADVISOR_ENGINE_PORT'] || "8000"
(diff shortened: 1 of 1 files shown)


ekohl (commented):
I'm trying to figure out what's going on in `lib/foreman_rh_cloud.rb` and I think it's extremely complex. It's sort of reinventing the settings system we already have in place. Also, we have a generic proxy system and it doesn't appear to be reusing that.

For example:
https://github.com/theforeman/foreman_rh_cloud/blob/50e24b6ac6320a82e69ddca12c59cdc8db781611/lib/foreman_rh_cloud.rb#L65-L82

In https://github.com/Katello/katello/commit/d593923a18d445f853d61d8af9d9a86e6195c82c the last traces of this were removed. https://github.com/theforeman/foreman_rh_cloud/pull/944 gets rid of that at least.

chris1984 (approved):
Looks fine to me, @parthaa can you have a look as well?

parthaa (approved):
LGTM

PR #944, opened 2025-01-27
Drop cdn_proxy settings

ekohl:
Back in 2020 the last traces of this were removed from Katello so this can't have any affect anymore.

It may be better to look at some other proxy config, but I can't really judge that right now.

Link: https://github.com/Katello/katello/commit/d593923a18d445f853d61d8af9d9a86e6195c82c


chris1984 (approved):
Looks good to me, thanks for linking the Katello commit for visibility.

PR #945, opened 2025-01-29
Move Insights to top-level nav

jeremylenz:
Move the "Insights" nav item to the top level, after Hosts.

image


ehelms (approved):

chris1984 (approved):
Looks good as a top level, and works fine.

PR #947, opened 2025-02-04
SAT-27221 - Separate inventory upload for host into a new param

chris1984:
* Added a new param in Foreman called `host_registraton_insights_inventory`
* Defined that param `host_registraton_insights_inventory` and switched the generator to look for that for a host instead of `enable_client_param`
* Updated a test that looked for that param
* Added a new test for the new param

Testing results:
With host_registraton_insights_inventory off
```bash
[vagrant@ip-10-0-168-88 done]$ tar -xvf report_for_1.tar.xz 
./
./metadata.json
```

With host_registraton_insights_inventory on
```bash
[vagrant@ip-10-0-168-88 done]$ tar -xvf report_for_1.tar.xz 
./
./metadata.json
./a70f324e-2497-40f4-8a41-f48d7861bf86.json
```

### DIFF
--- a/db/seeds.d/189_add_host_inventory_param.rb
+++ b/db/seeds.d/189_add_host_inventory_param.rb
@@ -1,0 +1,7 @@
+CommonParameter.without_auditing do
+  params = [
+    { name: "host_registration_insights_inventory", key_type: "boolean", value: true },
+  ]
+
+  params.each { |param| CommonParameter.find_or_create_by(param) }
+end
--- a/test/unit/slice_generator_test.rb
+++ b/test/unit/slice_generator_test.rb
@@ -453,9 +453,9 @@ class SliceGeneratorTest < ActiveSupport::TestCase
     assert_equal 1, generator.hosts_count
   end
 
-  test 'excludes hosts with host_registration_insights set to false' do
+  test 'excludes hosts with host_registration_insights_inventory set to false' do
     @host.host_parameters << HostParameter.create(
-      name: 'host_registration_insights',
+      name: 'host_registration_insights_inventory',
       value: "false",
       parameter_type: 'boolean'
     )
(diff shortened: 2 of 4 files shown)


db/seeds.d/189_add_host_inventory_param.rb
```diff
@@ -0,0 +1,7 @@
+CommonParameter.without_auditing do
+  params = [
+    { name: "host_registraton_insights_inventory", key_type: "boolean", value: true },
```
ianballou:
```suggestion
    { name: "host_registration_insights_inventory", key_type: "boolean", value: true },
```

test/unit/slice_generator_test.rb
```diff
@@ -455,7 +455,7 @@ def fact_names
 
   test 'excludes hosts with host_registration_insights set to false' do
```
ianballou:
```suggestion
  test 'excludes hosts with host_registration_insights_inventory set to false' do
```

db/seeds.d/189_add_host_inventory_param.rb
ianballou: Other references too.

db/seeds.d/189_add_host_inventory_param.rb
```diff
@@ -0,0 +1,7 @@
+CommonParameter.without_auditing do
```
ianballou, line 1: Is there any problem with this param living in foreman_rh_cloud but `host_registration_insights` living in foreman? I don't fully get the history or connection.

ianballou:
@chris1984 how do I get `report_for_1.tar.xz`  ?

db/seeds.d/189_add_host_inventory_param.rb
ShimShtein: That's OK. The `host_registration_isnights` is used by the registration process to install `insights-client` and we were using it to indicate the need to upload to the cloud. Now that we split the roles, one can live in core and the other here.

chris1984:
> @chris1984 how do I get `report_for_1.tar.xz` ?

On a devel box they should be here:
```bash
[vagrant@ip-10-0-168-227 done]$ pwd
/home/vagrant/foreman/red_hat_inventory/uploads/done
[vagrant@ip-10-0-168-227 done]$ ls
report_for_1.tar.xz
```

ianballou (commented):
Tested, working great! Just the naming changes then :+1:

chris1984:
@ianballou updated

ianballou (approved):
Looks good!

PR #948, opened 2025-02-10
i18n - pulling from tx

ekohl:
Also syncs in the locale Makefile from the plugin template to resolve some bugs.


chris1984 (approved):
Nice catch on the `minimum_perc`

PR #949, opened 2025-02-11
Use trusted publishers to release gems

ekohl:
Link: https://community.theforeman.org/t/use-trusted-publishers-to-release-gems/42009

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,7 +1,7 @@
 ---
 name: Release
 
-on:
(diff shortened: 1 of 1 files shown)


chris1984 (approved):
I read the RFC you wrote, and it makes sense and long term it will be nice to have release workflows locked down more for security.

ekohl:
* Created the trusted publisher on Rubygems
* Deleted old GH secret
* Deleted old API key

PR #951, opened 2025-02-17
Hide more things when using local advisor engine

jeremylenz:
When using local advisor engine,

1. Do not define routes in web UI if they depend on being connected
2. Pass `disconnected=true` to Dynflow tasks
3. make `rake rh_cloud_insights:announce_to_sources` a no-op

### DIFF
--- a/lib/tasks/insights.rake
+++ b/lib/tasks/insights.rake
@@ -24,13 +24,16 @@ namespace :rh_cloud_insights do
   desc "Re-announce all organizations into Sources on RH cloud."
   task announce_to_sources: [:environment] do
     logger = Logging::Logger.new(STDOUT)
-    Organization.unscoped.each do |org|
-      presence = ForemanRhCloud::CloudPresence.new(org, logger)
-      presence.announce_to_sources
-    rescue StandardError => ex
-      logger.warn(ex)
+    if ForemanRhCloud.with_local_advisor_engine?
+      logger.warn('Task announce_to_sources is not available when using local advisor engine')
+    else
+      Organization.unscoped.each do |org|
+        presence = ForemanRhCloud::CloudPresence.new(org, logger)
+        presence.announce_to_sources
+      rescue StandardError => ex
+        logger.warn(ex)
+      end
+      logger.info('Reannounced all organizations')
     end
-
-    logger.info('Reannounced all organizations')
   end
 end
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -9,7 +9,9 @@ Rails.application.routes.draw do
     get 'settings', to: 'uploads_settings#index'
     post 'setting', to: 'uploads_settings#set_advanced_setting'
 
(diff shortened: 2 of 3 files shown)


lib/tasks/insights.rake
```diff
@@ -24,13 +24,19 @@ namespace :rh_cloud_insights do
   desc "Re-announce all organizations into Sources on RH cloud."
   task announce_to_sources: [:environment] do
     logger = Logging::Logger.new(STDOUT)
+    local_advisor = ForemanRhCloud.with_local_advisor_engine?
     Organization.unscoped.each do |org|
+      next if local_advisor
```
parthaa: Is it better to add this line before `Organization.unscoped.each` we don't need to do the each right

jeremylenz: Enclosed in a conditional instead.

parthaa:
Works well. Couple more spots to may be hide
```patch
diff --git a/config/routes.rb b/config/routes.rb
index 9683e9e..7b3a2b0 100644
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -9,7 +9,7 @@ Rails.application.routes.draw do
     get 'settings', to: 'uploads_settings#index'
     post 'setting', to: 'uploads_settings#set_advanced_setting'
 
-    post 'cloud_connector', to: 'uploads#enable_cloud_connector'
+    post 'cloud_connector', to: 'uploads#enable_cloud_connector' unless ForemanRhCloud.with_local_advisor_engine?
 
     resources :tasks, only: [:create, :show]
 
@@ -63,8 +63,10 @@ Rails.application.routes.draw do
       end
 
       namespace 'rh_cloud' do
-        post 'enable_connector', to: 'inventory#enable_cloud_connector'
-        post 'cloud_request', to: 'cloud_request#update'
+        unless ForemanRhCloud.with_local_advisor_engine?
+          post 'enable_connector', to: 'inventory#enable_cloud_connector'
+          post 'cloud_request', to: 'cloud_request#update'
+        end
         get 'advisor_engine_config', to: 'advisor_engine_config#show'
       end
 
index aa20c89..082bff2 100644
@@ -9,10 +9,10 @@ namespace :rh_cloud_inventory do
       else
         organizations = [Organization.where(:id => ENV['organization_id']).first]
       end
-
+      disconnected = ForemanRhCloud.with_local_advisor_engine?
       User.as_anonymous_admin do
         organizations.each do |organization|
-          ForemanTasks.async_task(ForemanInventoryUpload::Async::GenerateReportJob, ForemanInventoryUpload.generated_reports_folder, organization.id, false)
+          ForemanTasks.async_task(ForemanInventoryUpload::Async::GenerateReportJob, ForemanInventoryUpload.generated_reports_folder, organization.id, disconnected)
           puts "Generated and uploaded inventory report for organization '#{organization.name}'"
         end
       end
```

parthaa (approved):
Works well. Merge away once the changes are in :)

PR #952, opened 2025-02-19
Load all rules from advisor engine

parthaa:
The idea here is to load all the rules into the db in one go from insights advisor if the some rules are not in the database.

This will make the hits api more performant since it is not updating the db for the most expensive operations, i.e upserting rules and resolutions

To test this

1. Setup foreman with  advisor engine
2. Check out this pr
3. `rake console` - find the `InsightsRule.maximum(:id)` and `InsightsResolution.maximum(:id)` values. they may be nil if you dont have any rules .
4. register a host and introduce a vulnerability like `chmod 777 /etc/shadow`
5. run `insights-client` on the host
6. `rake console` - find the `InsightsRule.maximum(:id)` and `InsightsResolution.maximum(:id)` values. You should see around 900+ rules added in the db. Make a note of the maximum id
7. run `insights-client` on the host again
8. `rake console` - find the `InsightsRule.maximum(:id)` and `InsightsResolution.maximum(:id)` values. These values should not have changed.

### DIFF
--- a/app/services/foreman_rh_cloud/hits_uploader.rb
+++ b/app/services/foreman_rh_cloud/hits_uploader.rb
@@ -35,14 +35,9 @@ module ForemanRhCloud
 
     def update_rules_and_resolutions
       return if @payload[:rules].blank?
-      # rubocop:disable Rails/SkipsModelValidations
-      ::InsightsRule.upsert_all(@payload[:rules], unique_by: :rule_id)
-      rules = @payload[:rules].map { |rule| rule[:rule_id] }
-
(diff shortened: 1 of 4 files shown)


pablomh:
After applying this patch we (SatPerf) haven't seen the error mentioned in https://github.com/RedHatInsights/iop-advisor-engine/pull/5#discussion_r1956594406 anymore.

jeremylenz (approved):
ACK based on Pablo's testing --> https://github.com/theforeman/foreman_rh_cloud/pull/952#issuecomment-2678909347

PR #963, opened 2025-03-12
Rename RHCloud tab in Settings to Insights

jeremylenz:
I just changed the label, not the category name, since I think changing the category name might put existing setting values at risk.


ianballou (approved):
Looks good to me! I tested settings with and without the change and they persisted.

PR #971, opened 2025-04-03
Explicitly include CA file when making calls to advisor engine

ehelms:
When using custom certificates, calls to the advisor engine will fail. 

The "server CA cert" is trusted at the system level by default. When non-custom certificates are installed, the server-ca is equal to the default CA and thus the system ends up trusting the CA certificate and everything works. Once custom certificates are involved, the default CA is not trusted by the system and failures occur.

We should not rely on the host to trust the CAs we need to use when making calls, and should rely on explicit configuration.

### DIFF
--- a/app/services/foreman_rh_cloud/cloud_request_forwarder.rb
+++ b/app/services/foreman_rh_cloud/cloud_request_forwarder.rb
@@ -34,7 +34,13 @@ module ForemanRhCloud
         ),
       }
       requested_url = original_request.original_fullpath.end_with?('/') ? original_request.path + '/' : original_request.path
-      base_params.merge(path_params(requested_url, certs))
+      params = path_params(requested_url, certs)
+
+      if ForemanRhCloud.with_local_advisor_engine?
+        params[:ssl_ca_file] = ForemanRhCloud.ca_cert
+      end
+
+      base_params.merge(params)
     end
 
     def prepare_forward_payload(original_request, controller_name)
--- a/lib/foreman_rh_cloud/engine.rb
+++ b/lib/foreman_rh_cloud/engine.rb
@@ -212,4 +212,15 @@ module ForemanRhCloud
   def self.with_local_advisor_engine?
     SETTINGS.dig(:foreman_rh_cloud, :use_local_advisor_engine) || false
   end
+
+  def self.ca_cert
+    # The reference to candlepin ca_cert_file can be removed
+    # once the setting is dropped. Foreman 3.15 introduces
+    # a single CA file that bundles all CAs.
+    if ::SETTINGS.dig(:katello, :candlepin, :ca_cert_file)
+      ::SETTINGS[:katello][:candlepin][:ca_cert_file]
+    else
+      ::SETTINGS[:ssl_ca_file]
+    end
+  end
 end
--- a/app/services/foreman_rh_cloud/cloud_request.rb
+++ b/app/services/foreman_rh_cloud/cloud_request.rb
@@ -8,6 +8,10 @@ module ForemanRhCloud
         proxy: ForemanRhCloud.transformed_http_proxy_string,
       }.deep_merge(params)
 
+      if ForemanRhCloud.with_local_advisor_engine?
+        final_params[:ssl_ca_file] ||= ForemanRhCloud.ca_cert
+      end
+
       response = RestClient::Request.execute(final_params)
 
       logger.debug("Response headers for request url #{final_params[:url]} are: #{response.headers}")
--- a/app/services/foreman_rh_cloud/rules_ingester.rb
+++ b/app/services/foreman_rh_cloud/rules_ingester.rb
@@ -13,7 +13,18 @@ module ForemanRhCloud
 
     def fetch_rules_data
       advisor_url = "#{ForemanRhCloud.on_premise_url}/r/insights/v1/static/release/content.json"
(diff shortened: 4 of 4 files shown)


lib/foreman_rh_cloud/engine.rb
```diff
@@ -212,4 +212,15 @@ def self.register_scheduled_task(task_class, cronline)
...
+
+  def self.ca_cert
+    # The reference to candlepin ca_cert_file can be removed
+    # once the setting is dropped. Foreman 3.15 introduces
+    # a single CA file that bundles all CAs.
+    if ::SETTINGS.dig(:katello, :candlepin, :ca_cert_file)
+      ::SETTINGS[:katello][:candlepin][:ca_cert_file]
+    else
+      ::SETTINGS[:ssl_ca_file]
+    end
```
ehelms, line 224: This `if` block is to future proof this chunk of code when we drop the Candlepin configuration settings in favor of the change to bundle CAs. Two references to this work:

https://github.com/theforeman/puppet-certs/pull/490
https://github.com/Katello/katello/pull/11327

app/services/foreman_rh_cloud/cloud_request_forwarder.rb
```diff
@@ -34,7 +34,13 @@ def prepare_request_opts(original_request, forward_payload, forward_params, cert
         ),
       }
       requested_url = original_request.original_fullpath.end_with?('/') ? original_request.path + '/' : original_request.path
-      base_params.merge(path_params(requested_url, certs))
+      path_params = path_params(requested_url, certs))
```
parthaa:
```suggestion
      path_params = path_params(requested_url, certs)
```

app/services/foreman_rh_cloud/cloud_request_forwarder.rb
```diff
@@ -34,7 +34,13 @@ def prepare_request_opts(original_request, forward_payload, forward_params, cert
         ),
       }
       requested_url = original_request.original_fullpath.end_with?('/') ? original_request.path + '/' : original_request.path
-      base_params.merge(path_params(requested_url, certs))
+      path_params = path_params(requested_url, certs)
```
parthaa: Can you call the `path_params` variable some other name to differentiate it from the method in the same object

parthaa (commented):
Testing it right now

app/services/foreman_rh_cloud/cloud_request_forwarder.rb
ehelms: Done

evgeni:
Something similar needs to happen in 
https://github.com/theforeman/foreman_rh_cloud/blob/6d649d871fdc6aa0e18e4f1b075f4c6a6b79320a/app/services/foreman_rh_cloud/rules_ingester.rb#L14-L17

parthaa:
Proposed patch for both hits and remediations.

```diff
diff --git a/app/services/foreman_rh_cloud/cloud_request.rb b/app/services/foreman_rh_cloud/cloud_request.rb
index eca78f0..e6da897 100644
--- a/app/services/foreman_rh_cloud/cloud_request.rb
+++ b/app/services/foreman_rh_cloud/cloud_request.rb
@@ -8,6 +8,10 @@ module ForemanRhCloud
         proxy: ForemanRhCloud.transformed_http_proxy_string,
       }.deep_merge(params)
 
+      if ForemanRhCloud.with_local_advisor_engine?
+        final_params[:ssl_ca_file] ||= ForemanRhCloud.ca_cert
+      end
+
       response = RestClient::Request.execute(final_params)
 
       logger.debug("Response headers for request url #{final_params[:url]} are: #{response.headers}")
diff --git a/app/services/foreman_rh_cloud/rules_ingester.rb b/app/services/foreman_rh_cloud/rules_ingester.rb
index 664b12f..7793b4c 100644
--- a/app/services/foreman_rh_cloud/rules_ingester.rb
+++ b/app/services/foreman_rh_cloud/rules_ingester.rb
@@ -13,7 +13,18 @@ module ForemanRhCloud
 
     def fetch_rules_data
       advisor_url = "#{ForemanRhCloud.on_premise_url}/r/insights/v1/static/release/content.json"
-      JSON.parse(Net::HTTP.get(URI.parse(advisor_url)), symbolize_names: true)
+      uri = URI.parse(advisor_url)
+      http = Net::HTTP.new(uri.host, uri.port)
+      http.use_ssl = true
+
+      # Set CA certificate
+      http.ca_file = ForemanRhCloud.ca_cert
+      http.verify_mode = OpenSSL::SSL::VERIFY_PEER
+
+      request = Net::HTTP::Get.new(uri.request_uri)
+
+      response = http.request(request)
+      JSON.parse(response.body, symbolize_names: true)
     end
 
     def fetch_rules_and_resolutions
```

lib/foreman_rh_cloud/engine.rb
evgeni: Given both PRs are merged, do we even need this if/else? Or are you trying to be compatible with pre 3.15 deployments here?

ehelms: Trying to be compatible. I'd rather merge this with compatibility and then follow up to remove it so there is a commit available with it.

ehelms:
@parthaa Can you take another look?

chris1984 (approved):
It looks fine to me from a code side of things but have not tested. @parthaa want to take another look?

PR #977, opened 2025-04-17
SAT-31227 - enable the usr-overlay before running remediations

ianballou:
Edits the beginning of remediation playbooks to enable the bootc usr-overlay.

Known issue: the Ansible package module incorrectly auto-detects the package manager to use: https://github.com/ansible/ansible/issues/85011

To work around this known issue, we'd have to edit every single package action in a playbook to use `dnf`. This sounds very risky.

Questions:
1. Are there security issues with editing playbooks? Are they signed in some way and editing them breaks a promise?
2. Could this instead trigger a prerequisite REX Job to enable the bootc usr-overlay if editing the playbook is not ideal?
3. Will there ever be more than one host in the hosts list for a playbook from Insights?
4. Alternatively, how can the target host be queried besides grabbing it from the playbook?

If this is too risky, we can discuss next introducing the work directly to Insights. In that case, we could potentially add a suggestion in the UI to enable the usr-overlay on the remediations page as an interim solution.


ianballou:
Re: Will there ever be more than one host in the hosts list for a playbook from Insights?
According to https://github.com/Katello/katello/pull/11374/files#r2048670891, these playbooks are generated one per grouping of hosts. So, the bootc check will need to be moved into native Ansible. We can likely use bootc status.

ianballou:
Edit: this PR will need to use native ansible to detect the image mode host like we do here https://github.com/Katello/katello/pull/11374

Also, it sounds like Insights playbooks are signed, so it's sounding possible that editing the playbooks is not safe. Also, if the playbooks use the package module, Ansible will fail due to https://github.com/ansible/ansible/issues/84803#issuecomment-2713184856

ianballou:
We might actually be able to stick two playbooks in one yaml file, which would mean we could run a 'pre' template before the insights one. I need to test this, but it would mean no editing of the insights playbook.

This still wouldn't fix the issue with the use of the atomic_container package module, however.

ianballou:
Never mind my last idea, you can only have multiple plays in one playbook, and I believe the entire playbook is signed by Insights.

The only idea I have left is to have some chained REX job that first enables the usr-overlay (not sure how possible this is without a lot of work) or just warn users in the UI that said hosts will require the usr-overlay enabled.

However, as mentioned above, enabling the usr-overlay will not fix the fact that Ansible cannot properly install packages on image mode machines without specifying the use of `dnf`.

ianballou:
Closing since this is not the right solution. The playbooks will fail regardless due to the Ansible bug.

PR #978, opened 2025-04-18
SAT-31457 - Move download button to generate tab instead of upload

chris1984:
Before:
Screenshot 2025-04-18 100841

After:
Screenshot 2025-04-18 100728

### DIFF
--- a/lib/foreman_inventory_upload.rb
+++ b/lib/foreman_inventory_upload.rb
@@ -30,6 +30,11 @@ module ForemanInventoryUpload
     File.join(ForemanInventoryUpload.done_folder, filename)
   end
 
+  def self.report_file_paths(organization_id)
+    filename = facts_archive_name(organization_id)
+    Dir[ForemanInventoryUpload.uploads_file_path(filename), ForemanInventoryUpload.done_file_path(filename)]
+  end
+
   def self.generated_reports_folder
     @generated_reports_folder ||= ensure_folder(
       File.join(
--- a/webpack/ForemanInventoryUpload/Components/TabHeader/TabHeader.js
+++ b/webpack/ForemanInventoryUpload/Components/TabHeader/TabHeader.js
@@ -8,7 +8,13 @@ import { selectSubscriptionConnectionEnabled } from '../InventorySettings/Invent
 import { isExitCodeLoading } from '../../ForemanInventoryHelpers';
 import './tabHeader.scss';
 
-const TabHeader = ({ exitCode, onRestart, onDownload, toggleFullScreen }) => {
+const TabHeader = ({
+  exitCode,
+  onRestart,
+  onDownload,
+  downloadButtonDisabled,
+  toggleFullScreen,
+}) => {
   const subscriptionConnectionEnabled = useSelector(
     selectSubscriptionConnectionEnabled
   );
@@ -33,7 +39,7 @@ const TabHeader = ({ exitCode, onRestart, onDownload, toggleFullScreen }) => {
             </Button>
           ) : null}
           {onDownload ? (
-            <Button onClick={onDownload}>
+            <Button onClick={onDownload} disabled={downloadButtonDisabled()}>
               {__('Download Report')} <Icon name="download" />
             </Button>
           ) : null}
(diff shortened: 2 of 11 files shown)


webpack/ForemanInventoryUpload/Components/TabHeader/TabHeader.js
```diff
@@ -22,7 +22,7 @@ const TabHeader = ({ exitCode, onRestart, onDownload, toggleFullScreen }) => (
             {__('Generate and upload report')}
           </Button>
         ) : null}
-        {onDownload ? (
+        {onRestart ? (
```
lfu: This did not work. Nothing is downloaded when clicking the button.

lfu: I think the `Download button` should gray out until the report is generated.

chris1984: > I think the `Download button` should gray out until the report is generated.

Can we even do that with this old React and Redux code?

lfu: @jeremylenz Is it doable?

jeremylenz: PF5 (and PF4) buttons have `isLoading` and `isDisabled` props. It's quite doable! I would recommend changing the import from `patternfly-react` to `@patternfly/react-core` for everything here.

jeremylenz: The only reason these aren't PF5 already is that the codemod we ran was only for PF4 > PF5, and this is PF3.

chris1984: Is there way to specifically use pf4 when doing the imports? Do you see the page breaking when we change the imports.

jeremylenz: Importing from `@patternfly/react-core` will use PF5. (It would use PF4 if our version of `@patternfly/react-core` were pinned that way, but it's not.) I don't see any reason not to use PF5 here.

chris1984: Sounds good will give that a go and watch the world burn 😂

jeremylenz: I think some props of `Grid` are changed since PF3, so you may have to adjust those. Other than that I don't foresee any issues :angel:

chris1984:
@lfu updated and download button works now.

lfu:
LGTM :+1: 

Buttons are grayed out when report is generating.
download1

Buttons are enabled with report generated.
d2

Upload page.
d3

chris1984:
@LadislavVasina1 here is this one for automation

LadislavVasina1:
@chris1984 Will jump on that right away

LadislavVasina1:
@chris1984 
Download report button is not greyed out when there is no report generated yet
image

and when I click the button I get an error. See the image.

image

chris1984:
> @chris1984 Download report button is not greyed out when there is no report generated yet image
> 
> and when I click the button I get an error. See the image.
> 
> image

@LadislavVasina1 fixed it, let me know if that works better for you:
screenshot-ip-10-0-168-90 rhos-01 prod psi rdu2 redhat com-2025 04 30-14_40_01

lib/foreman_inventory_upload.rb
```diff
@@ -30,6 +30,11 @@ def self.done_file_path(filename)
     File.join(ForemanInventoryUpload.done_folder, filename)
   end
 
+  def self.report_file_paths(organization_id)
+    filename = facts_archive_name(organization_id)
+    Dir["{#{ForemanInventoryUpload.uploads_file_path(filename)},#{ForemanInventoryUpload.done_file_path(filename)}}"]
```
lfu: This line can be written as:
```
Dir[ForemanInventoryUpload.uploads_file_path(filename), ForemanInventoryUpload.done_file_path(filename)]
```

chris1984: Fixed, thanks for catching that.

lfu (approved):
LGTM :+1:

PR #979, opened 2025-04-21
Allow the option for minimal data collection in inventory reports

jeremylenz:
[Red-Hat-Inventory.webm](https://github.com/user-attachments/assets/d1c17c3d-56bb-41d2-8129-9187236c0c9f)

This adds a new Setting and updates the UI to allow sending only minimal data in inventory upload reports.

Note the component is also reused on this screen in the Manage Manifest modal:
image

When "Minimal" is selected, the 3 settings below it are locked to on.
When "Optional" is selected, the user can change the 3 settings as they see fit.

Right now the new 'Minimal/Optional' setting doesn't do anything, other than the UI changes. @chris1984 will add the functionality in a followup, so that the uploaded reports are further trimmed down.


chris1984 (commented):
@lfu can you look at the react code here? I tested it when making my PR and it works fine, but would prefer a second set of 👀 The ruby code looks fine.

lfu:
It works well and the code looks great to me :+1: 
Here is some visual results that I got. Not sure if it only happens to me :)

advsetting


settings

chris1984:
> It works well and the code looks great to me 👍 Here is some visual results that I got. Not sure if it only happens to me :)
> 
> advsetting
> 
> settings

I saw something like this but then cleared my browser cache and it worked fine; can you try to clear your cache or use a different browser to see if you are getting the same thing?

jeremylenz:
squashed

lfu:
It's consistent with my local visual effects with 2 browsers :smile: 
I guess it just happens to me.

lfu (approved):

jeremylenz:
@lfu I do not see the issue with the button overflowing the modal.

I pushed a fix to the default `lockedValue` prop. 

- When 'Minimal' is selected, all child settings should slow as disabled and turned on
- When 'Optional' is selected, all child setting should be enabled and you can change them

lfu:
s2

It works.
The only thing is there is a DB transaction each time the switch is toggled. But there aren't too many of them.

jeremylenz:
> there is a DB transaction each time the switch is toggled.

It does actually change the settings via API. I think that's pretty lightweight.

lfu (approved):

jeremylenz:
Will wait to merge until QE automation ack :)

chris1984 (approved):

PR #983, opened 2025-05-02
Do not use http proxy settings for local advisor engine

tpapaioa:

Do not use the foreman or content http proxy setting if local advisor engine is enabled.

SAT-33408

1.) Client:
```
# insights-client
```

2.) Set a proxy (valid or not) on the Satellite:
```
# hammer http-proxy create --name=myproxy \
--url http://myproxy.example.com:8080  \
--username=proxy_username \
--password=proxy_password

# hammer settings set --name=content_default_http_proxy --value=myproxy
```

3.) Client:
```
# insights-client
```

### DIFF
--- a/lib/foreman_rh_cloud.rb
+++ b/lib/foreman_rh_cloud.rb
@@ -47,6 +47,8 @@ module ForemanRhCloud
   end
 
   def self.proxy_string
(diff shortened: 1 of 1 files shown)


parthaa (approved):
Works well. Tested on a sat.

PR #988, opened 2025-05-13
Add CVEs tab to Host Details page

nofaralfasi:
## Summary by Sourcery

Add a new CVEs tab to the host details page by creating the CVEsHostDetailsTab component, registering it in the fills and pages, and conditionally hiding it for non-RHEL hosts

New Features:
- Introduce CVEsHostDetailsTab component to display CVEs for a host
- Add a new 'CVEs' tab to the host details page with conditional visibility for RHEL hosts
- Register the CVEsHostDetailsTab in the application's page registry

Tests:
- Add initial test file for the CVEsHostDetailsTab component

### DIFF
--- a/app/views/api/v2/hosts/insights/single.rabl
+++ b/app/views/api/v2/hosts/insights/single.rabl
@@ -1,0 +1,5 @@
+node :vulnerability do
+  {
+    enabled: ForemanRhCloud.with_local_advisor_engine?,
+  }
+end
--- a/webpack/CVEsHostDetailsTab/CVEsHostDetailsTab.js
+++ b/webpack/CVEsHostDetailsTab/CVEsHostDetailsTab.js
@@ -1,0 +1,17 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { translate as __ } from 'foremanReact/common/I18n';
+
+const CVEsHostDetailsTab = ({ hostName }) => (
+  <div>
+    <h1>
+      {__('CVEs tab for host:')} {hostName}
+    </h1>
+  </div>
+);
+
+CVEsHostDetailsTab.propTypes = {
+  hostName: PropTypes.string.isRequired,
+};
+
+export default CVEsHostDetailsTab;
--- a/webpack/CVEsHostDetailsTab/__tests__/CVEsHostDetailsTab.test.js
+++ b/webpack/CVEsHostDetailsTab/__tests__/CVEsHostDetailsTab.test.js
@@ -1,0 +1,18 @@
+import React from 'react';
+import { render, screen } from '@testing-library/react';
+import CVEsHostDetailsTab from '../CVEsHostDetailsTab';
+
+describe('CVEsHostDetailsTab', () => {
+  it('renders without crashing', () => {
+    render(<CVEsHostDetailsTab hostName="test-host.example.com" />);
+    expect(
+      screen.getByText('CVEs tab for host: test-host.example.com')
+    ).toBeTruthy();
+  });
+
+  it('renders the host name', () => {
+    const hostName = 'test-host.example.com';
+    render(<CVEsHostDetailsTab hostName={hostName} />);
+    expect(screen.getByText(`CVEs tab for host: ${hostName}`)).toBeTruthy();
+  });
+});
--- a/webpack/ForemanRhCloudFills.js
+++ b/webpack/ForemanRhCloudFills.js
@@ -3,7 +3,8 @@ import { addGlobalFill } from 'foremanReact/components/common/Fill/GlobalFill';
 import InventoryAutoUploadSwitcher from './ForemanInventoryUpload/SubscriptionsPageExtension/InventoryAutoUpload';
 import NewHostDetailsTab from './InsightsHostDetailsTab/NewHostDetailsTab';
 import { InsightsTotalRiskChartWrapper } from './InsightsHostDetailsTab/InsightsTotalRiskChartWrapper';
-import { isNotRhelHost } from './ForemanRhCloudHelpers';
+import { isNotRhelHost, vulnerabilityDisabled } from './ForemanRhCloudHelpers';
+import CVEsHostDetailsTab from './CVEsHostDetailsTab/CVEsHostDetailsTab';
 
 const fills = [
   {
@@ -26,6 +27,15 @@ const fills = [
     name: 'insights-total-risk-chart',
     component: props => <InsightsTotalRiskChartWrapper {...props} />,
     weight: 2800,
+  },
+  {
+    slot: 'host-details-page-tabs',
+    name: 'CVEs',
+    component: props => <CVEsHostDetailsTab {...props} />,
+    weight: 300,
+    metadata: {
+      hideTab: vulnerabilityDisabled,
+    },
   },
 ];
 
--- a/webpack/ForemanRhCloudHelpers.js
+++ b/webpack/ForemanRhCloudHelpers.js
@@ -11,3 +11,6 @@ export const isNotRhelHost = ({ hostDetails }) =>
     // eslint-disable-next-line camelcase
     hostDetails?.operatingsystem_name
   );
+
+export const vulnerabilityDisabled = ({ hostDetails }) =>
+  isNotRhelHost({ hostDetails }) || !hostDetails?.vulnerability?.enabled;
--- a/lib/foreman_rh_cloud/engine.rb
+++ b/lib/foreman_rh_cloud/engine.rb
@@ -128,7 +128,7 @@ module ForemanRhCloud
 
           register_facet InsightsFacet, :insights do
             configure_host do
-              api_view :list => 'api/v2/hosts/insights/insights'
+              api_view :list => 'api/v2/hosts/insights/insights', :single => 'api/v2/hosts/insights/single'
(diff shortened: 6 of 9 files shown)


nofaralfasi:
I'm currently seeing the following error when rendering the new CVEs tab:
```
Failed prop type: The prop `hostID` is marked as required in `CvesHostDetailsTab`, but its value is `undefined`.
    in CvesHostDetailsTab (created by component)
    ...
```
It seems like `hostID` is not being passed correctly to the React component. I’m investigating the source now, but if anyone has insights on how `hostID` should be provided in this context (especially with the new pagelet or slot setup), I’d appreciate the input.

MariaAga:
Doesnt look like hostID is passed at all, can you use hostname instead? https://github.com/theforeman/foreman/blob/42c6c59f2f1b3ab6aeb11006816cd9feb76e1519/webpack/assets/javascripts/react_app/components/HostDetails/Tabs/TabRouter/index.js#L29

nofaralfasi:
> Doesnt look like hostID is passed at all, can you use hostname instead? https://github.com/theforeman/foreman/blob/42c6c59f2f1b3ab6aeb11006816cd9feb76e1519/webpack/assets/javascripts/react_app/components/HostDetails/Tabs/TabRouter/index.js#L29

Yeah, that worked. Thanks! I wasn't sure where to look.

MariaAga:
You can read about it here: https://github.com/theforeman/foreman/blob/develop/developer_docs/slot-and-fill.asciidoc
I search `host-details-page-tabs` in foreman core to see which props it has

webpack/ForemanRhCloudPages.js
```diff
@@ -4,11 +4,13 @@ import { registerRoutes as foremanRegisterRoutes } from 'foremanReact/routes/Rou
 import ForemanInventoryUpload from './ForemanInventoryUpload';
 import InsightsCloudSync from './InsightsCloudSync';
 import InsightsHostDetailsTab from './InsightsHostDetailsTab';
+import CVEsHostDetailsTab from './CVEsHostDetailsTab';
 
 const pages = [
   { name: 'ForemanInventoryUpload', type: ForemanInventoryUpload },
   { name: 'InsightsCloudSync', type: InsightsCloudSync },
   { name: 'InsightsHostDetailsTab', type: InsightsHostDetailsTab },
+  { name: 'CVEsHostDetailsTab', type: CVEsHostDetailsTab },
```
ShimShtein: I think this is needed only if you intend to use it from erb. Since we're using it from slot and fill, I think it's not needed here.

@MariaAga , please keep me honest here

webpack/CVEsHostDetailsTab/__tests__/CVEsHostDetailsTab.test.js
```diff
@@ -0,0 +1,27 @@
...
+      'CVEs tab for host: test-host.example.com'
+    );
+  });
+
+  it('requires hostName prop', () => {
+    expect(() => {
+      shallow(<CVEsHostDetailsTab />);
+    }).toThrow();
+  });
+});
```
MariaAga: Please use React Testing library or capybara for tests (this guideline is not yet written anywhere sorry)

webpack/ForemanRhCloudPages.js
MariaAga: Yup, componentRegistry is only for erb, right. The component is registered in this pr in `addGlobalFill`

webpack/CVEsHostDetailsTab/__tests__/CVEsHostDetailsTab.test.js
MariaAga: And delete the snapshot tests

webpack/CVEsHostDetailsTab/CVEsHostDetailsTab.js
```diff
@@ -0,0 +1,14 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+const CVEsHostDetailsTab = ({ hostName }) => (
+  <div>
+    <h1>CVEs tab for host: {hostName}</h1>
```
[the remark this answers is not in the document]
ShimShtein: @nofaralfasi This one needs translation, thanks AI!

webpack/ForemanRhCloudFills.js
```diff
@@ -27,6 +28,15 @@ const fills = [
     component: props => <InsightsTotalRiskChartWrapper {...props} />,
     weight: 2800,
   },
+  {
+    slot: 'host-details-page-tabs',
+    name: 'CVEs',
+    component: props => <CVEsHostDetailsTab {...props} />,
+    weight: 300,
+    metadata: {
+      hideTab: isNotRhelHost,
```
ShimShtein: @MariaAga , what do you think, can we incorporate here an if statement similar to https://github.com/theforeman/foreman_rh_cloud/blob/dff7c9f0aeeb76e370b2d5de09a90aab7c78bfbe/lib/foreman_rh_cloud/engine.rb#L118 ?

MariaAga: If we can pass it here: https://github.com/theforeman/foreman/blob/7a3905412e3f201b690913625902fa06a2522435/app/controllers/userdata_controller.rb#L17
And then we can use useForemanSetContext from foreman and add it there

ShimShtein: Awesome! Thanks! @nofaralfasi if you could add that, it would be awesome

app/views/api/v2/hosts/insights/single.rabl
```diff
@@ -0,0 +1,5 @@
+node :cves_attributes do
+  {
+    show_cves: !ForemanRhCloud.with_local_advisor_engine?
```
ShimShtein: `show_cves` should be `true` when running a local engine.
```suggestion
    show_cves: ForemanRhCloud.with_local_advisor_engine?
```

webpack/ForemanRhCloudHelpers.js
```diff
@@ -11,3 +11,7 @@ export const isNotRhelHost = ({ hostDetails }) =>
     // eslint-disable-next-line camelcase
     hostDetails?.operatingsystem_name
   );
+
+export const isNonRhelAndNoLocalAdvisor = ({ hostDetails }) =>
```
ShimShtein:
```suggestion
export const vulnrabilityEnabled = ({ hostDetails }) =>
```
I think it would be a bit more indicative

app/views/api/v2/hosts/insights/single.rabl
nofaralfasi: You're right, I copy-pasted it without thinking :)

jeremylenz:
Answering @ShimShtein from the email

> the main issue here is how to pass the ForemanRhCloud.with_local_advisor_engine? value to slot and fill in the UI

What I ended up doing seems a bit hacky but it works reliably -- you can use `useAdvisorEngineConfig` which makes its own API request to get the value:  https://github.com/theforeman/foreman_rh_cloud/blob/a7dbbb8e8b94a21ea9d342abcb5626c38405ff04/webpack/InsightsCloudSync/Components/InsightsSettings/InsightsSettings.js#L6

The component can then get the value whenever it needs it, so passing it thru a slot is not necessarily required.

webpack/ForemanRhCloudFills.js
```diff
@@ -27,6 +29,15 @@ const fills = [
     component: props => <InsightsTotalRiskChartWrapper {...props} />,
     weight: 2800,
   },
+  {
+    slot: 'host-details-page-tabs',
+    name: 'CVEs',
+    component: props => <CVEsHostDetailsTab {...props} />,
+    weight: 300,
+    metadata: {
+      hideTab: vulnrabilityEnabled,
```
ShimShtein: I think we want the opposite - we want to hide the tab when vulnerability feature is disabled.
```suggestion
      hideTab: vulnrabilityDisabled,
```
or
```suggestion
      hideTab: (hostDetails) =>  { !vulnrabilityEnabled(hostDetails) },
```

app/views/api/v2/hosts/insights/single.rabl
```diff
@@ -0,0 +1,5 @@
+node :cves_attributes do
+  {
+    show_cves: ForemanRhCloud.with_local_advisor_engine?
```
ShimShtein: Let's rename the nodes to be a bit more indicative
```suggestion
node :vulnerability do
  {
    enabled: ForemanRhCloud.with_local_advisor_engine?
```

ShimShtein (commented):
@nofaralfasi can you please address the linter issues in addition to my comments?
You can run the linter locally from the foreman directory: `npm run lint:plugins foreman_rh_cloud`

ShimShtein:
@jeremylenz Thanks for the answer! 
While we were looking at it, we came up with a bit different solution of extending the host info with this flag. It is a bit more future-proof, since we make the flag host-specific instead of system-wide.
WDYT?

jeremylenz:
I think if you can add the info to the existing API response as you have done, that's ideal. (There were weird stupid reasons I couldn't use that solution in the Recommendations table row kebabs IIRC.) I am fine keeping both solutions and we can use whatever makes the most sense given the situation.

ShimShtein (approved):

PR #997, opened 2025-06-05
SAT-34713 - Update Hybrid rake task to prompt for org_id and custom_url

chris1984:
* foreman-rake was changed which now clears all env vars except TERM and --whitelist
* Adds new prompts to the user to ask for things we had in the env vars
* Added some comments, added better validation for inputs and added color to the prompts.
* Added some error handling so the ruby traceback is not ugly, with keeping the option to turn it back on(commented)

## Summary by Sourcery

Update the Hybrid Cloud Rake task to interactively prompt for organization ID, custom Insights URL, and API token instead of relying on environment variables, with input validation, colored output, and enhanced error handling.

Enhancements:
- Replace environment variable inputs with interactive prompts for organization ID, custom URL, and API token.
- Validate that the organization ID is numeric and that the organization and token exist before proceeding.
- Add colored console output and separators to improve prompt clarity and user experience.
- Warn when using the default Insights URL and support custom URL input.
- Refactor task into logical sections with helper methods and inline comments for better readability.
- Implement specific error handling for unauthorized and other HTTP errors with clear logging and exit behavior.

### DIFF
--- a/lib/tasks/hybrid_cloud.rake
+++ b/lib/tasks/hybrid_cloud.rake
@@ -17,14 +17,40 @@ namespace :rh_cloud do
     include ::ForemanRhCloud::CertAuth
     include ::InsightsCloud::CandlepinCache
 
+    def default_registrations_url
+      URI.join(ForemanRhCloud.base_url, '/api/identity/certificate/registrations').to_s
+    end
+
     # Helper method to get the registrations URL, with a warning for default usage
     def registrations_url(custom_url)
       if custom_url.empty?
-        logger.warn("Custom url is not set, using the default one: #{ForemanRhCloud.base_url}")
-        URI.join(ForemanRhCloud.base_url, '/api/identity/certificate/registrations').to_s
+        logger.warn("Custom url is not set, using the default one: #{default_registrations_url}")
+        default_registrations_url
       else
-        URI.join(custom_url, '/api/identity/certificate/registrations').to_s
+        if URI(custom_url).scheme.nil?
+          logger.warn("Custom URL lacks a scheme; prepending https:// prefix.")
+          custom_url = "https://" + custom_url
+        end
+        custom_url
       end
+    end
+
+    def get_organization(user_org_id)
+      maybe_organization = Organization.find_by(id: user_org_id)
+      if maybe_organization.nil?
+        logger.error("Organization with ID '#{user_org_id}' not found.")
+        exit(1)
+      end
+      maybe_organization
+    end
+
+    def get_uid(organization)
+      maybe_uid = cp_owner_id(organization)
+      if maybe_uid.nil?
+        logger.error("Organization '#{organization}' does not have a manifest imported.")
+        exit(1)
+      end
+      maybe_uid
     end
 
     # --- Input Collection ---
@@ -40,38 +66,26 @@ namespace :rh_cloud do
     end
 
     puts "\n" + "-" * 50 + "\n\n"
-    puts 'Paste in your Custom Insights URL. If nothing is entered, the default will be used.'
+    puts "Paste in your custom Insights URL. If nothing is entered, the default will be used (#{default_registrations_url})."
     insights_user_input = STDIN.gets.chomp
-
-    # --- Data Preparation ---
-
-    organization = Organization.find_by(id: @user_org_id)
-
-    if organization.nil?
-      logger.error("Organization with ID '#{@user_org_id}' not found.")
-      exit(1)
-    end
-
-    uid = cp_owner_id(organization)
-    if uid.nil?
-      logger.error('Organization provided does not have a manifest imported.')
-      exit(1)
-    end
-
-    hostname = ForemanRhCloud.foreman_host_name
-    insights_url = registrations_url(insights_user_input)
 
     puts "\n" + "-" * 50 + "\n\n"
     puts 'Paste in your Hybrid Cloud API token, output will be hidden.'
     puts 'This token can be retrieved from the Hybrid Cloud console.'
     token = STDIN.noecho(&:gets).chomp
-
     if token.empty?
       logger.error('Token was not entered.')
       exit(1)
     end
 
-    # --- API Request Configuration ---
+    # --- Data Preparation ---
+
+    organization = get_organization(@user_org_id)
+    uid = get_uid(organization)
+    hostname = ForemanRhCloud.foreman_host_name
+    insights_url = registrations_url(insights_user_input)
+
+    # --- API Request ---
 
     headers = {
       Authorization: "Bearer #{token}",
@@ -82,8 +96,6 @@ namespace :rh_cloud do
       "display_name": "#{hostname}+#{organization.label}",
     }
 
-    # --- Execute Request ---
-
     begin
       response = execute_cloud_request(
         organization: organization,
@@ -93,8 +105,8 @@ namespace :rh_cloud do
         payload: payload.to_json
       )
       logger.debug("Cloud request completed: status=#{response.code}, body_preview=#{response.body&.slice(0, 200)}")
-    # Add a more specific rescue for 401 Unauthorized errors
     rescue RestClient::Unauthorized => _ex
+      # Add a more specific rescue for 401 Unauthorized errors
       logger.error('Registration failed: Your token is invalid or unauthorized. Please check your token and try again.')
       # Optionally, you can still log the full debug info if helpful for advanced troubleshooting
       # logger.debug(ex.backtrace.join("\n"))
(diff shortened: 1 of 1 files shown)


lib/tasks/hybrid_cloud.rake
```diff
@@ -1,64 +1,123 @@
...
-      logger.warn("Custom url is not set, using the default one: #{ForemanRhCloud.base_url}") if ENV['SATELLITE_RH_CLOUD_URL'].empty?
-      ForemanRhCloud.base_url + '/api/identity/certificate/registrations'
+    # Helper method to get the registrations URL, with a warning for default usage
+    def registrations_url(custom_url = nil)
+      if custom_url.blank?
+        logger.warn("Custom url is not set, using the default one: #{ForemanRhCloud.base_url}")
+        ForemanRhCloud.base_url + '/api/identity/certificate/registrations'
+      else
+        logger.warn("Custom Insights URL is set to: #{custom_url}")
+        custom_url
```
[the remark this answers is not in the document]
chris1984: Fixed

chris1984:
@sourcery-ai review

lib/tasks/hybrid_cloud.rake
```diff
@@ -1,64 +1,123 @@
...
+  # The task will then send a POST request to the Hybrid Cloud API to register the organization.
+  # The response will be logged, and any errors will be caught and logged as well.
+  # The task will exit with an error message if the organization does not have a manifest imported or if the token is not entered.
+  # The task will also log a warning if the custom URL is not set and the default one is used.
   task hybridcloud_register: [:environment] do
     include ::ForemanRhCloud::CertAuth
     include ::InsightsCloud::CandlepinCache
 
+    # Helper method to get a logger instance
     def logger
```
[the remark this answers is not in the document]
chris1984: Fixed

lib/tasks/hybrid_cloud.rake
```diff
@@ -1,64 +1,123 @@
...
-        organization: @organization,
-        method: method,
-        url: registrations_url,
+        organization: organization,
+        method: :post,
+        url: insights_url,
         headers: headers,
         payload: payload.to_json
       )
       logger.debug(response)
```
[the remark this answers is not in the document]
chris1984: Fixed

jeremy-audet:
This commit message says:

> foreman-rake was changed which now clears all env vars except TERM and --whitelist

But the full help text from runuser(1) is:

> -, -l, --login
> Start the shell as a login shell with an environment similar to a real login:
> •clears all the environment variables except for TERM and variables specified by --whitelist-environment
> •initializes the environment variables HOME, SHELL, USER, LOGNAME, and PATH
> •changes to the target user’s home directory
> •sets argv[0] of the shell to '-' in order to make the shell a login shell

chris1984:
> This commit message says:
> 
> > foreman-rake was changed which now clears all env vars except TERM and --whitelist
> 
> But the full help text from runuser(1) is:
> 
> > -, -l, --login
> > Start the shell as a login shell with an environment similar to a real login:
> > •clears all the environment variables except for TERM and variables specified by --whitelist-environment
> > •initializes the environment variables HOME, SHELL, USER, LOGNAME, and PATH
> > •changes to the target user’s home directory
> > •sets argv[0] of the shell to '-' in order to make the shell a login shell

Fixed the commit message to include the helptext you put into the Jira and comment.

lib/tasks/hybrid_cloud.rake
```diff
@@ -1,64 +1,124 @@
...
+  # The task will also log a warning if the custom URL is not set and the default one is used.
   task hybridcloud_register: [:environment] do
     include ::ForemanRhCloud::CertAuth
     include ::InsightsCloud::CandlepinCache
 
-    def logger
-      @logger ||= Logger.new(STDOUT)
+    # Helper method to get the registrations URL, with a warning for default usage
+    def registrations_url(custom_url = nil)
+      if custom_url.blank?
```
jeremy-audet: `nil.blank?` will throw an exception:

```
irb(main):001> "foo".empty?
=> false
irb(main):002> "".empty?
=> true
irb(main):003> nil.empty?
(irb):3:in `<main>': undefined method `empty?' for nil (NoMethodError)

nil.empty?
   ^^^^^^^
        from <internal:kernel>:187:in `loop'
        from /usr/share/gems/gems/irb-1.13.1/exe/irb:9:in `<top (required)>'
        from /usr/bin/irb:25:in `load'
        from /usr/bin/irb:25:in `<main>'
```

Can the function signature be changed to `registrations_url(custom_url = "")`?

lib/tasks/hybrid_cloud.rake
```diff
@@ -1,64 +1,124 @@
...
+        logger.warn("Custom Insights URL is set to: #{custom_url}")
+        URI.join(custom_url, '/api/identity/certificate/registrations').to_s
+      end
     end
 
-    def registrations_url
-      logger.warn("Custom url is not set, using the default one: #{ForemanRhCloud.base_url}") if ENV['SATELLITE_RH_CLOUD_URL'].empty?
-      ForemanRhCloud.base_url + '/api/identity/certificate/registrations'
+    # --- Input Collection ---
+    puts "\033[32mPaste in your organization ID, this can be retrieved with the command: hammer organization list\033[0m"
```
jeremy-audet: Is it a good idea to unconditionally print ANSI escape codes for setting font colors to the terminal? Not all terminals support colors.

chris1984: You make a good point, I will get rid of the colors and just keep the font normal.

jeremy-audet: Color support is a really nice enhancement, as long as controls are provided. As a point of comparison, here's what git-log(1) says:

```
       --color[=<when>]
           Show colored diff.  --color (i.e. without =<when>) is the same as --color=always.
           <when> can be one of always, never, or auto.
```

But I've definitely encountered cases where escape codes were not supported, and them being printed made it hard to read the output I cared about.

lib/tasks/hybrid_cloud.rake
```diff
@@ -1,64 +1,124 @@
...
 
-    def logger
-      @logger ||= Logger.new(STDOUT)
+    # Helper method to get the registrations URL, with a warning for default usage
+    def registrations_url(custom_url = nil)
+      if custom_url.blank?
+        logger.warn("Custom url is not set, using the default one: #{ForemanRhCloud.base_url}")
+        URI.join(ForemanRhCloud.base_url, '/api/identity/certificate/registrations').to_s
+      else
+        logger.warn("Custom Insights URL is set to: #{custom_url}")
```
jeremy-audet: If a user provides a custom URL, does it make sense to warn the user? My expectation is that a warning would only be provided in the case where a default URL is used.

jeremy-audet:
This issue's title references SAT-34717, but I believe the actual issue is SAT-34713.

lib/tasks/hybrid_cloud.rake
chris1984: Set it to:
```ruby
def registrations_url(custom_url)
[1] pry(main)> custom_url
=> ""
[2] pry(main)> custom_url.empty?
=> true
```

It's coming back as `""` if I don't type anything and hit just hit enter.

lib/tasks/hybrid_cloud.rake
chris1984: Removed the colors, for a script that is not used that much, no point to get fancy with it.

lib/tasks/hybrid_cloud.rake
chris1984: Removed the warning if using a custom URL

chris1984:
> This issue's title references SAT-34717, but I believe the actual issue is SAT-34713.

Changed the commit message and PR title

lib/tasks/hybrid_cloud.rake
```diff
@@ -1,64 +1,122 @@
...
     end
 
-    def payload
-      {
-        "uid": @uid,
-        "display_name": "#{@hostname}+#{@organization.label}",
-      }
+    uid = cp_owner_id(organization)
+    if uid.nil?
+      logger.error('Organization provided does not have a manifest imported.')
```
jeremy-audet: Could we add the organization name in this message? Something like:

```ruby
logger.error("Organization '#{organization}' does not have a manifest imported.")
```

jeremy-audet:
Can we look at this together on a call? I have a bunch of suggestions that aren't _crucial_, but IMO would make this code nicer, and it would be way faster if we looked things over together.

lib/tasks/hybrid_cloud.rake
chris1984: Fixed in our sync up

jeremy-audet (approved):
Tested against commercial, to the extent possible. I don't believe commercial supports mtls auth.

chris1984:
Merging since Jeremy Audet tested and we did a pair program session to resolve his comments.

PR #1002, opened 2025-06-17
Sat-34090 - Fix host obfuscation so it honors settings and insights-client

chris1984:
* Documented each method so we know what it's doing and not have to have hour debug sessions to figure out what's going on each time there is an issue
* Updated and added some new tests
* Changed the `obfuscate_hostname?` method to always return true if the `Setting[:obfuscate_inventory_hostnames]` is enabled, same with the `obfuscate_ips?` method.
* Changed the `fqdn` method to JSON parse the fact name `insights_client::obfuscated_hostname`
* Removed `@stream.simple_field('katello_agent_running', false)` from the slice generator since katello-agent is no longer in the product. 
* Added new params that `insights-client` is using and removed params that are no longer in `insights-client`

```ruby
'insights_client::obfuscate_ipv4_enabled',
'insights_client::obfuscate_ipv6_enabled',
'insights_client::obfuscated_ipv4',
'insights_client::obfuscated_ipv6',
'insights_client::obfuscated_hostname',
```

* Made sure that if a host is running `insights-client` and has host obfuscation enabled we report that correctly regardless of the setting

* Check out PR
* Register a host with `insights-client` just stock no config changes
* Register a host with `insights-client` and enable host obfuscation
* Turn on obfuscation of hostnames and ip's and generate a report
* Verify the `metadata.json` shows the obfuscated hostname
* Look at the slice report and make sure all hostnames and ip's are obfuscated and the `insights-client` hostname is what `insights-client` is reporting in not what we set them to.
* Turn off host and ip obfuscation and generate a report
* Check the `metadata.json` and see if the satellite/foreman hostname is not obfuscated
* Check the report slice and make sure the host with the stock `insights-client` is reporting the full fqdn and the one that is obfuscated with `insights-client` shows the correct obfuscated hostname.

## Summary by Sourcery

Fix host and IP obfuscation logic to consistently honor global settings and insights-client facts, parse JSON-encoded obfuscated data, and clean up legacy fields and queries.

Bug Fixes:
- Ensure hostname and IP obfuscation always applies when the corresponding global settings are enabled
- Respect host-specific insights-client facts for enabling or disabling obfuscation

Enhancements:
- Revamp `fqdn`, `obfuscate_hostname?`, and `obfuscate_ips?` helpers to parse JSON arrays of obfuscated hostnames and IPv4/IPv6 addresses from insights_client facts
- Introduce a dynamic `default_proc` in `obfuscated_ips` to generate sequential obfuscated IPs when not already provided
- Remove the obsolete `katello_agent_running` field from the slice generator

Tests:
- Add and update unit tests to cover scenarios combining global settings and insights-client facts for hostname and IP obfuscation


lib/foreman_inventory_upload/generators/fact_helpers.rb
```diff
@@ -57,55 +57,187 @@ def cloud_provider(host)
...
+        # dynamically obfuscate new IPs if they are not already present in the facts.
+
+        # Parse the 'insights_client::obfuscated_ipv4' fact value as a JSON array.
+        # This fact is expected to contain a list of hashes, where each hash maps
+        # an 'original' IP to its 'obfuscated' counterpart.
+        # If the fact is not present or empty, default to an empty JSON array to prevent parsing errors.
+        # Example format of `insights_client_ips`:
+        # [{"original": "192.168.1.10", "obfuscated": "10.230.230.1"},
+        #  {"original": "192.168.1.11", "obfuscated": "10.230.230.2"}]
+        insights_client_ips = JSON.parse(fact_value(host, 'insights_client::obfuscated_ipv4') || '[]')
```
jeremylenz: May be good to add a rescue here similar to the JsonParserError rescue you added above.

lib/foreman_inventory_upload/generators/fact_helpers.rb
```diff
@@ -57,55 +57,187 @@ def cloud_provider(host)
...
+
+        # If the host-specific setting is not nil (meaning it was explicitly set to true or false),
+        # return its boolean value directly. This prioritizes the host-specific setting
+        # over the global default if the host has a specific configuration.
         return insights_client_setting unless insights_client_setting.nil?
 
+        # If the host-specific setting was nil (meaning it wasn't explicitly set for this host),
+        # fall back to the global 'obfuscate_inventory_hostnames' setting.
+        # This serves as the default behavior when no specific host configuration is found.
         Setting[:obfuscate_inventory_hostnames]
```
jeremylenz: Could be a bit more concise here with a ternary. But if you think it adds value I am also okay with leaving it more explicit, up to you.

```suggestion
        # If the host-specific setting is nil (meaning it wasn't explicitly set for this host),
        # fall back to the global 'obfuscate_inventory_hostnames' setting.
        # This serves as the default behavior when no specific host configuration is found.
        # Otherwise, return the host-specific setting directly.
        insights_client_setting.nil? ? Setting[:obfuscate_inventory_hostnames] : insights_client_setting
```

lib/foreman_inventory_upload/generators/fact_helpers.rb
```diff
@@ -57,55 +57,187 @@ def cloud_provider(host)
...
+        # This method determines whether the hostname associated with a host
+        # should be hidden or masked based on global settings and host-specific facts.
+
+        # Check the global setting for hostname obfuscation.
+        # If 'obfuscate_inventory_hostnames' is set to true in the application settings,
+        # then hostname obfuscation is always enabled, regardless of host-specific settings.
+        return true if Setting[:obfuscate_inventory_hostnames]
+
+        # Retrieve the host-specific fact related to hostname obfuscation from 'insights_client'.
+        # This fact indicates whether hostname obfuscation is enabled for this particular host.
```
jeremylenz: This line doesn't really say anything.
```suggestion
```

lib/foreman_inventory_upload/generators/fact_helpers.rb
```diff
@@ -57,55 +57,187 @@ def cloud_provider(host)
...
+          begin
+            parsed_insights_array = JSON.parse(fact_value(host, 'insights_client::obfuscated_hostname') || '[]')
+          rescue JSON::ParserError
+            # If the fact contains invalid JSON, default to an empty array to prevent upload failures.
+            parsed_insights_array = []
+          end
+
+          # Find the specific item in the parsed array where the 'original' hostname matches
+          # the host's actual FQDN.
+          parsed_insights_item = parsed_insights_array.find { |item| item['original'] == host.fqdn }
```
jeremylenz: I think the `find` is pretty self-explanatory here
```suggestion
          parsed_insights_item = parsed_insights_array.find { |item| item['original'] == host.fqdn }
```

lib/foreman_inventory_upload/generators/fact_helpers.rb
```diff
@@ -57,55 +57,187 @@ def cloud_provider(host)
...
 
       def obfuscate_ips?(host)
-        insights_client_setting = fact_value(host, 'insights_client::obfuscate_ip_enabled')
-        insights_client_setting = ActiveModel::Type::Boolean.new.cast(insights_client_setting)
-        return insights_client_setting unless insights_client_setting.nil?
-
-        Setting[:obfuscate_inventory_ips]
+        # Returns true if IP obfuscation should be applied for a given host.
+        # This method determines whether IP addresses associated with a host
+        # should be hidden or masked based on global settings and host-specific facts.
```
jeremylenz: This is kinda just expanding on the previous sentence and isn't too useful IMO.
```suggestion
```

lib/foreman_inventory_upload/generators/fact_helpers.rb
```diff
@@ -57,55 +57,187 @@ def cloud_provider(host)
...
+        # Retrieve host-specific facts related to IP obfuscation from 'insights_client'.
+        # These facts indicate whether IPv4 and IPv6 obfuscation are enabled for this particular host.
+        insights_client_ipv4_setting = fact_value(host, 'insights_client::obfuscate_ipv4_enabled')
+        insights_client_ipv6_setting = fact_value(host, 'insights_client::obfuscate_ipv6_enabled')
+
+        # Cast the retrieved settings to boolean values.
+        # This ensures that the values from the facts (which might be strings or other types)
+        # are correctly interpreted as true or false.
+        casted_ipv4_setting = ActiveModel::Type::Boolean.new.cast(insights_client_ipv4_setting)
+        casted_ipv6_setting = ActiveModel::Type::Boolean.new.cast(insights_client_ipv6_setting)
```
jeremylenz: I don't think "casted" is a word

```suggestion
        cast_ipv4_setting = ActiveModel::Type::Boolean.new.cast(insights_client_ipv4_setting)
        cast_ipv6_setting = ActiveModel::Type::Boolean.new.cast(insights_client_ipv6_setting)
```

lib/foreman_inventory_upload/generators/fact_helpers.rb
```diff
@@ -57,55 +57,187 @@ def cloud_provider(host)
...
 
       def host_ips(host)
+        # Determines and returns the IP addresses associated with a host,
+        # applying obfuscation if enabled.
+
+        # If IP obfuscation is enabled for the host (checked via `obfuscate_ips?`),
+        # return a representation of obfuscated IP addresses.
+        # The `obfuscated_ips` method (not shown here) would handle the actual
+        # obfuscation logic and likely return a structure that maps original IPs
+        # to their obfuscated counterparts.
```
jeremylenz: Why is this hypothetical? The method is right there.

lib/foreman_inventory_upload/generators/fact_helpers.rb
```diff
@@ -57,55 +57,187 @@ def cloud_provider(host)
...
+        # it defaults to the integer representation of '10.230.230.0'.
+        # This default provides a starting point for obfuscated IPs if none exist yet.
         max_obfuscated = ips_dict.values.map { |v| IPAddr.new(v).to_i }.max || IPAddr.new('10.230.230.0').to_i
 
+        # Generate the new obfuscated IP address.
+        # We increment the `max_obfuscated` integer by 1 to ensure the new IP is unique
+        # and sequential.
+        # Then, we convert this integer back into an IPv4 address string.
+        # `Socket::AF_INET` specifies that we are dealing with an IPv4 address.
         IPAddr.new(max_obfuscated + 1, Socket::AF_INET).to_s
```
jeremylenz, line 161: Does this always return an IPv4 address even when the unobfuscated one is IPv6? If so, should we think about changing that in the future?

jeremylenz (commented):
Thanks for assembling this @chris1984! Some initial comments below.

As far as tests, I think it would also be good to have a test (if we don't already) making sure that when the Setting is set, we always obfuscate regardless of insights_client setting. I think that was the source of the original request.

ColeHiggins2 (approved):
Ack pending comments

Passed PACKIT build with prt build #11916

parthaa:
I like this PR but feel the comments are little too much. @jeremylenz @chris1984 can over rule me on this but this would be my Suggested patch
```patch
diff --git a/lib/foreman_inventory_upload/generators/fact_helpers.rb b/lib/foreman_inventory_upload/generators/fact_helpers.rb
index 7d42599..d8b7f45 100644
--- a/lib/foreman_inventory_upload/generators/fact_helpers.rb
+++ b/lib/foreman_inventory_upload/generators/fact_helpers.rb
@@ -62,17 +62,11 @@ module ForemanInventoryUpload
         # should be hidden or masked based on global settings and host-specific facts.
 
         # Check the global setting for hostname obfuscation.
-        # If 'obfuscate_inventory_hostnames' is set to true in the application settings,
-        # then hostname obfuscation is always enabled, regardless of host-specific settings.
         return true if Setting[:obfuscate_inventory_hostnames]
 
-        # Retrieve the host-specific fact related to hostname obfuscation from 'insights_client'.
+        # Retrieve the host-sspecific fact related to hostname obfuscation from 'insights_client'.
         # This fact indicates whether hostname obfuscation is enabled for this particular host.
         insights_client_setting = fact_value(host, 'insights_client::obfuscate_hostname_enabled')
-
-        # Cast the retrieved setting to a boolean value.
-        # This ensures that the value from the fact (which might be a string or other type)
-        # is correctly interpreted as true or false.
         insights_client_setting = ActiveModel::Type::Boolean.new.cast(insights_client_setting)
 
         # If the host-specific setting is not nil (meaning it was explicitly set to true or false),
@@ -80,10 +74,8 @@ module ForemanInventoryUpload
         # over the global default if the host has a specific configuration.
         return insights_client_setting unless insights_client_setting.nil?
 
-        # If the host-specific setting was nil (meaning it wasn't explicitly set for this host),
-        # fall back to the global 'obfuscate_inventory_hostnames' setting.
-        # This serves as the default behavior when no specific host configuration is found.
-        Setting[:obfuscate_inventory_hostnames]
+        # return false if the host-specific setting was nil (meaning it wasn't explicitly set for this host),
+        false
       end
 
       def fqdn(host)
@@ -96,7 +88,6 @@ module ForemanInventoryUpload
           # This fact is expected to contain a JSON array of mappings between original and obfuscated hostnames.
 
           # Parse the 'insights_client::obfuscated_hostname' fact value as a JSON array.
-          # If the fact is not present or empty, default to an empty array to prevent parsing errors.
           # Example format of `parsed_insights_array`:
           # [{"original"=>"host.example.com", "obfuscated"=>"0dd449d0a027.example.com"},
           #  {"original"=>"satellite.example.com", "obfuscated"=>"host2.example.com"}]
@@ -112,7 +103,6 @@ module ForemanInventoryUpload
           parsed_insights_item = parsed_insights_array.find { |item| item['original'] == host.fqdn }
 
           # Return the 'obfuscated' value from the found item if it exists.
-          # The `&.` (safe navigation operator) prevents an error if `parsed_insights_item` is nil.
           # If no matching obfuscated hostname is found in the fact, or if the fact itself was empty,
           # then dynamically obfuscate the host's FQDN using the `obfuscate_fqdn` helper method.
           parsed_insights_item&.[]('obfuscated') || obfuscate_fqdn(host.fqdn)
@@ -132,48 +122,37 @@ module ForemanInventoryUpload
         # should be hidden or masked based on global settings and host-specific facts.
 
         # Check the global setting for IP obfuscation.
-        # If 'obfuscate_inventory_ips' is set to true in the application settings,
-        # then IP obfuscation is always enabled, regardless of host-specific settings.
         return true if Setting[:obfuscate_inventory_ips]
 
-        # Retrieve host-specific facts related to IP obfuscation from 'insights_client'.
         # These facts indicate whether IPv4 and IPv6 obfuscation are enabled for this particular host.
         insights_client_ipv4_setting = fact_value(host, 'insights_client::obfuscate_ipv4_enabled')
         insights_client_ipv6_setting = fact_value(host, 'insights_client::obfuscate_ipv6_enabled')
 
-        # Cast the retrieved settings to boolean values.
-        # This ensures that the values from the facts (which might be strings or other types)
-        # are correctly interpreted as true or false.
         casted_ipv4_setting = ActiveModel::Type::Boolean.new.cast(insights_client_ipv4_setting)
         casted_ipv6_setting = ActiveModel::Type::Boolean.new.cast(insights_client_ipv6_setting)
 
         # Determine the final obfuscation decision.
         # IP obfuscation is enabled if:
         # - The host's IPv4 obfuscation setting is true OR
-        # - The host's IPv6 obfuscation setting is true OR
-        # - The global 'obfuscate_inventory_ips' setting is true (this condition was already checked above,
-        #   but it's good to keep it for logical completeness if the initial return guard wasn't there).
-        casted_ipv4_setting || casted_ipv6_setting || Setting[:obfuscate_inventory_ips]
+        # - The host's IPv6 obfuscation setting is true
+        casted_ipv4_setting || casted_ipv6_setting || false
       end
 
       def host_ips(host)
         # Determines and returns the IP addresses associated with a host,
         # applying obfuscation if enabled.
 
-        # If IP obfuscation is enabled for the host (checked via `obfuscate_ips?`),
+        # If IP obfuscation is enabled for the host
         # return a representation of obfuscated IP addresses.
-        # The `obfuscated_ips` method (not shown here) would handle the actual
-        # obfuscation logic and likely return a structure that maps original IPs
-        # to their obfuscated counterparts.
         return obfuscated_ips(host) if obfuscate_ips?(host)
 
         # If IP obfuscation is NOT needed, return a special kind of Hash.
-        # This Hash acts as a "pass-through proxy." When you try to access a key in it,
+        # where when you try to access a key in it,
         # if the key doesn't exist, it simply returns the key itself.
         # This is useful because it means if you try to get an IP from this hash,
-        # and that IP isn't explicitly defined as something to be obfuscated,
         # you'll just get the original IP back. It allows the calling code to
         # use the same interface whether obfuscation is applied or not.
+
         Hash.new { |h, k| k }
       end
 
@@ -182,17 +161,15 @@ module ForemanInventoryUpload
         # This method primarily uses pre-obfuscated IPs from host facts, but can also
         # dynamically obfuscate new IPs if they are not already present in the facts.
 
-        # Parse the 'insights_client::obfuscated_ipv4' fact value as a JSON array.
+        # Parse the 'insights_client::obfuscated_ipv4' fact value.
         # This fact is expected to contain a list of hashes, where each hash maps
         # an 'original' IP to its 'obfuscated' counterpart.
-        # If the fact is not present or empty, default to an empty JSON array to prevent parsing errors.
         # Example format of `insights_client_ips`:
         # [{"original": "192.168.1.10", "obfuscated": "10.230.230.1"},
         #  {"original": "192.168.1.11", "obfuscated": "10.230.230.2"}]
         insights_client_ips = JSON.parse(fact_value(host, 'insights_client::obfuscated_ipv4') || '[]')
 
         # Create a new Hash to store the mapping from original IP addresses to their obfuscated versions.
-        # The `map` function transforms each `ip_record` into a key-value pair,
         # where the 'original' IP is the key and the 'obfuscated' IP is the value.
         obfuscated_ips = Hash[
           insights_client_ips.map { |ip_record| [ip_record['original'], ip_record['obfuscated']] }
@@ -213,8 +190,6 @@ module ForemanInventoryUpload
           hash[key] = obfuscate_ip(key, hash)
         end
 
-        # Return the `obfuscated_ips` hash, which now serves as the primary interface
-        # for retrieving obfuscated IP addresses for the given host.
         obfuscated_ips
       end
 
@@ -224,20 +199,9 @@ module ForemanInventoryUpload
         # numerically one greater than the highest existing obfuscated IP,
         # ensuring a sequential and non-conflicting assignment within a specific range.
 
-        # Determine the maximum integer value among all currently obfuscated IP addresses
-        # found in `ips_dict`. `ips_dict` is expected to be a hash where values are
-        # already obfuscated IP strings.
-        # We convert each obfuscated IP string to an integer for numerical comparison.
-        # If `ips_dict` is empty or contains no valid IP addresses,
-        # it defaults to the integer representation of '10.230.230.0'.
-        # This default provides a starting point for obfuscated IPs if none exist yet.
         max_obfuscated = ips_dict.values.map { |v| IPAddr.new(v).to_i }.max || IPAddr.new('10.230.230.0').to_i
 
         # Generate the new obfuscated IP address.
-        # We increment the `max_obfuscated` integer by 1 to ensure the new IP is unique
-        # and sequential.
-        # Then, we convert this integer back into an IPv4 address string.
-        # `Socket::AF_INET` specifies that we are dealing with an IPv4 address.
         IPAddr.new(max_obfuscated + 1, Socket::AF_INET).to_s
       end
```

ehelms:
> I like this PR but feel the comments are little too much.

In most cases the comments are just restating what the code already says and thus creating noise that makes it hard to read the code. I am a proponent of self-documenting code (https://en.wikipedia.org/wiki/Self-documenting_code) and that comments should be used sparingly to help with tricky concepts or explain why a design choice was made. As I play with AI code tools, I am finding they have a tendency to over comment everything and I am often having to ask it to remove all the comments to prevent hard to read code blocks.

jeremylenz:
To give a bit of context here: This PR came out of several hours of debugging and reverse-engineering sessions, though it may not look like it. We did not understand some basics about how foreman_rh_cloud handles obfuscation, and the idea with the comments was to prevent that reverse engineering from ever needing to happen in the future. So it's just trying to find a balance. The original code, as it was, was definitely not self-documenting.

ehelms:
> To give a bit of context here: This PR came out of several hours of debugging and reverse-engineering sessions, though it may not look like it. We did not understand some basics about how foreman_rh_cloud handles obfuscation, and the idea with the comments was to prevent that reverse engineering from ever needing to happen in the future. So it's just trying to find a balance. The original code, as it was, was definitely not self-documenting.

Fair. Do you think there are ways to improve the readability of the code through better naming of variables or use of methods? Your description of the process of understanding this whole area makes me think this would benefit from a design write-up that outlines how the obfuscation should work. I am guessing the bones of that are captured in the comments + sourcery summary. But to capture it somewhere more permanent.

chris1984:
I would prefer to go with Parthas suggestion and reducing the comments to only what is needed. Since we are close to dev freeze I am against changing methods and function especially since Cole has already wrote automation with packit. Right now it solves the issue in the Jira.

chris1984:
We do have a pr up from deepwiki that goes over the entire repo in a deep dive. I'm happy to do a design write up after this gets merged.

jeremylenz:
> I would prefer to go with Parthas suggestion and reducing the comments to only what is needed. 

Reducing extraneous comments is +1 from me, let's do it. 

One other thought I had that would help - I often like to look at tests as documentation. If you have a set of tests that says "This should do A,B, and C, in this way" it can also avoid comments - especially if the test method names/descriptions are, well, descriptive enough. So maybe take each one of the comments and make sure we have a unit test that asserts that same thing.

lib/foreman_inventory_upload/generators/fact_helpers.rb
chris1984: Added a Jira to track that work so we can come back to eventually in the future https://issues.redhat.com/browse/SAT-35600

lib/foreman_inventory_upload/generators/fact_helpers.rb
chris1984: Fixed

lib/foreman_inventory_upload/generators/fact_helpers.rb
```diff
@@ -57,55 +57,187 @@ def cloud_provider(host)
...
+        # ensuring a sequential and non-conflicting assignment within a specific range.
+
+        # Determine the maximum integer value among all currently obfuscated IP addresses
+        # found in `ips_dict`. `ips_dict` is expected to be a hash where values are
+        # already obfuscated IP strings.
+        # We convert each obfuscated IP string to an integer for numerical comparison.
+        # If `ips_dict` is empty or contains no valid IP addresses,
+        # it defaults to the integer representation of '10.230.230.0'.
+        # This default provides a starting point for obfuscated IPs if none exist yet.
         max_obfuscated = ips_dict.values.map { |v| IPAddr.new(v).to_i }.max || IPAddr.new('10.230.230.0').to_i
```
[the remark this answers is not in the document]
chris1984: We are just putting fake ip's in a report for console.redhat.com so it does not matter what the values are.

lib/foreman_inventory_upload/generators/fact_helpers.rb
chris1984: Added

chris1984:
@jeremylenz took parthas patch, fixed all your comments, and went through and reread all of them removing even more comments, when you are free to take a look again.

lib/foreman_inventory_upload/generators/fact_helpers.rb
```diff
@@ -57,55 +57,141 @@ def cloud_provider(host)
       end
 
       def obfuscate_hostname?(host)
+        # Returns true if hostname obfuscation should be applied for a given host.
+        # This method determines whether the hostname associated with a host
+        # should be hidden or masked based on global settings and host-specific facts.
+
+        # Check the global setting for hostname obfuscation.
```
jeremylenz:
```suggestion
        # Returns true if hostname obfuscation should be applied for a given host, based on hierarchy:

        # 1. Global setting
```

lib/foreman_inventory_upload/generators/fact_helpers.rb
```diff
@@ -57,55 +57,141 @@ def cloud_provider(host)
...
+        return true if Setting[:obfuscate_inventory_hostnames]
+
         insights_client_setting = fact_value(host, 'insights_client::obfuscate_hostname_enabled')
         insights_client_setting = ActiveModel::Type::Boolean.new.cast(insights_client_setting)
-        return insights_client_setting unless insights_client_setting.nil?
 
-        Setting[:obfuscate_inventory_hostnames]
+        # If the insights_client_setting is nil fall back to the global 'obfuscate_inventory_hostnames' setting.
+        insights_client_setting.nil? ? false : insights_client_setting
```
jeremylenz:
```suggestion
        # 2. host fact reported by insights_client
        # 3. if neither of the above, don't obfuscate.
        insights_client_setting.nil? ? false : insights_client_setting
```

lib/foreman_inventory_upload/generators/fact_helpers.rb
```diff
@@ -57,55 +57,141 @@ def cloud_provider(host)
...
       end
 
       def fqdn(host)
-        return host.fqdn unless obfuscate_hostname?(host)
-
-        fact_value(host, 'insights_client::hostname') || obfuscate_fqdn(host.fqdn)
+        # Returns the Fully Qualified Domain Name (FQDN) for a given host,
+        # potentially obfuscating it if hostname obfuscation is enabled.
```
jeremylenz:
```suggestion
```

lib/foreman_inventory_upload/generators/fact_helpers.rb
```diff
@@ -57,55 +57,141 @@ def cloud_provider(host)
...
+        if obfuscate_hostname?(host)
+          # If obfuscation is enabled, attempt to retrieve an already obfuscated hostname
+          # from the 'insights_client::obfuscated_hostname' fact.
+          # This fact is expected to contain a JSON array of mappings between original and obfuscated hostnames.
+
+          # Parse the 'insights_client::obfuscated_hostname' fact value as a JSON array.
+          # If the fact is not present or empty, default to an empty array to prevent parsing errors.
```
jeremylenz:
```suggestion
```

lib/foreman_inventory_upload/generators/fact_helpers.rb
```diff
@@ -57,55 +57,141 @@ def cloud_provider(host)
...
+          # Example format of `parsed_insights_array`:
+          # [{"original"=>"host.example.com", "obfuscated"=>"0dd449d0a027.example.com"},
+          #  {"original"=>"satellite.example.com", "obfuscated"=>"host2.example.com"}]
+          begin
+            parsed_insights_array = JSON.parse(fact_value(host, 'insights_client::obfuscated_hostname') || '[]')
+          rescue JSON::ParserError
+            # If the fact contains invalid JSON, default to an empty array to prevent upload failures.
```
jeremylenz:
```suggestion
```

lib/foreman_inventory_upload/generators/fact_helpers.rb
```diff
@@ -57,55 +57,141 @@ def cloud_provider(host)
...
+            parsed_insights_array = []
+          end
+
```
jeremylenz:
```suggestion
# Obfuscate using the following hierarchy:

# 1. the obfuscated_hostname fact sent by insights_client
```

lib/foreman_inventory_upload/generators/fact_helpers.rb
```diff
@@ -57,55 +57,141 @@ def cloud_provider(host)
...
+          parsed_insights_item = parsed_insights_array.find { |item| item['original'] == host.fqdn }
+          # If no matching obfuscated hostname is found in the fact, or if the fact itself was empty,
+          # then dynamically obfuscate the host's FQDN using the `obfuscate_fqdn` helper method.
```
jeremylenz:
```suggestion
          # 2. our own helper method
```

lib/foreman_inventory_upload/generators/fact_helpers.rb
```diff
@@ -57,55 +57,141 @@ def cloud_provider(host)
...
       end
 
       def obfuscate_ips?(host)
-        insights_client_setting = fact_value(host, 'insights_client::obfuscate_ip_enabled')
-        insights_client_setting = ActiveModel::Type::Boolean.new.cast(insights_client_setting)
-        return insights_client_setting unless insights_client_setting.nil?
+        # Returns true if IP obfuscation should be applied for a given host.
 
-        Setting[:obfuscate_inventory_ips]
+        # Check the global setting for IP obfuscation.
```
jeremylenz:
```suggestion
        # Obfuscate using the following hierarchy:
        
        # 1. Global setting
```

lib/foreman_inventory_upload/generators/fact_helpers.rb
```diff
@@ -57,55 +57,141 @@ def cloud_provider(host)
...
+        return true if Setting[:obfuscate_inventory_ips]
+
+        # These facts indicate whether IPv4 and IPv6 obfuscation are enabled for this particular host.
```
jeremylenz:
```suggestion
        # 2. Host facts sent by insights_client (either IPv4 or IPv6)
```

lib/foreman_inventory_upload/generators/fact_helpers.rb
```diff
@@ -57,55 +57,141 @@ def cloud_provider(host)
...
+        insights_client_ipv4_setting = fact_value(host, 'insights_client::obfuscate_ipv4_enabled')
+        insights_client_ipv6_setting = fact_value(host, 'insights_client::obfuscate_ipv6_enabled')
+
+        cast_ipv4_setting = ActiveModel::Type::Boolean.new.cast(insights_client_ipv4_setting)
+        cast_ipv6_setting = ActiveModel::Type::Boolean.new.cast(insights_client_ipv6_setting)
+
+        # Determine the final obfuscation decision.
+        # IP obfuscation is enabled if:
+        # - The host's IPv4 obfuscation setting is true OR
+        # - The host's IPv6 obfuscation setting is true OR
```
jeremylenz:
```suggestion
        # 3. If none of the above, don't obfuscate.
```

lib/foreman_inventory_upload/generators/fact_helpers.rb
```diff
@@ -57,55 +57,141 @@ def cloud_provider(host)
...
+        cast_ipv4_setting || cast_ipv6_setting || false
       end
 
       def host_ips(host)
+        # Determines and returns the IP addresses associated with a host,
+        # applying obfuscation if enabled.
+
+        # If IP obfuscation is enabled for the host
+        # return a representation of obfuscated IP addresses.
```
jeremylenz:
```suggestion
```

lib/foreman_inventory_upload/generators/fact_helpers.rb
```diff
@@ -57,55 +57,141 @@ def cloud_provider(host)
...
         return obfuscated_ips(host) if obfuscate_ips?(host)
 
-        # return a pass through proxy hash in case no obfuscation needed
+        # If IP obfuscation is NOT needed, return a special kind of Hash.
+        # where when you try to access a key in it
+        # if the key doesn't exist, it simply returns the key itself.
+        # This is useful because it means if you try to get an IP from this hash,
+        # you'll just get the original IP back. It allows the calling code to
+        # use the same interface whether obfuscation is applied or not.
         Hash.new { |h, k| k }
```
jeremylenz, line 127: I know experienced Rubyists will call this comment unnecessary, but I vote to keep it in (since I had to refresh my memory there)

lib/foreman_inventory_upload/generators/fact_helpers.rb
```diff
@@ -57,55 +57,141 @@ def cloud_provider(host)
...
       def obfuscated_ips(host)
-        insights_client_ips = JSON.parse(fact_value(host, 'insights_client::ips') || '[]')
+        # Retrieves and provides a mechanism to access obfuscated IP addresses for a given host.
+        # This method primarily uses pre-obfuscated IPs from host facts, but can also
+        # dynamically obfuscate new IPs if they are not already present in the facts.
+
+        # Parse the 'insights_client::obfuscated_ipv4' fact value
+        # This fact is expected to contain a list of hashes, where each hash maps
+        # an 'original' IP to its 'obfuscated' counterpart.
+        # If the fact is not present or empty, default to an empty JSON array to prevent parsing errors.
```
jeremylenz:
```suggestion
```

lib/foreman_inventory_upload/generators/fact_helpers.rb
```diff
@@ -57,55 +57,141 @@ def cloud_provider(host)
...
+        # Example format of `insights_client_ips`:
```
jeremylenz:
```suggestion
        # Example format of `parsed_insights_array`:
```

lib/foreman_inventory_upload/generators/fact_helpers.rb
```diff
@@ -57,55 +57,141 @@ def cloud_provider(host)
...
+        # This `default_proc` defines what happens when you try to access a key (an original IP)
+        # that does not yet exist in the `obfuscated_ips` hash.
+        # When an unknown key `k` is accessed:
+        # 1. `obfuscate_ip(key, hash)` is called to generate a new, unique obfuscated IP for `key`.
+        #    The `hash` itself is passed to `obfuscate_ip` so it can consider existing obfuscated IPs
+        #    when generating a new one (e.g., to ensure uniqueness and sequential assignment).
+        # 2. The newly generated obfuscated IP is then stored in the `obfuscated_ips` hash with `key`.
+        # 3. The newly generated obfuscated IP is returned.
+        # This ensures that any IP requested through this hash will either retrieve a pre-existing
+        # obfuscated value or generate a new one on the fly and store it for future access.
```
jeremylenz: Replace this with a comment similar to the one above ("special kind of Hash")

lib/foreman_inventory_upload/generators/fact_helpers.rb
```diff
@@ -57,55 +57,141 @@ def cloud_provider(host)
...
         end
-
         obfuscated_ips
       end
 
       def obfuscate_ip(ip, ips_dict)
+        # Generates an obfuscated IP address based on existing obfuscated IPs.
+        # This method aims to produce a new, unique obfuscated IP that is
+        # numerically one greater than the highest existing obfuscated IP,
+        # ensuring a sequential and non-conflicting assignment within a specific range.
```
jeremylenz:
```suggestion
        # Produce a new, unique obfuscated IP that is
        # numerically one greater than the highest existing obfuscated IP
```

lib/foreman_inventory_upload/generators/fact_helpers.rb
```diff
@@ -57,55 +57,141 @@ def cloud_provider(host)
...
         max_obfuscated = ips_dict.values.map { |v| IPAddr.new(v).to_i }.max || IPAddr.new('10.230.230.0').to_i
 
+        # Generate the new obfuscated IP address.
```
jeremylenz:
```suggestion
```

test/unit/slice_generator_test.rb
```diff
@@ -418,14 +419,38 @@ def fact_names
...
+    json_str = generator.render
+    actual = JSON.parse(json_str.join("\n"))
+
+    assert_not_nil(actual_host = actual['hosts'].first)
+    assert_equal @host.fqdn, actual_host['fqdn'], "FQDN should not be obfuscated when obfuscate_hostname_enabled is missing and setting is false"
+    assert_not_nil(actual_facts = actual_host['facts'].first['facts'])
+    assert_not_equal true, actual_facts['is_hostname_obfuscated']
+  end
+
+  test 'generates obfuscated ip_address fields with insights-client' do
```
jeremylenz:
```suggestion
  test 'generates obfuscated ip_address fields when insights-client facts are present' do
```

test/unit/slice_generator_test.rb
```diff
@@ -449,8 +474,16 @@ def fact_names
   end
 
   test 'obfuscates fqdn when instructed by insights-client' do
```
jeremylenz:
```suggestion
  test 'obfuscates fqdn when insights-client facts are present' do
```

test/unit/slice_generator_test.rb
```diff
@@ -487,9 +520,35 @@ def fact_names
     assert_equal 1, generator.hosts_count
   end
 
+  test 'obfuscates fqdn on host with insights-client when setting set' do
```
jeremylenz:
```suggestion
  test 'obfuscates host fqdn with insights-client when setting set' do
```

test/unit/slice_generator_test.rb
```diff
@@ -487,9 +520,35 @@ def fact_names
...
+    assert_equal '00000000-0000-0000-0000-000000000000', actual['report_slice_id']
+    assert_not_nil(actual_host = actual['hosts'].first)
+    assert_equal obfuscated_fqdn, actual_host['fqdn']
+    assert_equal '1234', actual_host['account']
+    assert_not_nil(actual_facts = actual_host['facts'].first['facts'])
+    assert_equal true, actual_facts['is_hostname_obfuscated']
+    assert_equal 1, generator.hosts_count
+  end
+
   test 'does not obfuscate fqdn when insights-client sets to false' do
```
jeremylenz:
```suggestion
  test 'does not obfuscate fqdn when host fact from insights-client has a value of false' do
```

jeremylenz (commented):
Some changes to comments and descriptions.

Also would love to see some more low level unit tests testing methods

obfuscate_hostname?
fqdn
obfuscate_ips?
obfuscate_ip

Will leave that up to you whether to do it now or in a followup.

chris1984:
@jeremylenz updated, kept the comment removal and new tests in a separate commit so it's easier to read. When I merge I will squash them.

lib/foreman_inventory_upload/generators/fact_helpers.rb
```diff
@@ -57,53 +57,112 @@ def cloud_provider(host)
...
+        return true if Setting[:obfuscate_inventory_ips]
 
-        Setting[:obfuscate_inventory_ips]
+        insights_client_ipv4_setting = fact_value(host, 'insights_client::obfuscate_ipv4_enabled')
+        insights_client_ipv6_setting = fact_value(host, 'insights_client::obfuscate_ipv6_enabled')
+
+        cast_ipv4_setting = ActiveModel::Type::Boolean.new.cast(insights_client_ipv4_setting)
+        cast_ipv6_setting = ActiveModel::Type::Boolean.new.cast(insights_client_ipv6_setting)
+
+        # 2. The host's IPv4 or IPv6 obfuscation setting is true
```
jeremylenz:
```suggestion
        # 2. The host's IPv4 or IPv6 obfuscation fact value is true
```

lib/foreman_inventory_upload/generators/fact_helpers.rb
```diff
@@ -57,53 +57,112 @@ def cloud_provider(host)
...
+        # 3. If neither of the above, don't obfuscate.
+        cast_ipv4_setting || cast_ipv6_setting || false
       end
 
       def host_ips(host)
+        # Determines and returns the IP addresses associated with a host, applying obfuscation if enabled.
+        # If IP obfuscation is enabled for the host return a representation of obfuscated IP addresses.
```
jeremylenz: Adding a blank line between these will make it more clear that the first comment describes the method overall, and the second describes what's below it.

chris1984: Fixed

lib/foreman_inventory_upload/generators/fact_helpers.rb
chris1984: Fixed

jeremylenz (approved):
I think this one's been through enough. Let's get it in.

@chris1984 please rebase and then I'll merge

PR #1003, opened 2025-06-18
Patch Katello controllers to enable access by smart proxy

ShimShtein:

Patching the Katello controllers to enable access from an IoP smart proxy

The other option is to implement an RBAC scheme similar to the `User` object

Access the Katello API using cert based auth with a cert that belongs to a smart proxy

## Summary by Sourcery

Patch Katello API controllers to allow Smart Proxy access via certificate authentication, introduce feature listing for on-prem proxies, update proxy-setting tests for on-prem scenarios, and seed the corresponding features in the database.

Enhancements:
- Enable certificate-based Smart Proxy authentication for Katello Organizations and Repositories API by patching controllers with SmartProxyAuth and custom filters
- Introduce on_prem_smart_proxy_features helper to list allowed Smart Proxy features

Tests:
- Adjust HTTP proxy tests to stub on-prem advisor engine detection and add test for empty proxy setting in on-prem mode

Chores:
- Add database seed script to create on-prem Smart Proxy features

### DIFF
--- a/db/seeds.d/200_features.rb
+++ b/db/seeds.d/200_features.rb
@@ -1,0 +1,4 @@
+ForemanRhCloud.on_prem_smart_proxy_features.each do |feature_name|
+  f = Feature.where(:name => feature_name).first_or_create
+  raise "Unable to create proxy feature: #{SeedHelper.format_errors f}" if f.nil? || f.errors.any?
+end
--- a/lib/foreman_rh_cloud/engine.rb
+++ b/lib/foreman_rh_cloud/engine.rb
@@ -210,6 +210,30 @@ module ForemanRhCloud
       )
     end
 
+    # Ideally this code belongs to an initializer. The problem is that Katello controllers are not initialized completely until after the end of the to_prepare blocks
+    # This means I can patch the controller only in the after_initialize block that is promised to run after the to_prepare
+    # initializer 'foreman_rh_cloud.allow_smart_proxy_actions', :before => :finisher_hook, :after => 'katello.register_plugin'  do |_app|
+    # end
+    config.after_initialize do
+      # skip overrides in migrations, since the controller initialization depends on tables existense
+      if defined?(Katello) && !Foreman.in_setup_db_rake?
+        Katello::Api::V2::OrganizationsController.include Foreman::Controller::SmartProxyAuth
+        # patch the callbacks order for :download_debug_certificate, since local_find_taxonomy has to run after the user is already initialized
+        Katello::Api::V2::OrganizationsController.skip_before_action(:local_find_taxonomy, only: :download_debug_certificate)
+        Katello::Api::V2::OrganizationsController.add_smart_proxy_filters(
+          [:index, :download_debug_certificate],
+          features: ForemanRhCloud.on_prem_smart_proxy_features
+        )
+        Katello::Api::V2::OrganizationsController.before_action(:local_find_taxonomy, only: :download_debug_certificate)
+
+        Katello::Api::V2::RepositoriesController.include Foreman::Controller::SmartProxyAuth
+        Katello::Api::V2::RepositoriesController.add_smart_proxy_filters(
+          :index,
+          features: ForemanRhCloud.on_prem_smart_proxy_features
+        )
+      end
+    end
+
     rake_tasks do
       Rake::Task['db:seed'].enhance do
         ForemanRhCloud::Engine.load_seed
@@ -230,5 +254,9 @@ module ForemanRhCloud
     else
       ::SETTINGS[:ssl_ca_file]
     end
+  end
+
+  def self.on_prem_smart_proxy_features
+    ['insights']
   end
 end
--- a/lib/foreman_rh_cloud/engine.rb
+++ b/lib/foreman_rh_cloud/engine.rb
@@ -257,6 +257,6 @@ module ForemanRhCloud
   end
 
   def self.on_prem_smart_proxy_features
-    ['insights']
+    ['Insights']
   end
 end
--- a/test/controllers/insights_cloud/api/cloud_request_controller_test.rb
+++ b/test/controllers/insights_cloud/api/cloud_request_controller_test.rb
@@ -51,8 +51,7 @@ module InsightsCloud::Api
       mock_composer = mock('composer')
       ::JobInvocationComposer.expects(:for_feature).with do |feature, host_ids, params|
         feature == :rh_cloud_connector_run_playbook &&
-        host_ids.first == host1.id &&
-        host_ids.last == host2.id
+        host_ids.sort == [host1.id, host2.id].sort
       end.returns(mock_composer)
       mock_composer.expects(:trigger!)
       mock_composer.expects(:job_invocation)
--- a/test/unit/rh_cloud_http_proxy_test.rb
+++ b/test/unit/rh_cloud_http_proxy_test.rb
@@ -4,6 +4,7 @@ class RhCloudHttpProxyTest < ActiveSupport::TestCase
   setup do
     @global_content_proxy_mock = 'http://global:content@localhost:80'
     @global_foreman_proxy_mock = 'http://global:foreman@localhost:80'
+    ForemanRhCloud.stubs(:with_local_advisor_engine?).returns(false)
   end
 
   test 'selects global content proxy' do
@@ -16,6 +17,13 @@ class RhCloudHttpProxyTest < ActiveSupport::TestCase
     setup_global_foreman_proxy
 
     assert_equal @global_foreman_proxy_mock, ForemanRhCloud.proxy_setting
+  end
+
+  test 'returns empty string in on-prem setup' do
+    ForemanRhCloud.unstub(:with_local_advisor_engine?)
+    ForemanRhCloud.stubs(:with_local_advisor_engine?).returns(true)
+
+    assert_empty ForemanRhCloud.proxy_setting
   end
 
   def setup_global_content_proxy


lib/foreman_rh_cloud/engine.rb
```diff
@@ -210,6 +210,17 @@ def self.register_scheduled_task(task_class, cronline)
...
     end
 
+    # Ideally this code belongs to an initializer. The problem is that Katello controllers are not initialized completely until after the end of the to_prepare blocks
+    # This means I can patch the controller only in the after_initialize block that is promised to run after the to_prepare
+    # initializer 'foreman_rh_cloud.allow_smart_proxy_actions', :before => :finisher_hook, :after => 'katello.register_plugin'  do |_app|
+    # end
+    config.after_initialize do
+      if defined?(Katello)
+        Katello::Api::V2::RepositoriesController.include Foreman::Controller::SmartProxyAuth
+        Katello::Api::V2::RepositoriesController.add_smart_proxy_filters(:index) # Add :features => proc { Foreman::Plugin.fact_importer_registry.fact_features }
```
ekohl: In this specific case we can expect to only have a single feature. Not sure what the name will be, but this plugin should also ensure it exists as part of a DB seed. Then it's sufficient to check for a single string instead of a proc from a registry.
```suggestion
        Katello::Api::V2::RepositoriesController.add_smart_proxy_filters(:index, features: 'My Feature')
```

db/seeds.d/200_features.rb
ekohl, line 1: Not a blocker, but I have thought about enhancing the Foreman plugin API to declare Smart Proxy features and let the seed in core solve it rather than duplicating the seeds in every plugin that declares features.

db/seeds.d/200_features.rb
```diff
@@ -0,0 +1,4 @@
+ForemanRhCloud.on_prem_smart_proxy_features.each do |feature_name|
+  f = Feature.where(:name => feature_name).first_or_create
+  raise "Unable to create proxy feature: #{format_errors f}" if f.nil? || f.errors.any?
```
ekohl: Is `format_errors` defined? It's defined [in katello](https://github.com/Katello/katello/blob/1a793acb30116340e3446cdb86610536386ca531/db/seeds.d/104-proxy.rb#L7-L10) but I'm not sure you can count on that.

In Foreman it uses `SeedHelper.format_errors` (https://github.com/theforeman/foreman/blob/42c1bdc37059f8cab8475008f48db7ab6427c5b3/db/seeds.d/110-smart_proxy_features.rb#L7)

lib/foreman_rh_cloud/engine.rb
```diff
@@ -210,6 +210,29 @@ def self.register_scheduled_task(task_class, cronline)
...
+        Katello::Api::V2::OrganizationsController.include Foreman::Controller::SmartProxyAuth
+        Katello::Api::V2::OrganizationsController.add_smart_proxy_filters(
+          :download_debug_certificate,
+          features: ForemanRhCloud.on_prem_smart_proxy_features
+        )
+        Katello::Api::V2::OrganizationsController.add_smart_proxy_filters(
+          :index,
+          features: ForemanRhCloud.on_prem_smart_proxy_features
+        )
+        Katello::Api::V2::RepositoriesController.include Foreman::Controller::SmartProxyAuth
```
ekohl, line 229: Perhaps visually separate the various controllers?
```suggestion

        Katello::Api::V2::RepositoriesController.include Foreman::Controller::SmartProxyAuth
```

lib/foreman_rh_cloud/engine.rb
```diff
@@ -210,6 +210,29 @@ def self.register_scheduled_task(task_class, cronline)
...
+    # end
+    config.after_initialize do
+      if defined?(Katello)
+        Katello::Api::V2::OrganizationsController.include Foreman::Controller::SmartProxyAuth
+        Katello::Api::V2::OrganizationsController.add_smart_proxy_filters(
+          :download_debug_certificate,
+          features: ForemanRhCloud.on_prem_smart_proxy_features
+        )
+        Katello::Api::V2::OrganizationsController.add_smart_proxy_filters(
+          :index,
```
ekohl: The label says `actions` and it's passed to `before_action`'s `only`. https://guides.rubyonrails.org/action_controller_overview.html#before-action has an example with multiple symbols so would this work:
```suggestion
          %i[download_debug_certificate index],
```

db/seeds.d/200_features.rb
ShimShtein: Created https://projects.theforeman.org/issues/38515 to track it. It will also solve the `before_action` ordering problem I am describing later.

lib/foreman_rh_cloud/engine.rb
```diff
@@ -210,6 +210,29 @@ def self.register_scheduled_task(task_class, cronline)
...
 
+    # Ideally this code belongs to an initializer. The problem is that Katello controllers are not initialized completely until after the end of the to_prepare blocks
+    # This means I can patch the controller only in the after_initialize block that is promised to run after the to_prepare
+    # initializer 'foreman_rh_cloud.allow_smart_proxy_actions', :before => :finisher_hook, :after => 'katello.register_plugin'  do |_app|
+    # end
+    config.after_initialize do
+      if defined?(Katello)
+        Katello::Api::V2::OrganizationsController.include Foreman::Controller::SmartProxyAuth
+        # patch the callbacks order for :download_debug_certificate, since local_find_taxonomy has to run after the user is already initialized
+        Katello::Api::V2::OrganizationsController.skip_before_action(:local_find_taxonomy, only: :download_debug_certificate)
```
ShimShtein, line 222: The suggested solution for https://projects.theforeman.org/issues/38515 will also solve this patching.

ShimShtein (commented):
@ekohl Fixed your comments and created a proper ticket to track the plugin interface enhancement.

jeremylenz:
How do I test this? How do I get a smart proxy that provides this service?

ShimShtein:
@jeremylenz I have introduced the smart proxy interface to the IoP gateway in https://github.com/RedHatInsights/iop-gateway/commit/6ef809c35ade57d3b36f6210881bb78a90289631. You need to spin that container, and try to register it in your Foreman

ekohl:
> @jeremylenz I have introduced the smart proxy interface to the IoP gateway in [RedHatInsights/iop-gateway@6ef809c](https://github.com/RedHatInsights/iop-gateway/commit/6ef809c35ade57d3b36f6210881bb78a90289631). You need to spin that container, and try to register it in your Foreman

This doesn't implement `/version` which the status code calls: https://github.com/theforeman/foreman/blob/6ca51b16316625e1974995ce82d90a55893578b8/app/models/smart_proxy.rb#L144.

In hindsight I regret not adding the top level version info in `/v2/features` so we could deprecate the old endpoints.

test/unit/rh_cloud_http_proxy_test.rb
```diff
@@ -18,6 +19,13 @@ class RhCloudHttpProxyTest < ActiveSupport::TestCase
     assert_equal @global_foreman_proxy_mock, ForemanRhCloud.proxy_setting
   end
 
+  test 'returns empty string in on-prem setup' do
+    ForemanRhCloud.unstub(:with_local_advisor_engine?)
+    ForemanRhCloud.stubs(:with_local_advisor_engine?).returns(true)
+
+    assert_equal '', ForemanRhCloud.proxy_setting
```
ekohl: Isn't this a better assertion?
```suggestion
    assert_empty ForemanRhCloud.proxy_setting
```

lib/foreman_rh_cloud/engine.rb
```diff
@@ -210,6 +210,31 @@ def self.register_scheduled_task(task_class, cronline)
       )
     end
 
+    # Ideally this code belongs to an initializer. The problem is that Katello controllers are not initialized completely until after the end of the to_prepare blocks
+    # This means I can patch the controller only in the after_initialize block that is promised to run after the to_prepare
+    # initializer 'foreman_rh_cloud.allow_smart_proxy_actions', :before => :finisher_hook, :after => 'katello.register_plugin'  do |_app|
+    # end
+    config.after_initialize do
+      # skip overrides in migrations, since the controller initialization depends on tables existense
+      if defined?(Katello) && ActiveRecord::Base.connection.data_source_exists?(SmartProxy.table_name)
```
ekohl: Perhaps this is cleaner:
```suggestion
      if defined?(Katello) && !Foreman.in_setup_db_rake?
```

It specifically checks for the setup tasks:
https://github.com/theforeman/foreman/blob/f4246cdaca1d0039d93de65cda9e52862c02f81b/lib/foreman.rb#L25-L27

Also, isn't Katello a hard dependency? It has `require 'katello'` as the first line so I'd think `defined?(Katello)` is always true.

Though I'm also wondering why this already needs DB access. What actually loads the DB? That's a bug in itself.

ShimShtein:
Got it working. Ready for review

lib/foreman_rh_cloud/engine.rb
ShimShtein: > Though I'm also wondering why this already needs DB access. What actually loads the DB? That's a bug in itself.

I am getting the following call stack if I disable the check:

```
ActiveRecord::StatementInvalid: PG::UndefinedTable: ERROR:  relation "smart_proxies" does not exist
...
/home/vagrant/foreman/app/models/smart_proxy.rb:44:in `block in <class:SmartProxy>'
...
/home/vagrant/katello/app/models/katello/concerns/smart_proxy_extensions.rb:88:in `pulp_primary'
/home/vagrant/katello/app/services/katello/repository_type_manager.rb:13:in `pulp_primary'
/home/vagrant/katello/app/services/katello/repository_type_manager.rb:34:in `enabled_repository_types'
/home/vagrant/katello/app/services/katello/repository_type_manager.rb:53:in `generic_repository_types'
/home/vagrant/katello/app/controllers/katello/api/v2/repositories_controller.rb:505:in `<class:RepositoriesController>'
```
If I understand the error correctly, the controller assumes that the database is already created and makes some DB queries directly from the module code. I think the proper solution will work around it, since the checks will be dynamic.

ekohl: > If I understand the error correctly, the controller assumes that the database is already created and makes some DB queries directly from the module code. I think the proper solution will work around it, since the checks will be dynamic.

Hmm, I think we should make that code safe.

For context, it's https://github.com/Katello/katello/blob/35895ec09cbeaaa7881569c8d3a0ef526e3bb21d/app/controllers/katello/api/v2/repositories_controller.rb#L505-L510

The background is that the supported repository types is dynamically determined based on the content types Pulp supports. Those are stored as capabilities on the Smart Proxy with the Pulp feature.

@ofedoren do you have any suggestions on how to improve that pattern? Effectively make the params lazy loaded.

ShimShtein: I have to admit I don't like the fact that I reference the controllers so early in the process. As I have said, in the proper solution, we won't need to load the controller class at all.
Ideally I would like to mark something like this:
``` ruby
plugin do
  enable_proxy_auth_for controller: 'foo', action: 'bar'
``

File: lib/foreman_rh_cloud/engine.rb [hunk 8]
```diff
@@ -210,6 +210,31 @@ def self.register_scheduled_task(task_class, cronline)
...
+        )
+        Katello::Api::V2::OrganizationsController.before_action(:local_find_taxonomy, only: :download_debug_certificate)
+
+        Katello::Api::V2::RepositoriesController.include Foreman::Controller::SmartProxyAuth
+        Katello::Api::V2::RepositoriesController.add_smart_proxy_filters(
+          :index,
+          features: ForemanRhCloud.on_prem_smart_proxy_features
+        )
+      end
+    rescue ActiveRecord::NoDatabaseError
```
ekohl: Isn't this line redundant with `Foreman.in_setup_db_rake?`?

ShimShtein: You are correct. It was a leftover. Thanks!

lib/foreman_rh_cloud/engine.rb
ofedoren: > @ofedoren do you have any suggestions on how to improve that pattern? Effectively make the params lazy loaded.

As of now I don't have any. I'm afraid we can't do much without introducing something new in apipie-rails lib. I'll play around with one rather nasty idea to see if it helps.

ofedoren: @ekohl, I think I've found a slight but bearable workaround: https://github.com/Katello/katello/pull/11430

UPD: I didn't find anything, a patch in `apipie-rails` is pending :/

lib/foreman_rh_cloud/engine.rb
```diff
@@ -231,4 +255,8 @@ def self.ca_cert
       ::SETTINGS[:ssl_ca_file]
     end
   end
+
+  def self.on_prem_smart_proxy_features
+    ['insights']
```
jeremylenz: Since users will see it in the UI, should it be this?
```suggestion
    ['red_hat_lightspeed']
```

ekohl: Will the on premise version also be called lightspeed?

vkrizan: I'd keep it insights, as that would be the upstream name.

Another note, most of the features have first letter capitalized: -> `Insights`.

test/controllers/insights_cloud/api/cloud_request_controller_test.rb
```diff
@@ -50,9 +50,11 @@ class CloudRequestControllerTest < ActionController::TestCase
 
       mock_composer = mock('composer')
       ::JobInvocationComposer.expects(:for_feature).with do |feature, host_ids, params|
+        actual_ids = host_ids.sort
+        expected_ids = [host1.id, host2.id].sort
         feature == :rh_cloud_connector_run_playbook &&
-        host_ids.first == host1.id &&
-        host_ids.last == host2.id
+        actual_ids.first == expected_ids.first &&
+        actual_ids.last == expected_ids.last
```
ekohl: Would it be better to match the whole array instead?
```suggestion
        expected_ids == actual_ids
```

You can then also optimize it to avoid the additional variables
```suggestion
        [host1.id, host2.id].sort == host_ids.sort
```

jeremylenz: In Katello I added a helper that does that, `assert_equal_arrays`. guess we can't use it here though.

ShimShtein: I would prefer to avoid cross-referencing plugins.

ekohl: > In Katello I added a helper that does that, `assert_equal_arrays`. guess we can't use it here though.

This actually isn't an assertion. I made the same mistake first.

jeremylenz (approved):
thanks @ShimShtein! Let's give it a go.

jeremylenz (commented):
Screenshot 2025-07-03 at 1 22 08 PM

Confirmed the Insights feature is properly created via seeds.

ianballou:
This PR is causing User.current to be nil only when a Product ID is passed in for some reason in Katello. This causes a product's repositories page to be inaccessible. I just tested varying foreman_rh_cloud commits with Katello to verify.

chris1984:
> This PR is causing User.current to be nil only when a Product ID is passed in for some reason in Katello. This causes a product's repositories page to be inaccessible. I just tested varying foreman_rh_cloud commits with Katello to verify.

Do you know what the fix would be?

ianballou:
> > This PR is causing User.current to be nil only when a Product ID is passed in for some reason in Katello. This causes a product's repositories page to be inaccessible. I just tested varying foreman_rh_cloud commits with Katello to verify.
> 
> Do you know what the fix would be?

I _think_ foreman_rh_cloud needs to ensure that `add_smart_proxy_filters` creates the before_action call to require_smart_proxy_or_login before the before_action calls in the repositories controller, not after.

PR #1014, opened 2025-07-03
Update Insights branding to Lightspeed

ColeHiggins2:
As part of the re-branding effort we are now calling `Red Hat Insights` => `Red Hat Lightspeed`.

Linked PRs https://github.com/theforeman/foreman/pull/10607
https://github.com/RedHatSatellite/foreman_theme_satellite/pull/198

## Summary by Sourcery

Update Red Hat Insights branding elements and fix a typo in playbook task description

Bug Fixes:
- Correct typo in scheduled task description from 'genrated' to 'generated'

Enhancements:
- Add 'Insights' title metadata for the RH Cloud UI fill


webpack/ForemanRhCloudFills.js
```diff
@@ -15,7 +15,7 @@ const fills = [
   },
   {
     slot: 'host-details-page-tabs',
-    name: 'Insights',
+    name: __('Insights'),
```
jeremylenz: At the top of the file you need to import __ as translate (check elsewhere in foreman_rh_cloud for the exact syntax and just copy it.)

app/models/insights_client_report_status.rb
```diff
@@ -8,7 +8,7 @@ class InsightsClientReportStatus < HostStatus::Status
   scope :reporting, -> { where(status: REPORTING) }
 
   def self.status_name
-    N_('Insights')
+    _('Insights')
```
jeremylenz: Would be good to have a comment explaining why this is inconsistent, since the other status names still use N_(). Or, maybe it's better just to change all of them? Thoughts?

webpack/ForemanRhCloudFills.js
ShimShtein: Please don't change the `name` property. You can add a `title` under `metadata`: https://github.com/theforeman/foreman/blob/770e8cc3527a45a76358bf529f17dab63145fdf9/webpack/assets/javascripts/react_app/components/HostDetails/index.js#L267
The `name` is used as a key.

app/models/insights_client_report_status.rb
ColeHiggins2: ill add a comment for the ones I changed. Im sure there was a reason to set the others as N() right?

jeremylenz: My (very new) understanding is that if there's an `N_()`, there must also be a `_()` or `__()` somewhere else, e.g. `N_()` is for when you need to split up the marking and the actual translating. `N_()` returns the original string but puts it on a list that some human actually goes thru and translates, and the other two both put it on that list and return the translated string.

I guess if nothing is broken it's fine? but @ShimShtein or @ekohl will know better

ekohl: > My (very new) understanding is that if there's an N_(), there must also be a _() or __() somewhere else, e.g. N_() is for when you need to split up the marking and the actual translating. N_() returns the original string but puts it on a list that some human actually goes thru and translates, and the other two both put it on that list and return the translated string.

This is correct.

Let's take `to_label` below as an example of how you could rewrite it:

```ruby
class InsightsClientReportStatus < HostStatus::Status
  LABELS = {
    REPORTING => N_('Reporting'),
    NO_REPORT => N_('Not reporting'),
  }.freeze

  def to_label(_options = {})
    LABELS[status]
  end
end
```

Now `to_label` is special because we know that later the label is translated.

Effectively when you assign a value to a constant that you will later run through `_()` then you want to use `N_()`. If you show the value directly then you want to use `_()`.

In fact, if you look at the `HostStatus::Global` you can see it's done exactly like this: https://github.com/theforeman/foreman/blob/ef417a172b3681e5670d3e268f1e7bc3178936eb/app/models/host_status/global.rb#L7-L11 & https://github.com/theforeman/foreman/blob/ef417a172b3681e5670d3e268f1e7bc3178936eb/app/models/host_status/global.rb#L37-L39

I'm not saying you always need to rewrite it to this form, but one advantage of using constants is that you don't allocate a new string every method invocation. The current `to_label` will create a new instance every time (until frozen strings become the default, see https://www.prateekcodes.dev/ruby-34-frozen-string-literals-rails-upgrade-guide/).

Now `to_label` is a good case that needs to return 

See https://projects.theforeman.org/projects/foreman/wiki/Translating#Extracting-strings as well.

What I now see in the same `HostStatus::Global` is that the `status_name` there is also using `N_()`: https://github.com/theforeman/foreman/blob/ef417a172b3681e5670d3e268f1e7bc3178936eb/app/models/host_status/global.rb#L25-L27

I think one example of where it is translated is here: https://github.com/theforeman/foreman/blob/ef417a172b3681e5670d3e268f1e7bc3178936eb/app/helpers/hosts_helper.rb#L101

So in short, I think the changes in this file shouldn't be done.

lib/foreman_rh_cloud/engine.rb
```diff
@@ -191,14 +191,14 @@ def self.register_scheduled_task(task_class, cronline)
     config.to_prepare do
       RemoteExecutionFeature.register(
         :rh_cloud_remediate_hosts,
-        N_('Apply Insights recommendations'),
+        _('Apply Insights recommendations'),
```
ekohl: This should only mark them as well. During `config.to_prepare` you don't know the user's language yet so you can't translate it. The end result is likely that it won't be translated for anyone if you change this.

ekohl (changes requested):

lib/foreman_rh_cloud/engine.rb
jeremylenz: Our goal here is foreman_theme_satellite picking it up, not actual translation. You're saying this will break actual translations?

ekohl: Yes, this is wrong and the old code was right. If it's not being picked up then we should dive deeper.

ColeHiggins2:
Linked PRs  https://github.com/theforeman/foreman/pull/10607
https://github.com/RedHatSatellite/foreman_theme_satellite/pull/198

ColeHiggins2:
@ekohl @jeremylenz @ShimShtein  Take a look at this solution and let me know what you think!  PR linked above

jeremylenz (commented):
Needs a lint fixup but other than that LGTM :+1:

jeremylenz:
also please update the PR description, this one's been thru a lot and is no longer accurate

ekohl (commented):
Without testing it this reads correct to me, though I'm not great with the frontend.

ShimShtein (approved):

PR #998, opened 2025-06-06
load scalprum

MariaAga:
To setup:
1. Create  /etc/httpd/conf.d/05-foreman-ssl.d/consoledot.conf      
with `ProxyPass /scalprum http://HOSTNAME:8001` (change HOSTNAME)
2. run `sudo systemctl restart httpd`
3. and run `podman run -p 8001:8000 quay.io/redhat-services-prod/rh-platform-experien-tenant/landing-page-frontend:latest`

4. Also needs https://github.com/theforeman/foreman/pull/10342

This pr deletes the use of frontend components as they clash with scalprum and should be replaces in https://github.com/theforeman/foreman_rh_cloud/issues/973

## Summary by Sourcery

Enable Scalprum-based module federation by introducing a Scalprum context and provider wrapper for dynamic plugin loading, and adding the necessary Scalprum dependencies.

New Features:
- Add ScalprumContext and ScalprumContextWrapper components for managing runtime plugin configuration and loading.

Enhancements:
- Add "@scalprum/react-core" and "@scalprum/core" to project dependencies.
- Configure plugin SDK options to transform plugin manifests based on CDN paths.
- Provide a mock user API for ScalprumProvider authentication integration.

PR #1019, opened 2025-07-07
Integrate CVE list component into CVE list page

nofaralfasi:
Requires: https://github.com/theforeman/foreman_rh_cloud/pull/998

## Summary by Sourcery

Integrate the CVE list component into the Vulnerability page by setting up a Scalprum context for dynamic module federation and loading the remote landing app widget within the page layout.

New Features:
- Introduce a ScalprumContextWrapper and context for dynamic federated module configuration
- Embed the remote 'landing' app's AnsibleWidget (CVE list) into the Vulnerability page via ScalprumComponent

Enhancements:
- Refactor the InsightsVulnerability component to initialize Scalprum config on mount and wrap it with the context provider
- Switch the default export of the vulnerability page to the wrapped component for federation support

Build:
- Add @scalprum/react-core and @scalprum/core to package.json for module federation dependencies


webpack/common/ScalprumModule/ScalprumContext.js
```diff
@@ -0,0 +1,115 @@
...
+                return manifest;
+              },
+            },
+          }}
+          api={{
+            chrome: {
+              isBeta: () => false,
+              on: () => {},
+              auth: {
+                getUser: () => Promise.resolve(mockUser),
```
vkrizan: Is this mocked user required?

webpack/common/ScalprumModule/ScalprumContext.js
```diff
@@ -0,0 +1,115 @@
...
+  {config[scope] && (
+          <ScalprumComponent scope={scope} module={module}` />
+        )}
+*/
+
+export const ScalprumContextWrapper = ({ children }) => {
+  const [config, setConfig] = useState({});
+  const contextData = {
+    config,
+    setConfig: useCallback(
```
vkrizan: Does the context need to be dynamic, requiring the `setConfig`? Isn't the context stable? cc @hyperkid123

Hyperkid123: No it can be static. In fact it is better if its static.

MariaAga: How would you suggest us using a generic component?

webpack/common/ScalprumModule/ScalprumContext.js
nofaralfasi: I believe this question is best addressed in the PR where it was originally introduced: https://github.com/theforeman/foreman_rh_cloud/pull/998

nofaralfasi: I used this code from Maria's PR #998, so she might be the best person to answer this. 
cc @MariaAga

MariaAga: And I got it from here: https://github.com/Hyperkid123/scalprum-staellite-poc/blob/main/src/boostrap.js#L64

webpack/common/ScalprumModule/ScalprumContext.js
Hyperkid123: Is it a big deal if the ScalprumProvider has a full config with all used modules? Even though it might not be used in a particular page? I don't think so. So you can feed the metadata to the provider directly as a prop and not have to do this round trip.

webpack/common/ScalprumModule/ScalprumContext.js
Hyperkid123: The data within the API depends on the needs of the HCC modules used in Foreman. I don't think that a mock user is useful because the data is literally just a mock.

ShimShtein: For now we don't actually need to pass the user, so it is OK to have a mock there.
The authorization is done on the Foreman side anyway

webpack/common/ScalprumModule/ScalprumContext.js
ShimShtein: If we start using scalprum in multiple plugins, we won't have a single list of modules. For now it can be OK, but it would become our tech debt

webpack/InsightsVulnerability/InsightsVulnerability.js
```diff
@@ -1,13 +1,45 @@
...
+  const module = './AnsibleWidget'
+  const newConfig = {
+    [scope]: {
+      name: scope,
+      manifestLocation: `${window.location.origin}/scalprum/${cdnPath}/fed-mods.json`,
+      cdnPath: `${window.location.origin}/scalprum/${cdnPath}/`,
+    },
+  };
+  useEffect(() => {
+    setConfig(newConfig);
```
vkrizan: Changing the state of a context from bottom levels is generally discouraged. Ideally we could have the configuration on one place and static.

webpack/common/ScalprumModule/ScalprumContext.js
MariaAga: I think we can also just use scalprum directly instead of this wrapper,  like here: https://github.com/Hyperkid123/scalprum-staellite-poc/blob/main/src/boostrap.js

I made the wrapper as it was requested to simplify it & generalize it in case of mass config changes or something

ShimShtein: I would expect an inheritance for the `ScalprumProvider` and direct usage of `ScalprumComponent`.

So we will have something like `InsightsScalprumProvider` that will set the common properties like the user or the isBeta flag. It will receive a configuration object that will be merged into the defaults.
So from dev's point of view the code will look like:
``` js
const App = () => {
  return (
    <div>
      <BrowserRouter>
        <InsightsScalprumProvider
          api={{
            mySpecificFunc: () => foo,
          }}
          config={{
             mySpecificConfigFlag: 123,
          }}
        >
          <ChildC />
        </InsightsScalprumProvider>
      </BrowserRouter>
    </div>
  )
}
```
And the `InsightsScalprumProvider` will do something like:

```js
const finalApi = api={{
            chrome: {
              isBeta: () => false,
              on: () => console.log,
              auth: {
                getUser: () => Promise.resolve(mockUser)
              }
            }
          ...userApi
          }}
```
And so on.

webpack/InsightsVulnerability/InsightsVulnerability.js
```diff
@@ -1,13 +1,45 @@
...
 
-const InsightsVulnerability = () => (
-  <PageLayout searchable={false} header={__('Vulnerability')}>
-    <div className="insights-vulnerability">
-      <p>This page is under development. Please check back soon for updates.</p>
+const InsightsVulnerability = () => {
+  const { config, setConfig } = React.useContext(ScalprumContext);
+  const cdnPath = 'apps/landing'
+  const scope = 'landing'
+  const module = './AnsibleWidget'
```
vkrizan: The module should be `./CveListPage`.

webpack/InsightsVulnerability/InsightsVulnerability.js
```diff
@@ -1,13 +1,45 @@
...
+import { ScalprumContext, ScalprumContextWrapper } from '../common/ScalprumModule/ScalprumContext';
 
-const InsightsVulnerability = () => (
-  <PageLayout searchable={false} header={__('Vulnerability')}>
-    <div className="insights-vulnerability">
-      <p>This page is under development. Please check back soon for updates.</p>
+const InsightsVulnerability = () => {
+  const { config, setConfig } = React.useContext(ScalprumContext);
+  const cdnPath = 'apps/landing'
+  const scope = 'landing'
```
vkrizan: The scope should be `vulnerability`.

webpack/InsightsVulnerability/InsightsVulnerability.js
nofaralfasi: Thanks! I’ve updated everything. This component was only used for testing purposes, as the CveListPage isn't functional in this code yet.

webpack/common/ScalprumModule/ScalprumContext.js
```diff
@@ -0,0 +1,79 @@
...
+export const ScalprumContextWrapper = ({ children }) => {
+  const config = {
+    'vulnerability': {
+      name: 'vulnerability',
+      manifestLocation: `${window.location.origin}/pub/assets/apps/vulnerability/fed-mods.json`,
+      cdnPath: `${window.location.origin}/pub/assets/apps/vulnerability/`,
+    },
+  };
+
+  const scStore = useMemo(() => initialize(        {pluginSDKOptions: {
```
vkrizan: It looks like this was added from what Martin has written out as a potential quick fix for the Scalprum initialization. I think the initialization would be done by ScalprumProvider.

webpack/common/ScalprumModule/ScalprumContext.js
```diff
@@ -0,0 +1,79 @@
...
+        is_active: 'boolean',
+        is_internal: 'boolean',
+        is_org_admin: 'boolean',
+        locale: 'string',
+      },
+    },
+  };
+  return (
+      <ScalprumProvider
+        scalprum={scStore}
```
MariaAga: https://github.com/scalprum/scaffolding/blob/8671c64a31e7c5f7ac9675fa95067cc60faf933c/packages/react-core/src/scalprum-provider.tsx#L19
ScalprumProvider doesnt have a scalprum props
why not use the props as in https://github.com/theforeman/foreman_rh_cloud/pull/998/files ?

vkrizan: See my comment https://github.com/theforeman/foreman_rh_cloud/pull/1019#discussion_r2200738177.

vkrizan: This can be reworked how it was before:
```
      <ScalprumProvider
        pluginSDKOptions= {{
          pluginLoaderOptions: {
            transformPluginManifest: manifest => {
            ...
        }}
        api={{
          chrome: {...
        }}
        config={config}
    >
```

webpack/common/ScalprumModule/ScalprumContext.js
Hyperkid123: Nono, the CDN transformation has to happen. It's not a built-in feature of Scalprum. I think I misunderstood your question.

vkrizan: I've meant the hardcoded `initialize()`. The CDN transformations can be passed through params to the `ScalprumProvider`, right?

Hyperkid123: Oh yeah, right. You are correct.

package.json
```diff
@@ -21,6 +21,10 @@
   "peerDependencies": {
     "@theforeman/vendor": ">= 15.0.1"
```
vkrizan, line 22: Does this vendor pgk need to be still included?

MariaAga: Great question, that we will test and check after the release probably 
Technically, no, hopefully, no

webpack/common/ScalprumModule/ScalprumContext.js
```diff
@@ -0,0 +1,73 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { ScalprumProvider } from '@scalprum/react-core';
+
+export const ScalprumContextWrapper = ({ children }) => {
+  const config = {
+    vulnerability: {
+      name: 'vulnerability',
+      manifestLocation: `${window.location.origin}/scalprum/apps/vulnerability/fed-mods.json`,
+      cdnPath: `${window.location.origin}/scalprum/apps/vulnerability/`,
```
vkrizan: I think the path would be `/assets/apps/vulnerability/`

webpack/InsightsVulnerability/InsightsVulnerability.js
```diff
@@ -1,13 +1,31 @@
...
+    <div className="rh-cloud-insights-vulnerability-page">
+      <ScalprumComponent scope={scope} module={module} />
+
+      <PageLayout searchable={false} header={__('Vulnerability')}>
+        <div className="insights-vulnerability">
+          <p>
+            This page is under development. Please check back soon for updates.
+          </p>
+        </div>
+      </PageLayout>
```
vkrizan: I think you can remove this "under development" memo with header.

nofaralfasi:
Closing this PR in favor of #1031, where the work has been completed.

PR #1021, opened 2025-07-08
Add Recommendations column

jeremylenz:

This adds a Recommendations column to the new All Hosts page.

- It should now be available to add in Manage Columns
- If you already have the column added on the legacy page, that table preference will carry over to the New UI page

If you're using local Insights advisor, the column will just show a placeholder text. I will soon replace that with a Scalprum component in a separate PR.

If you're using hosted Insights, the column will show the number of Insights hits with a link to the Recommendations page.

I could also link to the host details page Insights tab. Thoughts?

Get a host registered to Insights
Send the inventory upload report
Sync recommendations
Get at least one recommendation for your host
Add the column on Hosts > All Hosts > New UI

## Summary by Sourcery

Add a Recommendations column to the new All Hosts page, backed by API support and a React extension that displays placeholder text for local advisor or links to hosted Insights.

New Features:
- Expose insights_hits_count and use_local_advisor_engine in the hosts API for Insights integration
- Register a new Recommendations column in the Hosts table with a React cell that links to Insights or shows a placeholder

Enhancements:
- Carry over existing column visibility preferences from the legacy Hosts UI to the new interface

### DIFF
--- a/app/views/api/v2/hosts/insights/base.rabl
+++ b/app/views/api/v2/hosts/insights/base.rabl
@@ -3,3 +3,9 @@ attributes :uuid
 node :insights_hit_details do |facet|
   facet&.host&.facts('insights::hit_details')&.values&.first
 end
+node :insights_hits_count do |facet|
+  facet.hits&.count
+end
+node :use_local_advisor_engine do |_facet|
+  ForemanRhCloud.with_local_advisor_engine?
+end
--- a/webpack/ForemanColumnExtensions/index.js
+++ b/webpack/ForemanColumnExtensions/index.js
@@ -1,0 +1,41 @@
+import React from 'react';
+import { translate as __ } from 'foremanReact/common/I18n';
+import { propsToCamelCase } from 'foremanReact/common/helpers';
+
+const RecommendationsCell = hostDetails => {
+  const insightsAttributes = propsToCamelCase(
+    // eslint-disable-next-line camelcase
+    hostDetails?.insights_attributes ?? {}
+  );
+  // Local insights advisor
+  if (insightsAttributes.useLocalAdvisorEngine) {
+    // TODO: Replace this placeholder with the actual local advisor integration
+    return <span>Local advisor placeholder</span>;
+  }
+
+  // Hosted insights advisor
+  const { insightsHitsCount: hitsCount } = insightsAttributes;
+  if (hitsCount === undefined || hitsCount === null) return '—';
+  const hostname = hostDetails?.name;
+  const encodedHostname = encodeURIComponent(hostname);
+  const hitsUrl = `/foreman_rh_cloud/insights_cloud?search=hostname+%3D+${encodedHostname}`;
+  return <a href={hitsUrl}>{hitsCount}</a>;
+};
+
+const hostsIndexColumnExtensions = [
+  {
+    columnName: 'insights_recommendations_count',
+    title: __('Recommendations'),
+    wrapper: RecommendationsCell,
+    weight: 1500,
+    isSorted: true,
+  },
+];
+
+hostsIndexColumnExtensions.forEach(column => {
+  column.tableName = 'hosts';
+  column.categoryName = 'Insights';
+  column.categoryKey = 'insights';
+});
+
+export default hostsIndexColumnExtensions;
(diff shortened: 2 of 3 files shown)


app/views/api/v2/hosts/insights/base.rabl
```diff
@@ -3,3 +3,9 @@ attributes :uuid
 node :insights_hit_details do |facet|
   facet&.host&.facts('insights::hit_details')&.values&.first
 end
+node :insights_hits_count do |facet|
+  facet.hits&.count
+end
+node :use_local_advisor_engine do |_facet|
+  ForemanRhCloud.with_local_advisor_engine?
+end
```
nofaralfasi: Is there a specific reason it's needed? I tested the flow without it and everything seems to work fine on my end.
Wondering if I’m missing a use case or if it's required for something I haven’t run into yet.

jeremylenz: Right now it shows a placeholder, but I plan a follow-up right after this which will load a Scalprum component in that table cell so that IoP folks can insert their component there.

jeremylenz: To see it, you can add this to your Foreman `config/settings.yaml`

```
:foreman_rh_cloud:
  :use_local_advisor_engine: true
```

jeremylenz:
Addressed sourcery-ai suggestions & rebased.

webpack/ForemanColumnExtensions/index.js
```diff
@@ -0,0 +1,41 @@
...
+  return <a href={hitsUrl}>{hitsCount}</a>;
+};
+
+const hostsIndexColumnExtensions = [
+  {
+    columnName: 'insights_recommendations_count',
+    title: __('Recommendations'),
+    wrapper: RecommendationsCell,
+    weight: 1500,
+    isSorted: true,
```
chris1984, line 31: What determines if this is true or not? I see a bunch of them false in https://github.com/theforeman/foreman/blob/380cf587557fd05737ff938fec5ca01898da1716/webpack/assets/javascripts/react_app/components/HostsIndex/Columns/core.js#L41

webpack/global_index.js
```diff
@@ -1,7 +1,11 @@
+import { registerColumns } from 'foremanReact/components/HostsIndex/Columns/core';
 import { registerReducers } from './ForemanRhCloudReducers';
 import { registerFills } from './ForemanRhCloudFills';
 import { registerRoutes } from './ForemanRhCloudPages';
+import hostsIndexColumnExtensions from './ForemanColumnExtensions/index';
 
 registerReducers();
 registerFills();
 registerRoutes();
+
```
chris1984: Is there a reason we want an empty line here?

webpack/ForemanColumnExtensions/index.js
```diff
@@ -0,0 +1,41 @@
...
+    hostDetails?.insights_attributes ?? {}
+  );
+  // Local insights advisor
+  if (insightsAttributes.useLocalAdvisorEngine) {
+    // TODO: Replace this placeholder with the actual local advisor integration
+    return <span>Local advisor placeholder</span>;
+  }
+
+  // Hosted insights advisor
+  const hitsCount = insightsAttributes.insightsHitsCount;
```
chris1984: What do you think of? `const { insightsHitsCount: hitsCount } = insightsAttributes;`

webpack/ForemanColumnExtensions/index.js
jeremylenz: See how in this one `columnName` is `insights_recommendations_count`? That means that if `isSorted` is true,

1. there will be a sorting button on the column; and
2. when you click that button, it will make an API request with order by = "insights_recommendations_count". 

If `insights_recommendations_count` is a field on the model (in this case, InsightsHit), then on the backend, `InsightsHit.order(:insights_recommendations_count)` will work fine, and `isSorted` can be true.

If `insights_recommendations_count` is not a database field, for example if it's just a method, but does not appear as a database column, then  `InsightsHit.order(:insights_recommendations_count)` will not work, so `isSorted` should not be true.

chris1984: Makes more sense now, thanks for explaining it.

webpack/global_index.js
jeremylenz: removed

jeremylenz:
@chris1984 Updated :+1:

chris1984 (approved):
Code looks fine and works well, thanks for addressing the comments:

Screenshot 2025-07-14 193643
Screenshot 2025-07-14 193941

PR #1023, opened 2025-07-11
Release 11.4.2

Odilhao:
## Summary by Sourcery

Release version 11.4.2 and bypass proxy when using the local advisor engine

Enhancements:
- Bypass HTTP proxy by returning an empty string if running with the local advisor engine

Build:
- Bump version to 11.4.2 in version.rb and package.json


chris1984 (approved):
LGTM

PR #1029, opened 2025-07-16
SAT-34609 - Use shared GH action for JS testing

chris1984:
## Summary by Sourcery

Switch the JavaScript CI workflow to use the shared Foreman plugin action and broaden its triggers.

Enhancements:
- Replace custom Node.js setup, lint, and test steps with theforeman/actions/foreman_plugin_js.yml shared workflow
- Rename the job to 'JavaScript' and pass the plugin parameter to the shared action

CI:
- Add a push trigger for the develop branch and remove path filters on pull requests

### DIFF
--- a/.github/workflows/js_tests.yml
+++ b/.github/workflows/js_tests.yml
@@ -1,32 +1,17 @@
 name: JS
+
+permissions:
+  contents: read
+
 on:
+  push:
+    branches:
+      - develop
   pull_request:
-    paths:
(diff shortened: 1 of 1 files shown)


chris1984:
@evgeni does this look right?

evgeni (approved):
The "nodejs 22" part in the title is wrong -- we didn't switch to 22 in core yet, but once we do this action will automatically use it.

ACK

PR #1025, opened 2025-07-15
Integrate SystemDetailTable component into Host Details CVEs tab

nofaralfasi:
Tests still need to be fixed. 

## Summary by Sourcery

Add new CVEs tab and host-filtering capabilities, enhance obfuscation and cloud forwarding logic, inject package profile upload hooks, and bump version and dependencies

New Features:
- Add CVEsHostDetailsTab and InsightsVulnerability pages using Scalprum integration and placeholder layout
- Support host-level filtering in inventory report generation via hosts_filter parameter across jobs, rake tasks, archive naming, and queries
- Introduce a package_profile upload extension to trigger asynchronous host-specific report generation after package profile uploads in InsightsCloud
- Enhance CloudRequestForwarder to accept host context and include a Forwarded header with subscription uuid for telemetry requests
- Enable dynamic certificate source selection (manifest vs local advisor engine) in upload_report_job and cert_auth

Enhancements:
- Refactor fact_helpers to support hierarchical hostname and IPv4/IPv6 obfuscation using global settings and insights-client facts with JSON parsing and SHA1/IP sequencing fallbacks
- Skip inventory_hosts_sync execution for local advisor engine
- Remove obsolete katello_agent_running field from slice generator

Build:
- Bump foreman_rh_cloud package version to 12.1.3 and add @scalprum/react-core and @scalprum/core dependencies

Documentation:
- Update README with DeepWiki badge and project overview link

Tests:
- Extend and update unit tests for obfuscation logic, hosts_filter support, CloudRequestForwarder, certificate handling, and new page components

PR #1031, opened 2025-07-16
New Vulnerabilities list tab and page

ShimShtein:
It's based on https://github.com/theforeman/foreman_rh_cloud/pull/1025

It introduces CVEs list to host details as a new "Vulnerabiliy" tab.
It also introduces the CVE List page from https://github.com/theforeman/foreman_rh_cloud/pull/1019.

## Summary by Sourcery

Introduce a new Vulnerability tab on host details and a standalone Vulnerability page by integrating remote modules via Scalprum and provide a context wrapper for plugin configuration.

New Features:
- Add a new "Vulnerability" tab on the host details page that loads the SystemDetailTable module via Scalprum
- Add a standalone Vulnerability page that loads the CveListPage module via Scalprum
- Introduce a ScalprumContextWrapper to configure the vulnerability plugin environment with mocked user and manifest settings

Enhancements:
- Wrap CVE components with ScalprumContextWrapper and Foreman PageLayout for consistent layout and context
- Update fill registrations and routes to use the new wrapped components

Build:
- Add @scalprum/react-core and @scalprum/core as project dependencies


ShimShtein:
As for @MariaAga 's request got rid of the `ScalprumProviderWrapper` and simplified the code.
 @jeremylenz I think we can reuse the common scalprum properties and make sure we have all the paths e.t.c. in the same place. The same goes if we need changes to user definition or any other context.

vkrizan:
@ShimShtein can you please update the title and description to reflect that this adds CVE list as well? PR #1025 was closed in favor of this one.

package.json
```diff
@@ -21,6 +21,10 @@
   "peerDependencies": {
     "@theforeman/vendor": ">= 15.0.1"
```
vkrizan, line 22: https://github.com/theforeman/foreman_rh_cloud/pull/1019#discussion_r2207274071

webpack/InsightsVulnerability/InsightsVulnerability.js
```diff
@@ -1,13 +1,21 @@
...
+import { ScalprumComponent, ScalprumProvider } from '@scalprum/react-core';
+import { providerOptions } from '../common/ScalprumModule/ScalprumContext';
 
-const InsightsVulnerability = () => (
-  <PageLayout searchable={false} header={__('Vulnerability')}>
-    <div className="insights-vulnerability">
-      <p>This page is under development. Please check back soon for updates.</p>
+const InsightsVulnerability = () => {
+  const scope = 'vulnerability';
+  const module = './CveListPage';
```
vkrizan: This addition could have been in its own commit.

webpack/CVEsHostDetailsTab/__tests__/CVEsHostDetailsTab.test.js
```diff
@@ -1,18 +1,28 @@
...
     ).toBeTruthy();
-  });
-
-  it('renders the host name', () => {
-    const hostName = 'test-host.example.com';
-    render(<CVEsHostDetailsTab hostName={hostName} />);
-    expect(screen.getByText(`CVEs tab for host: ${hostName}`)).toBeTruthy();
+    // expect(
+    //   screen.getByText(/1/i, { exact: false})
+    // ).toBeInTheDocument();
```
vkrizan: Commented code

webpack/InsightsVulnerability/InsightsVulnerability.js
```diff
@@ -1,13 +1,21 @@
...
+  return (
+    <div className="rh-cloud-insights-vulnerability-page">
+      <ScalprumComponent scope={scope} module={module} />
     </div>
-  </PageLayout>
+  );
+};
+
+const InsightsVulnerabilityWrap = () => (
```
vkrizan: Ambiguous naming. There are there components that is being added as Insights Vulnerability. I would suggest to be more detailed in the component and file naming here. Examples: `CveListPageWrapper`, `InsightsCveListPage`, `InsightsVulnerabilityCveListPage`, `InsightsVulnerabilityCves`. cc @lesamo

webpack/CVEsHostDetailsTab/CVEsHostDetailsTab.js
```diff
@@ -1,17 +1,37 @@
...
+  );
+};
 
 CVEsHostDetailsTab.propTypes = {
-  hostName: PropTypes.string.isRequired,
+  systemId: PropTypes.string.isRequired,
+};
+
+const CVEsHostDetailsTabWrapper = ({ response }) => (
+  <ScalprumProvider {...providerOptions}>
```
vkrizan, line 21: We could have used the ScaplrumWrapper, instead of pulling constants. I think it would have fit better with a general composition.

vkrizan (commented):
Thank you. I've left a couple of comments.

webpack/ForemanRhCloudFills.js
```diff
@@ -33,10 +33,11 @@ const fills = [
   {
     slot: 'host-details-page-tabs',
     name: 'CVEs',
-    component: props => <CVEsHostDetailsTab {...props} />,
+    component: props => <CVEsHostDetailsTabWrapper {...props} />,
     weight: 300,
     metadata: {
       hideTab: vulnerabilityDisabled,
```
vkrizan, line 39: The tab is not showing for me. How do I debug it?

package.json
ShimShtein: As Maria said, this would be after the initial version is out

webpack/CVEsHostDetailsTab/CVEsHostDetailsTab.js
ShimShtein: I have refactored the code to stop using the generic provider based on Marias comment: https://github.com/theforeman/foreman_rh_cloud/pull/998#issuecomment-3070208195

webpack/ForemanRhCloudFills.js
ShimShtein: 1. Are there any errors in web console?
2. Take a look at the `/hosts/:id` API call, there is a flag that is passed there under the `vulnerability: enabled`            path.If it is passed as false, it means you need to set the advisor_local setting

ShimShtein:
Ready for another round!

webpack/ForemanRhCloudFills.js
vkrizan: There is no such key `vulnerability` on the API response. Probably I've registered the system w/o Insights enabled. How to enable it?

vkrizan: @jeremylenz ^ :pray:

jeremylenz: Looks like the `vulnerability` node is only sent with index and not show: https://github.com/theforeman/foreman_rh_cloud/blob/da91f30d605bc211a3b5f13089d824f9a75630c5/lib/foreman_rh_cloud/engine.rb#L131-L132

Need to add it to app/views/api/v2/hosts/insights/base.rabl.

jeremylenz: Another option is to have `vulnerabilityDisabled` look at the `:use_local_advisor_engine` node instead, which is already in the base rabl.

jeremylenz: From Shim: "We are adding it to API, specifically to a single rabl. The only thing that needs to exist, is the insights facet. If the facet is not there, we won't add the node. Now the facet should be created on each package upload. Maybe in Viliam's case the host was there before my package upload pr got in, so he doesn't have the insights facet on the host. The easiest way would be to sync hosts with a button. I think I have enabled this use case, and it will properly sync the ids"

webpack/ForemanRhCloudFills.js
```diff
@@ -33,10 +33,11 @@ const fills = [
...
+      title: __('Vulnerability'),
```
vkrizan: The tab name should be "Vulnerabilities". Please consult mocks.

webpack/CVEsHostDetailsTab/CVEsHostDetailsTab.js
```diff
@@ -1,17 +1,37 @@
...
+    </div>
+  );
+};
 
 CVEsHostDetailsTab.propTypes = {
-  hostName: PropTypes.string.isRequired,
+  systemId: PropTypes.string.isRequired,
+};
+
+const CVEsHostDetailsTabWrapper = ({ response }) => (
```
vkrizan, line 20: The tab content is missing a bounding box. Its contents are rendered very close to the edges.

image

webpack/ForemanRhCloudFills.js
vkrizan: I did a resycn with "Generate and upload report" but some hosts still don't show it. Interesting fact is that navigating from CVE counts form the hosts list table gets the tab shown, but it doesn't provide it a subscription/system id.

webpack/ForemanRhCloudFills.js
```diff
@@ -33,10 +33,11 @@ const fills = [
   {
     slot: 'host-details-page-tabs',
     name: 'CVEs',
```
vkrizan: If this is the tab name in hash, then this should be also updated to reflect the actual tab title ("Vulnerabilities").

vkrizan (approved):
Feel free to merge. The remaining issues can be done in a followup.

PR #1034, opened 2025-07-17
SAT-35946: Fix forwarding of requests with pipe | character

jeremylenz:
(cherry picked from commit 33dd32074fc8183d146e7018e9c4ba525d113deb)

Requests to Insights APIs via the Foreman_rh_cloud plugin are failing with HTTP 502 (error bad URI(is not URI?)) when the request URI contains a pipe symbol. Pipes are regularly used by Advisor within rule names.

## Summary by Sourcery

Bug Fixes:
- Prevent HTTP 502 errors when forwarding requests containing '|' by using original_fullpath to capture the path segment.

### DIFF
--- a/app/controllers/insights_cloud/ui_requests_controller.rb
+++ b/app/controllers/insights_cloud/ui_requests_controller.rb
@@ -7,9 +7,12 @@ module InsightsCloud
     # The method that "proxies" requests over to Cloud
     def forward_request
       begin
+        path_match = request.original_fullpath.match(%r{^/insights_cloud/(?<path>[^?]*)})&.[](:path)
+        path_to_forward = path_match || params.require(:path)
+
         @cloud_response = ::ForemanRhCloud::InsightsApiForwarder.new.forward_request(
           request,
-          params.require(:path),
+          path_to_forward,
           controller_name,
           User.current,
           @organization,


app/controllers/insights_cloud/ui_requests_controller.rb
```diff
@@ -9,7 +9,7 @@ def forward_request
       begin
         @cloud_response = ::ForemanRhCloud::InsightsApiForwarder.new.forward_request(
           request,
-          params.require(:path),
+          request.original_fullpath.match(%r{^/insights_cloud/(?<path>[^?]*)})[:path],
```
jeremylenz: Rails "helpfully" unescapes `params.require(:path)`, but we need to keep it escaped here so it doesn't get forwarded with the bad pipe character.

vkrizan: Can you please add some condition when the regex isn't matched, or is the `nil` handled somewhere down the line?

jeremylenz: Should I fall back to the original `params.require(:path)`?

vkrizan: How would that resolve it?

jeremylenz: Added safe navigation (sorry I didn't understand what you meant originally). That should address the nil when the regex isn't matched. Also added the fallback but not sure it's needed.

vkrizan (approved):

chris1984 (approved):
Looks fine to me, my only question is since we will be adding more services in the future, if we need to adjust the regex, should we put it in a variable instead and just reference the variable from the `forward_request` method? I am fine with what you have or that suggestion so going to ack as far as code goes. I have not tested it yet

```ruby
path_match= request.original_fullpath.match(%r{^/insights_cloud/(?<path>[^?]*)})&.[](:path)
path_to_forward = path_match || params.require(:path)

@cloud_response = ::ForemanRhCloud::InsightsApiForwarder.new.forward_request(
  request,
  path_to_forward,
  controller_name,
  User.current,
  @organization,
)
```

chris1984 (approved):

PR #1038, opened 2025-07-22
Fixes the RepositoriesController callback order

parthaa:
`RepositoriesController` called `find_product` before the user was initialized. This caused access denied errors. This fix orchestrates the call backs in a particular way so that find_product only gets called after the smart proxy user auth

- Create a Product and repo
- Try to run the following curl command replacing the product_id
- `curl -u admin:changeme 'http://localhost:3000/katello/api/products/<product_id>/repositories'`

Before PR
access denied perms errors

After PR
Correctly pulls the json.

## Summary by Sourcery

Bug Fixes:
- Ensure find_product and find_optional_organization run after SmartProxyAuth in the RepositoriesController index action


chris1984 (commented):
Looks fine to me, @jeremylenz want to take a look? I have not tested it yet

jeremylenz (approved):
Thanks @parthaa !

ACK

PR #1042, opened 2025-07-24
Add Advisor and Vulnerabilities to Foreman ping

jeremylenz:

Add a `ForemanRhCloud::Ping` model to `::Foreman::Ping.`

I am currently pinging only two URLs (see the code). Please let me know if there are any I should add, or change.

First you need a dev environment with with IoP services running.

You can see the list of services with
```
sudo foreman-maintain service status --brief
```

You should see all the IoP services:
```
$ sudo foreman-maintain service status --brief
Running Status Services
================================================================================
Get status of applicable services: 

Displaying the following service(s):
redis, postgresql, iop-core-engine, iop-core-gateway, iop-core-host-inventory, iop-core-host-inventory-api, iop-core-host-inventory-migrate, iop-core-ingress, iop-core-kafka, iop-core-puptoo, iop-core-yuptoo, iop-service-advisor-backend-api, iop-service-advisor-backend-service, iop-service-remediations-api, iop-service-vmaas-reposcan, iop-service-vmaas-webapp-go, iop-service-vuln-dbupgrade, iop-service-vuln-evaluator-recalc, iop-service-vuln-evaluator-upload, iop-service-vuln-grouper, iop-service-vuln-listener, iop-service-vuln-manager, iop-service-vuln-taskomatic, pulpcore-api, pulpcore-content, pulpcore-worker@1.service, pulpcore-worker@2.service, pulpcore-worker@3.service, pulpcore-worker@4.service, pulpcore-worker@5.service, pulpcore-worker@6.service, tomcat, httpd, foreman-proxy
| displaying redis                                 [OK]                         
| displaying postgresql                            [OK]                         
/ displaying iop-core-engine                       [OK]                         
/ displaying iop-core-gateway                      [OK]                         
/ displaying iop-core-host-inventory               [OK]                         
/ displaying iop-core-host-inventory-api           [OK]                         
/ displaying iop-core-host-inventory-migrate       [OK]                         
/ displaying iop-core-ingress                      [OK]                         
/ displaying iop-core-kafka                        [OK]                         
/ displaying iop-core-puptoo                       [OK]                         
/ displaying iop-core-yuptoo                       [OK]                         
/ displaying iop-service-advisor-backend-api       [OK]                         
/ displaying iop-service-advisor-backend-service   [OK]                         
/ displaying iop-service-remediations-api          [OK]                         
/ displaying iop-service-vmaas-reposcan            [OK]                         
/ displaying iop-service-vmaas-webapp-go           [OK]                         
/ displaying iop-service-vuln-dbupgrade            [OK]                         
/ displaying iop-service-vuln-evaluator-recalc     [OK]                         
/ displaying iop-service-vuln-evaluator-upload     [OK]                         
/ displaying iop-service-vuln-grouper              [OK]                         
/ displaying iop-service-vuln-listener             [OK]                         
/ displaying iop-service-vuln-manager              [OK]                         
/ displaying iop-service-vuln-taskomatic           [OK]                         
/ displaying pulpcore-api                          [OK]                         
/ displaying pulpcore-content                      [OK]                         
/ displaying pulpcore-worker@1.service             [OK]                         
/ displaying pulpcore-worker@2.service             [OK]                         
/ displaying pulpcore-worker@3.service             [OK]                         
/ displaying pulpcore-worker@4.service             [OK]                         
/ displaying pulpcore-worker@5.service             [OK]                         
/ displaying pulpcore-worker@6.service             [OK]                         
/ displaying tomcat                                [OK]                         
/ displaying httpd                                 [OK]                         
/ displaying foreman-proxy                         [OK]                         
/ All services are running                                            [OK]      
--------------------------------------------------------------------------------
```

Until you have the IoP services, don't continue.

In `bundle exec rails console`:
```
pry(main)> ::Ping.ping
...
{:foreman=>{:database=>{:active=>true, :duration_ms=>"1"}, :cache=>{:servers=>[{:status=>"ok", :duration_ms=>"0"}]}},
 :foreman_rh_cloud=>{:services=>{:advisor=>{:status=>"ok", :duration_ms=>"34"}, :vulnerability=>{:status=>"ok", :duration_ms=>"15"}}, :status=>"ok"},
 :katello=>
  {:services=>
    {:candlepin=>{:status=>"ok", :duration_ms=>"21"},
     :candlepin_auth=>{:status=>"ok", :duration_ms=>"21"},
     :foreman_tasks=>{:status=>"ok", :duration_ms=>"251"},
     :katello_events=>{:status=>"ok", :message=>"0 Processed, 0 Failed", :duration_ms=>"2"},
     :candlepin_events=>{:status=>"ok", :message=>"0 Processed, 0 Failed", :duration_ms=>"0"},
     :pulp3=>{:status=>"ok", :duration_ms=>"56"},
     :pulp3_content=>{:status=>"ok", :duration_ms=>"469"}},
   :status=>"ok"}}
```
You should see the `:foreman_rh_cloud` services. also try
```
[2] pry(main)> ForemanRhCloud::Ping.ping
=> {:services=>{:advisor=>{:status=>"ok", :duration_ms=>"26"}, :vulnerability=>{:status=>"ok", :duration_ms=>"15"}}, :status=>"ok"}
```

You can test the FAIL state by stopping a service, such as
```
sudo systemctl stop iop-service-advisor-backend-api
```

## Summary by Sourcery

Add a new Ping model and integrate ping/status endpoints for advisor and vulnerabilities services

New Features:
- Introduce ForemanRhCloud::Ping model to perform health checks against advisor and vulnerabilities services
- Register ping and status extensions into the ForemanRhCloud engine to expose these checks

### DIFF
--- a/app/models/foreman_rh_cloud/ping.rb
+++ b/app/models/foreman_rh_cloud/ping.rb
@@ -3,19 +3,28 @@ module ForemanRhCloud
     OK_RETURN_CODE = 'ok'.freeze
     FAIL_RETURN_CODE = 'FAIL'.freeze
 
-    SERVICE_URLS = {
-      :advisor => "https://localhost:24443/api/insights/v1/status/live/",
-      :vulnerability => "https://localhost:24443/api/vulnerability/v1/apistatus"
-    }
-
+    
     class << self
+      include ForemanRhCloud::CertAuth
+      
+      def iop_smart_proxy_url
+        @iop_smart_proxy_url ||= ForemanRhCloud.iop_smart_proxy.url
+      end
+
+      def service_urls
+        {
+          :advisor => "#{iop_smart_proxy_url}/api/insights/v1/status/live/",
+          :vulnerability => "#{iop_smart_proxy_url}/api/vulnerability/v1/apistatus"
+        }
+      end
+
       def services
-        SERVICE_URLS.keys
+        service_urls.keys
       end
 
       def status
         {
-          version: ForemanRhCloud::VERSION,
+          iop_smart_proxy_exists: ForemanRhCloud.with_iop_smart_proxy?,
           timeUTC: Time.now.getutc,
         }
       end
@@ -60,8 +69,8 @@ module ForemanRhCloud
 
         options = {}
         options[:ssl_ca_file] = ca_file unless ca_file.nil?
-        options[:ssl_client_cert] = OpenSSL::X509::Certificate.new(File.read(Setting[:ssl_certificate]))
-        options[:ssl_client_key] = OpenSSL::PKey.read(File.read(Setting[:ssl_priv_key])) 
+        options[:ssl_client_cert] = OpenSSL::X509::Certificate.new(foreman_certificate[:cert])
+        options[:ssl_client_key] = OpenSSL::PKey.read(foreman_certificate[:key]) 
 
         client = RestClient::Resource.new(url, options)
 
@@ -77,7 +86,7 @@ module ForemanRhCloud
 
       def ping_service(service_name, service_result_hash)
         exception_watch(service_result_hash) do
-          ping_url(SERVICE_URLS[service_name])
+          ping_url(service_urls[service_name])
         end
       end
     end
--- a/app/models/foreman_rh_cloud/ping.rb
+++ b/app/models/foreman_rh_cloud/ping.rb
@@ -5,7 +5,14 @@ module ForemanRhCloud
 
     class << self
       def services
-        [:insights]
+        [:advisor, :vulnerabilities]
+      end
+
+      def status
+        {
+          version: ForemanRhCloud::VERSION,
+          timeUTC: Time.now.getutc,
+        }
       end
 
       def exception_watch(result, &blk)
@@ -33,7 +40,7 @@ module ForemanRhCloud
         result = {}
         services.each do |service|
           result[service] = {}
-          ping_service(result[service])
+          ping_service(service, result[service])
         end
 
         # set overall status result code
@@ -42,9 +49,22 @@ module ForemanRhCloud
         result
       end
 
-      def ping_service(service_result)
-        exception_watch(service_result) do
-          puts "hi"
+      def ping_url(url)
+        ca_file = Setting[:ssl_ca_file]
+        request_id = ::Logging.mdc['request']
+
+        options = {}
+        options[:ssl_ca_file] = ca_file unless ca_file.nil?
+        # options[:headers] = { 'Correlation-ID' => request_id } if request_id
+        client = RestClient::Resource.new(url, options)
+
+        response = client.get
+        response.empty? ? {} : JSON.parse(response).with_indifferent_access
+      end
+
+      def ping_service(service_name, service_result_hash)
+        exception_watch(service_result_hash) do
+          ping_url("https://ip-10-0-168-225.rhos-01.prod.psi.rdu2.redhat.com/insights_cloud/api/insights/v1/rule/?impacting=true")
         end
       end
     end
--- a/app/models/foreman_rh_cloud/ping.rb
+++ b/app/models/foreman_rh_cloud/ping.rb
@@ -3,9 +3,14 @@ module ForemanRhCloud
     OK_RETURN_CODE = 'ok'.freeze
     FAIL_RETURN_CODE = 'FAIL'.freeze
 
+    SERVICE_URLS = {
+      :advisor => "http://localhost:24443/api/insights/v1/status/live/",
+      :vulnerabilities => "http://localhost:24443/api/insights/v1/status/live/"
+    }
+
     class << self
       def services
-        [:advisor, :vulnerabilities]
+        SERVICE_URLS.keys
       end
 
       def status
@@ -64,7 +69,7 @@ module ForemanRhCloud
 
       def ping_service(service_name, service_result_hash)
         exception_watch(service_result_hash) do
-          ping_url("https://ip-10-0-168-225.rhos-01.prod.psi.rdu2.redhat.com/insights_cloud/api/insights/v1/rule/?impacting=true")
+          ping_url(SERVICE_URLS[service_name])
         end
       end
     end
--- a/app/models/foreman_rh_cloud/ping.rb
+++ b/app/models/foreman_rh_cloud/ping.rb
@@ -62,17 +62,15 @@ module ForemanRhCloud
         result
       end
 
+      def logger
+        Rails.logger
+      end
+
       def ping_url(url)
-        ca_file = Setting[:ssl_ca_file]
-
-        options = {}
-        options[:ssl_ca_file] = ca_file unless ca_file.nil?
-        options[:ssl_client_cert] = OpenSSL::X509::Certificate.new(foreman_certificate[:cert])
-        options[:ssl_client_key] = OpenSSL::PKey.read(foreman_certificate[:key])
-
-        client = RestClient::Resource.new(url, options)
-
-        response = client.get
+        response = execute_cloud_request(
+          method: :get,
+          url: url
+        )
         return {} if response.empty?
         begin
           result = JSON.parse(response).with_indifferent_access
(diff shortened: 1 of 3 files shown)


app/models/foreman_rh_cloud/ping.rb
```diff
@@ -0,0 +1,85 @@
...
+        options[:ssl_client_cert] = OpenSSL::X509::Certificate.new(File.read(Setting[:ssl_certificate]))
+        options[:ssl_client_key] = OpenSSL::PKey.read(File.read(Setting[:ssl_priv_key])) 
+
+        client = RestClient::Resource.new(url, options)
+
+        response = client.get
+        return {} if response.empty?
+        begin 
+          result = JSON.parse(response).with_indifferent_access
+        rescue JSON::ParserError, NoMethodError
```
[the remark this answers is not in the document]
jeremylenz: it's because `.with_indifferent_access` expects an object and doesn't handle integers or strings

app/models/foreman_rh_cloud/ping.rb
```diff
@@ -0,0 +1,85 @@
+module ForemanRhCloud
+  class Ping
+    OK_RETURN_CODE = 'ok'.freeze
+    FAIL_RETURN_CODE = 'FAIL'.freeze
+
+    SERVICE_URLS = {
+      :advisor => "https://localhost:24443/api/insights/v1/status/live/",
+      :vulnerability => "https://localhost:24443/api/vulnerability/v1/apistatus"
```
ehelms: You should be able to derive the base URL from the smart-proxy that represents the gateway allowing the `https://localhost:24443` to be informed by the smart-proxy and not hard-coded.

app/models/foreman_rh_cloud/ping.rb
```diff
@@ -0,0 +1,85 @@
...
+    }
+
+    class << self
+      def services
+        SERVICE_URLS.keys
+      end
+
+      def status
+        {
+          version: ForemanRhCloud::VERSION,
```
ehelms: Is `status` the unauthenticated one? If so, I think we should not return the version of the plugin. We've seen requests not to display version information in unauthenticated requests.

app/models/foreman_rh_cloud/ping.rb
ehelms: Further more, there are patterns for using the SmartProxy object we an lean on. This [PR](https://github.com/theforeman/foreman_rh_cloud/pull/1049/files#diff-d703d9179cea9d0d9bb78f9d6a1028620c7d4270479ebed553223bf4826ddc3dR128) adds a method to fetch the iop smart-proxy object. 

You could then use the Smart Proxy instance for the `iop` feature to pass into a ProxyAPI class for the `iop` smart-proxy where the properties, such as endpoints and specific requests are stored in a single classs creating a nice interface. An example from the Foreman code base:

https://github.com/theforeman/foreman/blob/develop/app/services/proxy_api/puppet_ca.rb

Which can be initialized like:

```
    @puppetca = ProxyAPI::PuppetCA.new :url => puppet_ca_proxy.url
```

There is likely other re-factorings in the code that could take advantage of this pattern. I think this PR and @nofaralfasi PR can synergize nicely with our Smart Proxy patterns.

vkrizan: Could the hardcoded values be replaced by the URL from the Smart Proxy. This assumes binding on port 24443 on localhost.

app/models/foreman_rh_cloud/ping.rb
jeremylenz: This was copied from Katello ping. Should I also remove the version there?

app/models/foreman_rh_cloud/ping.rb
jeremylenz: Sounds good. Will wait for Nofar's PR to be in and then refactor.

ShimShtein: @ehelms , As long as we also have the hosted mode, I would not try to refactor the actions into Proxy API class. We use the same code to access the proxy and the hosted service, but it would be a stretch to think about the hosted services as a proxy. 
I would say it's a good refactor if we decide to stop using hosted experience.
In the meantime, I would suggest putting the URLs into the `lib/insights_cloud.rb` file similar to all the other URLs we [manage](https://github.com/theforeman/foreman_rh_cloud/blob/73c3c887db3214691f0c81d21ce9b6c5218a7568/lib/insights_cloud.rb#L8)

app/models/foreman_rh_cloud/ping.rb
```diff
@@ -0,0 +1,85 @@
...
+        result[:status] = result[:services].each_value.any? { |v| v[:status] == FAIL_RETURN_CODE } ? FAIL_RETURN_CODE : OK_RETURN_CODE
+        result
+      end
+
+      def ping_url(url)
+        ca_file = Setting[:ssl_ca_file]
+        request_id = ::Logging.mdc['request']
+
+        options = {}
+        options[:ssl_ca_file] = ca_file unless ca_file.nil?
```
ShimShtein: You can DRY it by reusing https://github.com/theforeman/foreman_rh_cloud/blob/develop/app/services/foreman_rh_cloud/cert_auth.rb concern in the file.

app/models/foreman_rh_cloud/ping.rb
ehelms: I don't think I understand this. There are fundamentally two different paths of communication: one to hosted, one to the gateway. The latter is driven by our Smart Proxy design principles and benefit from adhering to those.

app/models/foreman_rh_cloud/ping.rb
ehelms: Some additional context, @archanaserver is working on hiding the Foreman version from API responses due to a user report (https://github.com/theforeman/foreman/pull/10545). I considered this to be of a similar tack. So the answer is "probably" but not a priority compared to this work.

app/models/foreman_rh_cloud/ping.rb
ShimShtein: @ehelms that's not correct. We are using the same requests in both cases (except for this ping of course).
The only difference is in the URL - when we have the proxy - we're accessing the local gateway, and when we don't - we access the hosted environment. I would argue that in the long run, we should isolate all those requests into a single model/api class and encapsulate the communication there, but I don't see too much value in it. Especially since we have the CertAuth concern that effectively encapsulates the communication quirks

ehelms: We can follow up separately as it does need to hold up this change. If we are going to use the smart-proxy paradigm, I believe we should lean in and use it entirely for consistency. And I still think that can work for your description of the only difference being the base URL.

ShimShtein: Let's keep this discussion for later refactoring.

app/models/foreman_rh_cloud/ping.rb
jeremylenz: Just having `version` seemed lonely, but I wasn't sure what else to put in `status`. I added the IoP flag :shrug:

app/models/foreman_rh_cloud/ping.rb
jeremylenz: I included it, but seems `foreman_certificate` only takes care of the `File.read`. I still needed the other stuff.

jeremylenz:
Okay, I attempted to address comments. Ready for another round.

app/models/foreman_rh_cloud/ping.rb
ShimShtein: I won't insist, but it I think it's a bit cleaner:
``` ruby
        response = execute_cloud_request(
          method: :get,
          url: url
        )
        return {} if response.empty?
        begin
          result = JSON.parse(response).with_indifferent_access
        rescue JSON::ParserError, NoMethodError
          result = { :response => response.body&.strip }
        end
        result
```

ShimShtein (commented):
Looks OK to me. The API returns the new values as expected. 
I am good with merging this, unless we do want to add a way to view it from the UI.

jeremylenz:
@ShimShtein Updated. I like execute_cloud_request better as well, but I also had to add the def logger. not sure if there's a more correct way to do that

chris1984 (approved):
@ShimShtein if you think this is good to go, feel free to merge it, then I can get a release out Monday for the snap.

ShimShtein:
Merged, thanks!

PR #1043, opened 2025-07-24
Allow hosts controller to find hosts by Subscription UUID

jeremylenz:

Override the `find_resource` method of Foreman's `Api::V2::HostsController` and `HostsController`, so that you can access hosts by ~~Insights UUID~~ Subscription UUID.

We already have a gem in place, FriendlyId, that allows us to use host FQDNs as IDs. Sadly, though, that gem does not allow using more than one field name as an ID. So I had to develop this custom solution.

Originally I had only extended `Api::V2::HostsController`, but the Host Statuses card was not working. I realized that makes a call to `/hosts/:id/statuses` (note the lack of `api/v2`) so I had to extend the other controller as well.

Try to access the host details page via Subscription UUID in the url, such as
```
https://centos9-katello-devel.fedora.example.com/new/hosts/c39d1627-c958-433b-afc9-b3242be6717a#/Insights
```

## Summary by Sourcery

Override the find_resource method in Foreman’s API V2 and UI HostsController to enable fetching hosts by Subscription UUID

Enhancements:
- Include a concern in the engine to extend both Api::V2::HostsController and HostsController with custom behavior
- Override find_resource to first attempt host lookup via Katello::Host::SubscriptionFacet by UUID before falling back to default ID lookup

### DIFF
--- a/app/controllers/foreman_rh_cloud/concerns/api/v2/hosts_controller_extensions.rb
+++ b/app/controllers/foreman_rh_cloud/concerns/api/v2/hosts_controller_extensions.rb
@@ -1,0 +1,22 @@
+module ForemanRhCloud
+  module Concerns
+    module Api
+      module V2
+        module HostsControllerExtensions
+          extend ActiveSupport::Concern
+
+          included do
+            def find_resource
+              insights_host = ::Katello::Host::SubscriptionFacet.find_by(uuid: params[:id])&.host
+              if insights_host.present?
+                Rails.logger.debug "ForemanRhCloud host found by subscription uuid: #{params[:id]}"
+              end
+              @host = insights_host || super
+              @host
+            end
+          end
+        end
+      end
+    end
+  end
+end
--- a/lib/foreman_rh_cloud/engine.rb
+++ b/lib/foreman_rh_cloud/engine.rb
@@ -83,6 +83,12 @@ module ForemanRhCloud
         :description => N_('Configure Cloud Connector on given hosts'),
         :proxy_selector_override => ::RemoteExecutionProxySelector::INTERNAL_PROXY
       )
+
+      # API controller extensions
+      ::Api::V2::HostsController.include ::ForemanRhCloud::Concerns::Api::V2::HostsControllerExtensions
+
+      # Controller extensions
+      ::HostsController.include ::ForemanRhCloud::Concerns::Api::V2::HostsControllerExtensions
     end
 
     # Ideally this code belongs to an initializer. The problem is that Katello controllers are not initialized completely until after the end of the to_prepare blocks


app/controllers/foreman_rh_cloud/concerns/api/v2/hosts_controller_extensions.rb
```diff
@@ -0,0 +1,17 @@
+module ForemanRhCloud
+  module Concerns
+    module Api::V2::HostsControllerExtensions
+      extend ActiveSupport::Concern
+
+      included do
+        def find_resource
+          insights_host = InsightsFacet.find_by(uuid: params[:id])&.host
```
vkrizan: Can you explain the `InsightsFacet` for someone that is not that much familiar with it? Does this need to be mapped to "Insights"? I was thinking about utilizing existing data that Foreman/Katello has, which is the subscription id. Subscription id is used as an Insights host id (for cases like these). Thanks!

jeremylenz: Facets are Foreman's way of extending its `::Host::Managed` model with additional fields about the host. Katello adds the subscription facet and content facet, while foreman_rh_cloud adds the Insights facet (For some reason rh_cloud prefers to use `Insights` to refer to the facet, which I think is weird, so I aliased Insights and InsightsFacet a while back.) Since we are extending this from foreman_rh_cloud and not from Katello, it seemed correct to look at the InsightsFacet.

app/controllers/foreman_rh_cloud/concerns/api/v2/hosts_controller_extensions.rb
```diff
@@ -0,0 +1,17 @@
...
+          Rails.logger.debug "ForemanRhCloud host: #{@host}"
+          params[:id] = insights_host.name if insights_host.present?
+          @host = insights_host || super
```
vkrizan: I'm still getting:
```
Couldn't find Host::Managed with 'id'=2a7bfadc-453a-4ef0-a67e-390acafa966a [WHERE "hosts"."type" = $1 AND "hosts"."organization_id" = $2 AND "hosts"."location_id" = $3]
```
on API call to `/api/hosts/2a7bfadc-453a-4ef0-a67e-390acafa966a?show_hidden_parameters=true`

jeremylenz: I extended `/api/v2/hosts` and `/hosts` but not `/api/hosts`. Do I need to extend all three?

jeremylenz: update: it appears there's no `::Api::HostsController` to extend.

jeremylenz:
@vkrizan I didn't realize before that Insights UUID and Subscription UUID were not the same. I changed it to use Subscription UUID.

(Now I have no idea what the Insights facet UUID is for.. :shrug: )

app/controllers/foreman_rh_cloud/concerns/api/v2/hosts_controller_extensions.rb
jeremylenz: narrator voice: It was not correct.

nofaralfasi (commented):
I tested the code and confirmed that it's working as expected. The host is successfully resolved via the subscription UUID.

However, when I click on the host in the UI, it redirects to the old Host Details page instead of the new one.

jeremylenz:
> when I click on the host in the UI, it redirects to the old Host Details page instead of the new one

You mean the link from Vuln? The link would need to use `/new/hosts/` and not just `/hosts/` in their url.

nofaralfasi:
> You mean the link from Vuln? The link would need to use `/new/hosts/` and not just `/hosts/` in their url.

Yes, when I click on one of the hosts from the https://sat.example.com/foreman_rh_cloud/insights_vulnerability/CVE-ID page.

I'm redirected to the old host details page:
https://sat.example.com/hosts/SUB_UUID#/Vulnerabilities
instead of the expected URL format:
https://sat.example.com/new/hosts/SUB_UUID#/Vulnerabilities.

vkrizan:
@nofaralfasi @ShimShtein please coordinate the required URLs/paths with @lesamo. Thanks!

jeremylenz:
Opened https://github.com/theforeman/foreman_rh_cloud/pull/1048 as an alternate solution.

jeremylenz:
Closed in favor of https://github.com/theforeman/foreman_rh_cloud/pull/1048

(also, for posterity, the first try (Insights UUID) was indeed the correct UUID.)

PR #1048, opened 2025-07-29
Add endpoint to translate Insights UUIDs and redirect

jeremylenz:

Add a new endpoint, `/insights_hosts/:uuid`, where you can supply an Insights UUID and Rails will redirect you to the host details page of that host (using the database ID).

see comment

Also, feedback is accepted on (a) what to call the endpoint; and (b) what controller it goes in. These are just initial guesses.

Get a host with an Insights facet
In the browser, visit `/insights_hosts/<insights_uuid>`
Confirm that it redirects correctly

In the browser, visit `/insights_hosts/foo`
Confirm that it redirects to error page

## Summary by Sourcery

Add a new endpoint to translate an Insights UUID to a host database ID and redirect to the host details page or a not-found page.

New Features:
- Add GET /insights_hosts/:uuid endpoint for translating Insights UUIDs
- Implement translate_insights_host action in the UI requests controller to look up the InsightsFacet by UUID and redirect accordingly
- Register translate_insights_host action in the plugin permissions

### DIFF
--- a/app/controllers/insights_cloud/ui_requests_controller.rb
+++ b/app/controllers/insights_cloud/ui_requests_controller.rb
@@ -75,6 +75,18 @@ module InsightsCloud
       res.headers[new_header] = header_content
     end
 
+    def translate_insights_host
+      Rails.logger.info "translate_insights_host"
+      facet = InsightsFacet.find_by(uuid: params[:uuid])
+      if facet.present?
+        Rails.logger.info "Found InsightsFacet #{params[:uuid]}"
+        redirect_to host_details_page_path(facet&.host_id)
+      else
+        Rails.logger.error "Could not find InsightsFacet for #{params[:uuid]}"
+        redirect_to '/page-not-found'
+      end
+    end
+
     private
 
     def ensure_org
--- a/app/controllers/insights_cloud/ui_requests_controller.rb
+++ b/app/controllers/insights_cloud/ui_requests_controller.rb
@@ -76,11 +76,10 @@ module InsightsCloud
     end
 
     def translate_insights_host
-      Rails.logger.info "translate_insights_host"
       facet = InsightsFacet.find_by(uuid: params[:uuid])
       if facet.present?
-        Rails.logger.info "Found InsightsFacet #{params[:uuid]}"
-        redirect_to host_details_page_path(facet&.host_id)
+        Rails.logger.debug "Found InsightsFacet #{params[:uuid]}"
+        redirect_to host_details_page_path(facet.host_id)
       else
         Rails.logger.error "Could not find InsightsFacet for #{params[:uuid]}"
         redirect_to '/page-not-found'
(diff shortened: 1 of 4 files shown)


app/controllers/insights_cloud/ui_requests_controller.rb
```diff
@@ -75,6 +75,18 @@ def assign_header(res, cloud_res, header, transform)
       res.headers[new_header] = header_content
     end
 
+    def translate_insights_host
+      Rails.logger.info "translate_insights_host"
+      facet = InsightsFacet.find_by(uuid: params[:uuid])
+      if facet.present?
+        Rails.logger.info "Found InsightsFacet #{params[:uuid]}"
+        redirect_to host_details_page_path(facet&.host_id)
```
jeremylenz: If I were to query `facet.host.name` here instead of `facet.host_id`, I could make the url look much nicer.

jeremylenz: (This also affects the title of your browser tab.)

app/controllers/insights_cloud/ui_requests_controller.rb
```diff
@@ -75,6 +75,18 @@ def assign_header(res, cloud_res, header, transform)
       res.headers[new_header] = header_content
     end
 
+    def translate_insights_host
+      Rails.logger.info "translate_insights_host"
```
ShimShtein: I think we can remove this one

app/controllers/insights_cloud/ui_requests_controller.rb
```diff
@@ -75,6 +75,18 @@ def assign_header(res, cloud_res, header, transform)
...
+      facet = InsightsFacet.find_by(uuid: params[:uuid])
+      if facet.present?
+        Rails.logger.info "Found InsightsFacet #{params[:uuid]}"
```
ShimShtein:
```suggestion
        Rails.logger.debug "Found InsightsFacet #{params[:uuid]}"
```
I don't think we need to display this message if we don't debug the system.

app/controllers/insights_cloud/ui_requests_controller.rb
ShimShtein: Nitpick
```suggestion
        redirect_to host_details_page_path(facet.host_id)
```
We know that the facet is always here

ShimShtein (changes requested):
Test failures are related, we need to add the command to the permissions block. We can use the available `view_insights_hits` permission here: https://github.com/theforeman/foreman_rh_cloud/blob/c06996aed7d43785e169074cdaed0f2e536d5cbd/lib/foreman_rh_cloud/plugin.rb#L59

ShimShtein:
Let's just add a couple of unit tests for this controller. It should be pretty straight forward. 
Otherwise LGTM

chris1984 (commented):
LGTM, agree with Shim on tests but overall looks good. Will ack when tests are in

jeremylenz:
Tests added.

ShimShtein (approved):

ShimShtein:
Merged, thanks @jeremylenz !

PR #1049, opened 2025-07-30
Refactor local advisor engine config with SmartProxy-based detection

nofaralfasi:
**Note:** This PR builds on the changes introduced in #1062. 

Replace configuration-based IoP detection with SmartProxy feature detection. The system now automatically detects Insights on Premise mode by checking for SmartProxy instances with the 'Insights' feature instead of relying on environment variables and settings.

The main goal was to preserve backward compatibility while moving from static environment variable configuration to dynamic detection based on SmartProxy settings. Instead of relying on manually set ENV vars, the system now determines IoP mode by querying the SmartProxy, which improves accuracy and makes it easier to maintain.
There's a small trade-off in performance (a DB query instead of reading a cached variable), but it removes the need for operational ENV management. Renaming the API field from `use_local_advisor_engine` to `use_iop_mode` makes the purpose clearer.

I'm still considering whether to keep the `with_local_advisor_engine?` method and mark it as deprecated, mainly to ensure older code paths continue to work during the transition.

1. Add Smart Proxy with Insights feature:
   * URL: `https://localhost:24443`
   * Ensure **Insights** is enabled.
2. UI menu checks (based on IoP mode):
   * Insights → Vulnerability: should be **visible** if IoP mode is **enabled**.
3. Host details:
   * Go to any host page.
   * **Vulnerabilities tab** should load correctly.

## Summary by Sourcery

Introduce SmartProxy-based detection for Insights On-Prem and refactor plugin registration while renaming the legacy use_local_advisor_engine flag to use_iop_mode

New Features:
- Detect Insights On-Prem mode dynamically by querying SmartProxy features via with_insights_smartproxy?
- Register Foreman plugin using Foreman::Plugin.register DSL to define Insights settings, permissions, roles, menus, facets, and page extensions

Enhancements:
- Rename configuration, API, and UI flags from use_local_advisor_engine to use_iop_mode across the codebase
- Replace ENV-based on-prem URL and proxy logic with SmartProxy-provided URL and feature detection
- Update route definitions and service logic to conditionally enable features based on SmartProxy detection

Tests:
- Update tests to stub with_insights_smartproxy? and verify updated status messages and behavior

## Summary by Sourcery

Replace static local advisor engine configuration with dynamic SmartProxy-based detection for Insights on-premise and rename related flags to use_iop_mode.

Enhancements:
- Introduce with_insights_smartproxy? to detect the Insights feature via SmartProxy and add insights_smartproxy accessor.
- Refactor code, routes, services, views, and React components to use SmartProxy detection instead of ENV-based with_local_advisor_engine? checks.
- Rename API field, configuration flags, RABL nodes, and front-end props from use_local_advisor_engine to use_iop_mode for clarity.

Tests:
- Update tests to stub with_insights_smartproxy? and assert the new use_iop_mode flag and updated status messages.

### DIFF
--- a/app/controllers/api/v2/rh_cloud/advisor_engine_config_controller.rb
+++ b/app/controllers/api/v2/rh_cloud/advisor_engine_config_controller.rb
@@ -7,7 +7,7 @@ module Api
         api :GET, "/rh_cloud/advisor_engine_config", N_("Show if system is configured to use local iop-advisor-engine.")
         def show
           render json: {
-            use_local_advisor_engine: ForemanRhCloud.with_local_advisor_engine?,
+            use_iop_mode: ForemanRhCloud.with_iop_smart_proxy?,
           }, status: :ok
         end
       end
--- a/app/controllers/concerns/foreman_rh_cloud/iop_smart_proxy_access.rb
+++ b/app/controllers/concerns/foreman_rh_cloud/iop_smart_proxy_access.rb
@@ -1,0 +1,28 @@
+module ForemanRhCloud
+  module IopSmartProxyAccess
+    extend ActiveSupport::Concern
+
+    private
+
+    def require_non_iop_smart_proxy
+      return unless ForemanRhCloud.with_iop_smart_proxy?
+
+      handle_iop_access_error('This feature is not available when using IoP Smart Proxy')
+    end
+
+    def require_iop_smart_proxy
+      return if ForemanRhCloud.with_iop_smart_proxy?
+
+      handle_iop_access_error('This feature is not available without IoP Smart Proxy')
+    end
+
+    def handle_iop_access_error(message)
+      if api_request?
+        not_found(message)
+      else
+        error(message)
+        redirect_back(fallback_location: main_app.root_path)
+      end
+    end
+  end
+end
--- a/lib/foreman_inventory_upload/async/generate_all_reports_job.rb
+++ b/lib/foreman_inventory_upload/async/generate_all_reports_job.rb
@@ -13,7 +13,7 @@ module ForemanInventoryUpload
           return
         end
 
-        if ForemanRhCloud.with_local_advisor_engine?
+        if ForemanRhCloud.with_iop_smart_proxy?
           plan_self # so that 'run' runs
         else
           after_delay do
@@ -34,7 +34,7 @@ module ForemanInventoryUpload
       end
 
       def run
-        output[:status] = _('The scheduled process is disabled because this Foreman is configured with the use_local_advisor_engine option.') if ForemanRhCloud.with_local_advisor_engine?
+        output[:status] = _('The scheduled process is disabled because this Foreman is configured with a local IoP Smart Proxy.') if ForemanRhCloud.with_iop_smart_proxy?
       end
 
       def rescue_strategy_for_self
--- a/lib/foreman_rh_cloud.rb
+++ b/lib/foreman_rh_cloud.rb
@@ -4,13 +4,12 @@ require 'uri'
 
 module ForemanRhCloud
   def self.on_premise_url
-    return unless ForemanRhCloud.with_local_advisor_engine?
-    port = ENV['ADVISOR_ENGINE_PORT'] || "24443"
-    ENV['ADVISOR_ENGINE_URL'] || "https://localhost:#{port}"
+    return unless ForemanRhCloud.with_iop_smart_proxy?
+    ForemanRhCloud.iop_smart_proxy&.url
   end
 
   def self.env_or_on_premise_url(env_var_name)
-    ENV[env_var_name] || on_premise_url
+    on_premise_url || ENV[env_var_name]
   end
 
   def self.base_url
@@ -47,7 +46,7 @@ module ForemanRhCloud
   end
 
   def self.proxy_string
-    return '' if ForemanRhCloud.with_local_advisor_engine?
+    return '' if ForemanRhCloud.with_iop_smart_proxy?
 
     HttpProxy.default_global_content_proxy&.full_url ||
     ForemanRhCloud.global_foreman_proxy ||
@@ -106,7 +105,7 @@ module ForemanRhCloud
   end
 
   def self.legacy_insights_ca
-    "#{ForemanRhCloud::Engine.root}/config/rh_cert-api_chain.pem" unless ForemanRhCloud.with_local_advisor_engine?
+    "#{ForemanRhCloud::Engine.root}/config/rh_cert-api_chain.pem" unless ForemanRhCloud.with_iop_smart_proxy?
   end
 
   def self.cloud_url_validator
--- a/lib/foreman_rh_cloud/engine.rb
+++ b/lib/foreman_rh_cloud/engine.rb
@@ -121,8 +121,12 @@ module ForemanRhCloud
     end
   end
 
-  def self.with_local_advisor_engine?
-    SETTINGS.dig(:foreman_rh_cloud, :use_local_advisor_engine) || false
+  def self.with_iop_smart_proxy?
+    SmartProxy.with_features('iop').exists?
+  end
+
+  def self.iop_smart_proxy
+    SmartProxy.with_features('iop').first
   end
 
   def self.ca_cert
--- a/lib/insights_cloud/async/insights_scheduled_sync.rb
+++ b/lib/insights_cloud/async/insights_scheduled_sync.rb
@@ -13,7 +13,7 @@ module InsightsCloud
           return
         end
 
-        if ForemanRhCloud.with_local_advisor_engine?
+        if ForemanRhCloud.with_iop_smart_proxy?
           plan_self
         else
           after_delay do
@@ -23,7 +23,7 @@ module InsightsCloud
       end
 
       def run
-        output[:status] = _('The scheduled process is disabled because this Foreman is configured with the use_local_advisor_engine option.') if ForemanRhCloud.with_local_advisor_engine?
+        output[:status] = _('The scheduled process is disabled because this Foreman is configured with a local IoP Smart Proxy.') if ForemanRhCloud.with_iop_smart_proxy?
       end
 
       def plan_full_sync
--- a/lib/inventory_sync/async/inventory_scheduled_sync.rb
+++ b/lib/inventory_sync/async/inventory_scheduled_sync.rb
@@ -13,7 +13,7 @@ module InventorySync
           return
         end
 
-        if ForemanRhCloud.with_local_advisor_engine?
+        if ForemanRhCloud.with_iop_smart_proxy?
           plan_self # so that 'run' runs
         else
           after_delay do
@@ -35,7 +35,7 @@ module InventorySync
       end
 
       def run
-        output[:status] = _('The scheduled process is disabled because this Foreman is configured with the use_local_advisor_engine option.') if ForemanRhCloud.with_local_advisor_engine?
+        output[:status] = _('The scheduled process is disabled because this Foreman is configured with a local IoP Smart Proxy.') if ForemanRhCloud.with_iop_smart_proxy?
       end
 
       def plan_remove_insights_hosts(org_id)
--- a/webpack/common/Hooks/ConfigHooks.js
+++ b/webpack/common/Hooks/ConfigHooks.js
@@ -14,6 +14,5 @@ export const useAdvisorEngineConfig = () => {
   );
 
   // eslint-disable-next-line camelcase
-  const isLocalAdvisorEngine = advisorEngineConfig?.use_local_advisor_engine;
-  return isLocalAdvisorEngine;
+  return advisorEngineConfig?.use_iop_mode;
 };
--- a/app/controllers/api/v2/rh_cloud/cloud_request_controller.rb
+++ b/app/controllers/api/v2/rh_cloud/cloud_request_controller.rb
@@ -1,6 +1,9 @@
 module Api::V2::RhCloud
   class CloudRequestController < ::Api::V2::BaseController
+    include ForemanRhCloud::IopSmartProxyAccess
     layout false
+
+    before_action :require_non_iop_smart_proxy, only: [:update]
 
     KNOWN_DIRECTIVES = {
       'playbook-sat' => :handle_run_playbook_request,
--- a/app/controllers/api/v2/rh_cloud/inventory_controller.rb
+++ b/app/controllers/api/v2/rh_cloud/inventory_controller.rb
@@ -5,6 +5,9 @@ module Api
         include ::Api::Version2
         include InventoryUpload::ReportActions
         include InventoryUpload::TaskActions
+        include ForemanRhCloud::IopSmartProxyAccess
+
+        before_action :require_non_iop_smart_proxy, only: [:enable_cloud_connector]
 
         api :GET, "/organizations/:organization_id/rh_cloud/report", N_("Download latest report")
         param :organization_id, Integer, required: true, desc: N_("Set the current organization context for the request")
--- a/app/controllers/concerns/insights_cloud/package_profile_upload_extensions.rb
+++ b/app/controllers/concerns/insights_cloud/package_profile_upload_extensions.rb
@@ -10,7 +10,7 @@ module InsightsCloud
     end
 
     def generate_host_report
-      return unless ForemanRhCloud.with_local_advisor_engine?
+      return unless ForemanRhCloud.with_iop_smart_proxy?
 
       logger.debug("Generating host-specific report for host #{@host.name}")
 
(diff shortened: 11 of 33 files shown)


lib/foreman_rh_cloud/engine.rb
```diff
@@ -121,8 +121,12 @@ def self.register_scheduled_task(task_class, cronline)
...
   end
 
-  def self.with_local_advisor_engine?
-    SETTINGS.dig(:foreman_rh_cloud, :use_local_advisor_engine) || false
+  def self.with_insights_smartproxy?
+    SmartProxy.with_features('Insights').exists?
+  end
+
+  def self.insights_smartproxy
+    SmartProxy.with_features('Insights').first
```
nofaralfasi: @ShimShtein is it possible we have more than one smart_proxy with the Insights feature enabled?

lib/foreman_rh_cloud.rb
```diff
@@ -4,13 +4,12 @@
 
 module ForemanRhCloud
   def self.on_premise_url
-    return unless ForemanRhCloud.with_local_advisor_engine?
-    port = ENV['ADVISOR_ENGINE_PORT'] || "24443"
-    ENV['ADVISOR_ENGINE_URL'] || "https://localhost:#{port}"
```
nofaralfasi, line 9: @ShimShtein I want to confirm whether it's safe to remove these lines.
I see that this method is still being used [here](https://github.com/theforeman/foreman_rh_cloud/blob/develop/app/services/foreman_rh_cloud/rules_ingester.rb#L15).
Should we keep it as a fallback for default values?
If we do need to keep it, the implementation will have to be updated due to recent changes in the `env_or_on_premise_url` method.

lib/foreman_rh_cloud/engine.rb
ShimShtein: Not really, at least not at this stage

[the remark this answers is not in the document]
nofaralfasi: See https://github.com/theforeman/foreman_rh_cloud/pull/1049#discussion_r2245277903

lib/foreman_rh_cloud.rb
ShimShtein: Seems OK. In on-prem we want the smart-proxy to force-drive the URLs

lib/insights_cloud/async/insights_scheduled_sync.rb
```diff
@@ -23,7 +23,7 @@ def plan
       end
 
       def run
-        output[:status] = _('The scheduled process is disabled because this Foreman is configured with the use_local_advisor_engine option.') if ForemanRhCloud.with_local_advisor_engine?
+        output[:status] = _('The scheduled process is disabled because this Foreman is configured with a local Insights SmartProxy.') if ForemanRhCloud.with_insights_smartproxy?
```
ShimShtein:
```suggestion
        output[:status] = _('The scheduled process is disabled because this Foreman is configured with a local Insights Smart Proxy.') if ForemanRhCloud.with_insights_smartproxy?
```
I think it should be two words

lib/foreman_inventory_upload/async/generate_all_reports_job.rb
```diff
@@ -34,7 +34,7 @@ def plan
       end
 
       def run
-        output[:status] = _('The scheduled process is disabled because this Foreman is configured with the use_local_advisor_engine option.') if ForemanRhCloud.with_local_advisor_engine?
+        output[:status] = _('The scheduled process is disabled because this Foreman is configured with a local Insights SmartProxy.') if ForemanRhCloud.with_insights_smartproxy?
```
ShimShtein:
```suggestion
        output[:status] = _('The scheduled process is disabled because this Foreman is configured with a local Insights Smart Proxy.') if ForemanRhCloud.with_insights_smartproxy?
```

lib/inventory_sync/async/inventory_scheduled_sync.rb
```diff
@@ -35,7 +35,7 @@ def plan_org_sync(org)
       end
 
       def run
-        output[:status] = _('The scheduled process is disabled because this Foreman is configured with the use_local_advisor_engine option.') if ForemanRhCloud.with_local_advisor_engine?
+        output[:status] = _('The scheduled process is disabled because this Foreman is configured with a local Insights SmartProxy.') if ForemanRhCloud.with_insights_smartproxy?
```
ShimShtein:
```suggestion
        output[:status] = _('The scheduled process is disabled because this Foreman is configured with a local Insights Smart Proxy.') if ForemanRhCloud.with_insights_smartproxy?
```

webpack/common/Hooks/ConfigHooks.js
```diff
@@ -14,6 +14,6 @@ export const useAdvisorEngineConfig = () => {
   );
 
   // eslint-disable-next-line camelcase
-  const isLocalAdvisorEngine = advisorEngineConfig?.use_local_advisor_engine;
+  const isLocalAdvisorEngine = advisorEngineConfig?.use_iop_mode;
   return isLocalAdvisorEngine;
```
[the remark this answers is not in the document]
ShimShtein: @nofaralfasi The AI is right here, I'll leave it up to you if you want to change it.

app/controllers/api/v2/rh_cloud/advisor_engine_config_controller.rb
```diff
@@ -7,7 +7,7 @@ class AdvisorEngineConfigController < ::Api::V2::BaseController
         api :GET, "/rh_cloud/advisor_engine_config", N_("Show if system is configured to use local iop-advisor-engine.")
         def show
           render json: {
-            use_local_advisor_engine: ForemanRhCloud.with_local_advisor_engine?,
+            use_iop_mode: ForemanRhCloud.with_insights_smartproxy?,
```
ehelms: Can we sub `iop` for `insights` ? 

@ShimShtein @vkrizan I am thinking we should change the smart-proxy feature as well. We have an inconsistency in nomenclature.

ShimShtein: The proxy feature is `Insights`. We can converge to `local_insights_mode` WDYT?

ehelms: I think my point is being missed. I am wanting to remove the use of the word Insights in favor of just using `iop`

vkrizan: I don't mind either. I'm with Shim, that the feature is name Insights, like the opensource counterpart. The "on Prem" can be implied for Foreman/Satellite.

ehelms: Proposed change -- https://github.com/RedHatInsights/iop-gateway/pull/17

lib/foreman_rh_cloud/engine.rb
```diff
@@ -121,8 +121,12 @@ def self.register_scheduled_task(task_class, cronline)
     end
   end
 
-  def self.with_local_advisor_engine?
-    SETTINGS.dig(:foreman_rh_cloud, :use_local_advisor_engine) || false
+  def self.with_insights_smartproxy?
```
jeremylenz: Smart Proxy is two words.

```suggestion
  def self.with_insights_smart_proxy?
```

lib/foreman_rh_cloud/engine.rb
```diff
@@ -121,8 +121,12 @@ def self.register_scheduled_task(task_class, cronline)
...
+    SmartProxy.with_features('Insights').exists?
+  end
+
+  def self.insights_smartproxy
```
jeremylenz:
```suggestion
  def self.insights_smart_proxy
```

jeremylenz (commented):
Non-blocking nit below :)

lib/foreman_rh_cloud/engine.rb
```diff
@@ -121,8 +121,12 @@ def self.register_scheduled_task(task_class, cronline)
...
   end
 
-  def self.with_local_advisor_engine?
-    SETTINGS.dig(:foreman_rh_cloud, :use_local_advisor_engine) || false
+  def self.with_iop_smart_proxy?
+    SmartProxy.with_features('Insights').exists?
+  end
+
+  def self.iop_smart_proxy
+    SmartProxy.with_features('Insights').first
```
ehelms: This has to be updated now to use `iop` instead of `Insights`

app/controllers/concerns/foreman_rh_cloud/iop_smart_proxy_access.rb
```diff
@@ -0,0 +1,19 @@
+module ForemanRhCloud
+  module IopSmartProxyAccess
+    extend ActiveSupport::Concern
+
+    private
+
+    def require_no_iop_smart_proxy
+      return unless ForemanRhCloud.with_iop_smart_proxy?
+
+      render json: { error: 'Not available when using IoP Smart Proxy' }, status: :not_found
```
ShimShtein: Can we use the `not_found` method we have defined in our controllers: https://github.com/theforeman/foreman/blob/5458a98b7acbbc56f3602a41c4ba39b2a42c58f6/app/controllers/application_controller.rb#L105 and https://github.com/theforeman/foreman/blob/5458a98b7acbbc56f3602a41c4ba39b2a42c58f6/app/controllers/api/base_controller.rb#L132

The only thing is that we are loosing the exception, and it will only be logged, and not shown to the user.
@jeremylenz WDYT?

Another option is to render custom response similar to https://github.com/theforeman/foreman/blob/5458a98b7acbbc56f3602a41c4ba39b2a42c58f6/app/controllers/application_controller.rb#L105

jeremylenz: Yeah, I don't like `render json` anyway. i know rh_cloud uses it everywhere :vomiting_face:  I think not_found would be better

lib/foreman_rh_cloud/engine.rb
```diff
@@ -121,8 +121,12 @@ def self.register_scheduled_task(task_class, cronline)
     end
   end
 
-  def self.with_local_advisor_engine?
-    SETTINGS.dig(:foreman_rh_cloud, :use_local_advisor_engine) || false
+  def self.with_iop_smart_proxy?
+    SmartProxy.with_features('iop').exists?
```
parthaa, line 125: Is this an expensive operation and if so can we cache it ?

parthaa: optional :)

parthaa: Ok this is a where clause. No need cache I think we are good

app/controllers/concerns/foreman_rh_cloud/iop_smart_proxy_access.rb
```diff
@@ -0,0 +1,19 @@
+module ForemanRhCloud
+  module IopSmartProxyAccess
+    extend ActiveSupport::Concern
+
+    private
+
+    def require_no_iop_smart_proxy
```
jeremylenz: I think "non_iop" is more clear
```suggestion
    def require_non_iop_smart_proxy
```

parthaa: How about _iopless_ :)

jeremylenz:
Now that https://github.com/theforeman/foreman_rh_cloud/pull/1059 is merged, https://github.com/theforeman/foreman_rh_cloud/blob/bebb3982548cba35606df914533ec2c698469936/app/services/foreman_rh_cloud/hit_remediations_retriever.rb#L5 will also need attention :)

parthaa:
Proposed patch. Sounds like the `with_local_advisor_engine` was still used in a few places

```patch
diff --git a/app/controllers/insights_cloud/api/machine_telemetries_controller.rb b/app/controllers/insights_cloud/api/machine_telemetries_controller.rb
index 1678a3b..be13f02 100644
--- a/app/controllers/insights_cloud/api/machine_telemetries_controller.rb
+++ b/app/controllers/insights_cloud/api/machine_telemetries_controller.rb
@@ -12,7 +12,7 @@ module InsightsCloud::Api
     skip_after_action :log_response_body, :only => [:forward_request]
     skip_before_action :check_media_type, :only => [:forward_request]
     after_action :update_host_insights_status, only: [:forward_request]
-    after_action :update_host_facet, only: [:forward_request], if: -> { ForemanRhCloud.with_local_advisor_engine? }
+    after_action :update_host_facet, only: [:forward_request], if: -> { ForemanRhCloud.with_iop_smart_proxy? }
 
     # The method that "proxies" requests over to Cloud
     def forward_request
diff --git a/lib/foreman_inventory_upload/generators/slice.rb b/lib/foreman_inventory_upload/generators/slice.rb
index c3c5106..ab2f026 100644
--- a/lib/foreman_inventory_upload/generators/slice.rb
+++ b/lib/foreman_inventory_upload/generators/slice.rb
@@ -253,7 +253,7 @@ module ForemanInventoryUpload
 
       def report_yum_repos(host)
         return unless host&.content_facet&.bound_repositories&.any?
-        return unless ForemanRhCloud.with_local_advisor_engine?
+        return unless ForemanRhCloud.with_iop_smart_proxy?
 
         @stream.array_field('yum_repos') do
           host.content_facet.bound_repositories.each_with_index do |repo, index|
index 8cbb957..b0d1dd8 100644
@@ -46,7 +46,7 @@ namespace :rh_cloud_inventory do
           archived_report_generator.render(organization: organization, filter: filter)
           puts "Successfully generated #{target} for organization id #{organization}"
 
-          next unless ForemanRhCloud.with_local_advisor_engine?
+          next unless ForemanRhCloud.with_iop_smart_proxy?
 
           puts 'Creating missing insights facets'
           hosts_without_facets = ForemanInventoryUpload::Generators::Queries.for_org(organization, hosts_query: 'null? insights_uuid')
diff --git a/test/controllers/insights_cloud/api/machine_telemetries_controller_test.rb b/test/controllers/insights_cloud/api/machine_telemetries_controller_test.rb
index 70228c4..8c6646b 100644
--- a/test/controllers/insights_cloud/api/machine_telemetries_controller_test.rb
+++ b/test/controllers/insights_cloud/api/machine_telemetries_controller_test.rb
@@ -154,7 +154,7 @@ module InsightsCloud::Api
 
       test "should create insights facet" do
         assert_nil InsightsFacet.find_by(host_id: @host.id)
-        ForemanRhCloud.stubs(:with_local_advisor_engine?).returns(true)
+        ForemanRhCloud.stubs(:with_iop_smart_proxy?).returns(true)
         req = RestClient::Request.new(:method => 'GET', :url => 'http://test.theforeman.org/')
         net_http_resp = Net::HTTPResponse.new(1.0, 200, "OK")
         net_http_resp[:content_type] = 'application/zip'
diff --git a/test/unit/slice_generator_test.rb b/test/unit/slice_generator_test.rb
index e630ac0..6d6e3e9 100644
--- a/test/unit/slice_generator_test.rb
+++ b/test/unit/slice_generator_test.rb
@@ -999,7 +999,7 @@ class SliceGeneratorTest < ActiveSupport::TestCase
   end
 
   test 'reports yum repos' do
-    ForemanRhCloud.stubs(:with_local_advisor_engine?).returns(true)
+    ForemanRhCloud.stubs(:with_iop_smart_proxy?).returns(true)
     FactoryBot.create(:katello_content, cp_content_id: '1', organization: @host.organization, name: 'Test Content', label: 'test-content')
     repo = FactoryBot.build(
       :katello_repository,
```

parthaa:
~Lets close this and use https://github.com/theforeman/foreman_rh_cloud/pull/1062 
That pr includes the code in this pr + fixes + rebased conflicts etc.~ closing 1062

nofaralfasi:
After discussing with @ShimShtein, we decided to drop the last commit from @parthaa's PR https://github.com/theforeman/foreman_rh_cloud/pull/1062. It required restarting Foreman when adding or removing a Smart Proxy with or without the IoP feature, which isn't ideal.
We think it's better not to cache this data to avoid that.

app/controllers/concerns/foreman_rh_cloud/iop_smart_proxy_access.rb
ShimShtein: I was thinking in `html` case, it would be nice to get the user to the previous page and set a specific message like `the page you were trying to load is not available in IoP`. How hard would it be to implement this?

nofaralfasi: I updated this commit https://github.com/theforeman/foreman_rh_cloud/pull/1049/commits/f01ccc7af722359672f3f7c9630839b8ac5b465f, but it's not fully tested yet.

ShimShtein (approved):

PR #1067, opened 2025-08-07
Fixed whitelisted paths

ShimShtein:
- Now the paths are properly recognized.
- Fixed top correct parameter names (tags instead of tag)
- Fixed an edge case where no hosts were present

Make sure that no hosts are present for empty org
Make sure hosts are displayed if the org is not empty

## Summary by Sourcery

Enhance API request forwarding by mapping whitelisted paths to dynamic tag parameters, update the default tag key to :tags, and prevent unnecessary tag update calls when no hosts exist.

Bug Fixes:
- Fix forwarded parameter key from :tag to :tags
- Ensure whitelisted paths are correctly recognized

Enhancements:
- Refactor SCOPED_REQUESTS to associate each path pattern with a custom tag_name
- Update prepare_tags and scope_request? to use the dynamic tag_name for scoped GET requests
- Skip executing the cloud tag update request when the host_id_list is empty

Tests:
- Adjust existing tests to expect :tags instead of :tag
- Add unit tests for scope_request? behavior (GET vs non-GET and unmatched paths)
- Add tests for prepare_tags with a custom tag_name
- Add tests to verify update_tag skips the cloud request when no hosts are present

### DIFF
--- a/app/services/foreman_rh_cloud/insights_api_forwarder.rb
+++ b/app/services/foreman_rh_cloud/insights_api_forwarder.rb
@@ -5,13 +5,13 @@ module ForemanRhCloud
     include ForemanRhCloud::GatewayRequest
 
     SCOPED_REQUESTS = [
-      %r{/api/vulnerability/v1/vulnerabilities/cves},
-      %r{/api/vulnerability/v1/dashbar},
-      %r{/api/vulnerability/v1/cves/[^/]+/affected_systems},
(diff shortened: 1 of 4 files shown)


jeremylenz (approved):
Thanks @ShimShtein 
image

Works as expected. I see the above in an empty org and I see recommendations in my default org.

PR #1068, opened 2025-08-07
Update to new name of API node for useIopMode

jeremylenz:

Since https://github.com/theforeman/foreman_rh_cloud/pull/1049, the All Hosts page Recommendations column displays '-' even if a host has recommendations.

It's because the name of the API response node changed from `use_local_advisor_engine` to `use_iop_mode` and we didn't catch it via grep because I used `propsToCamelCase`.

1. Set up IoP with hosts with recommendations
2. Go to Hosts - All Hosts and add the Recommendations column
3. verify that the numbers of recommendations are correct
4. verify that 'HostedRecommendationsCell' does not appear in the React tree when in IoP mode

## Summary by Sourcery

Enhancements:
- Rename insights attribute useLocalAdvisorEngine to useIopMode for selecting the appropriate recommendations cell

### DIFF
--- a/webpack/ForemanColumnExtensions/index.js
+++ b/webpack/ForemanColumnExtensions/index.js
@@ -46,7 +46,7 @@ const RecommendationsCell = hostDetails => {
     hostDetails?.insights_attributes ?? {}
   );
 
-  return insightsAttributes.useLocalAdvisorEngine ? (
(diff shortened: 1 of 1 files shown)


chris1984 (approved):
LGTM

PR #1073, opened 2025-08-18
Do not modify resolution_id unless isIoP

jeremylenz:

In non-IoP remediations, a regression was introduced recently that causes

```
'Error during rendering of a job template' error (NoMethodError): undefined method `[]' for nil:NilClass
```

This is because our new `getResolutionId` function has changed behavior and causes the hash lookup here to fail --> https://github.com/theforeman/foreman_rh_cloud/blob/5d433f499ab907ad9aa3fd945f08f92841baff44/app/services/foreman_rh_cloud/hit_remediations_retriever.rb#L58

Changing it back to the previous behavior in non-IoP configurations seems to resolve it.

On develop branch:
Get a host with a remediation that has 2 or more remediation options
Click Remediate to bring up the modal, but do not click any of the radio buttons -- immediately click the Remediate button to submit
In the REX job inputs, notice the `resolution_id` has two joined numbers, such as "27_28". The template rendering will fail with the NilClass error.

In contrast:
Click Remediate to bring up the modal
Click the second radio button, and then click the first again
Submit
You'd think this should produce the same result, but it doesn't. The input shows an unjoined number (such as "27") and the template rendering succeeds.

After checking out this PR, running the former (first) flow above should also succeed.

## Summary by Sourcery

Bug Fixes:
- Prevent resolution_id from being prefixed with the rule ID when isIop is false by introducing an isIop parameter to getResolutionId and passing it through modifyRows

## Summary by Sourcery

Prevent resolution_id from being prefixed with the rule ID for non-IoP remediations by introducing an isIop flag in getResolutionId and passing it through modifyRows.

Bug Fixes:
- Restore previous resolution_id format for non-IoP remediation flows to prevent nil lookup errors during template rendering

Enhancements:
- Add an isIop flag to getResolutionId and propagate it through modifyRows to control resolution_id formatting

### DIFF
--- a/webpack/InsightsCloudSync/Components/RemediationModal/RemediationHelpers.js
+++ b/webpack/InsightsCloudSync/Components/RemediationModal/RemediationHelpers.js
@@ -3,8 +3,10 @@ import React from 'react';
 import { orderBy } from 'lodash';
 import Resolutions from './Resolutions';
 
-export const getResolutionId = (selectedResolution, id) =>
-  `${id}_${selectedResolution}`;
+export const getResolutionId = (selectedResolution, id, isIop = true) => {
+  if (isIop) return `${id}_${selectedResolution}`;
+  return selectedResolution;
+};
 
 export const modifyRows = (
   remediations,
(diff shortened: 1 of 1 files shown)


ShimShtein (approved):

chris1984 (approved):
Works great, first and second workflow work fine with pr checked out and not seeing the `'Error during rendering of a job template' error (NoMethodError): undefined method []' for nil:NilClass`

Screenshot 2025-08-21 at 11-33-54

chris1984:
Updated screenshot, showing it's working(I linked the before by accident)
Screenshot 2025-08-25 at 10-52-24

PR #1077, opened 2025-08-20
Fix the upload URL to follow the iop scheme if present

ShimShtein:
Make sure the upload URL follows the IoP endpoint, if present

## Summary by Sourcery

Enhancements:
- Update upload_url to use ForemanRhCloud.cert_base_url for the default endpoint instead of a hardcoded URL and remove memoization

### DIFF
--- a/lib/foreman_inventory_upload.rb
+++ b/lib/foreman_inventory_upload.rb
@@ -58,7 +58,7 @@ module ForemanInventoryUpload
 
   def self.upload_url
     # for testing set ENV to 'https://ci.cloud.redhat.com/api/ingress/v1/upload'
-    @upload_url ||= ENV['SATELLITE_INVENTORY_UPLOAD_URL'] || 'https://cert.cloud.redhat.com/api/ingress/v1/upload'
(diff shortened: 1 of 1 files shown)


chris1984 (approved):

Gauravtalreja1 (approved):
ACK, tested for IoP/non-IoP setups and it works as expected 🍏

chris1984:
@ShimShtein do you know why the tests are failing? I kicked it off a few times but they fail with the same error

ShimShtein:
It's loading order problem: remote execution plugin tries to enhance the InterfacesController from core, but the core class is not loaded at this point: https://github.com/theforeman/foreman_remote_execution/blob/ffc0839dbdd9c8eb784c577f9ea3e41548c7bffa/lib/foreman_remote_execution/engine.rb#L81

https://github.com/theforeman/foreman_rh_cloud/actions/runs/17100892545/job/48499941936?pr=1077#step:21:62

jeremylenz:
Test failures unrelated; merging

PR #1078, opened 2025-08-20
Add VMaaS reposcan sync task for repository sync

nofaralfasi:
Triggers VMaaS vulnerability scanning updates when Katello repositories are synchronized.

Added VMaaS reposcan sync functionality that automatically triggers vulnerability scanning updates when Katello repositories are synchronized.

The sync is needed only when the content actually hosted by Satellite changes. Therefore, the existing trigger in the PR for Repository Sync should be sufficient, and you don't need a second trigger for Repository Sets.

1. Enable IoP Smart Proxy
2. Sync a Katello repository (Content → Products → (Choose Product) → Sync Now)
3. Check logs for VMaaS sync trigger

## Summary by Sourcery

Introduce an asynchronous VMaaS reposcan sync task that triggers vulnerability scanning updates via the IoP Smart Proxy when a Katello repository is synchronized

New Features:
- Add Async::VmaasReposcanSyncTask to automatically trigger VMaaS vulnerability scanning updates on Katello repository sync

Enhancements:
- Add vmaas_reposcan_sync_url helper to construct the VMaaS reposcan sync API endpoint

### DIFF
--- a/lib/insights_cloud/async/vmaas_reposcan_sync.rb
+++ b/lib/insights_cloud/async/vmaas_reposcan_sync.rb
@@ -1,0 +1,71 @@
+require 'rest-client'
+
+module InsightsCloud
+  module Async
+    # Triggers VMaaS reposcan sync via IoP gateway when repositories are synced
+    class VmaasReposcanSync < ::Actions::EntryAction
+      include ::ForemanRhCloud::CertAuth
+
+      # Subscribe to Katello repository sync hook action, if available
+      def self.subscribe
+        'Actions::Katello::Repository::SyncHook'.constantize
+      rescue NameError
+        Rails.logger.debug('VMaaS reposcan sync: Repository::SyncHook action not found')
+        nil
+      end
+
+      def plan(repo, *_args)
+        return unless ::ForemanRhCloud.with_iop_smart_proxy?
+
+        repo_id = repo.is_a?(Hash) ? (repo[:id] || repo['id']) : nil
+        unless repo_id
+          logger.error("VMaaS reposcan sync: missing repository id in SyncHook plan parameters: #{repo.inspect}")
+          return
+        end
+
+        plan_self
+      end
+
+      def run
+        url = ::InsightsCloud.vmaas_reposcan_sync_url
+
+        response = execute_cloud_request(
+          method: :put,
+          url: url,
+          headers: { 'Content-Type' => 'application/json' }
+        )
+
+        if response.code >= 200 && response.code < 300
+          message = "VMaaS reposcan sync triggered successfully: #{response.code}"
+          logger.info(message)
+        else
+          message = "VMaaS reposcan sync failed with status: #{response.code}, body: #{response.body}"
+          logger.error(message)
+        end
+        output[:message] = message
+
+        response
+      rescue RestClient::ExceptionWithResponse => e
+        message = "VMaaS reposcan sync failed: #{e.response&.code} - #{e.response&.body}"
+        logger.error(message)
+        output[:message] = message
+        raise
+      rescue StandardError => e
+        message = "Error triggering VMaaS reposcan sync: #{e.message}, response: #{e.respond_to?(:response) ? e.response : nil}"
+        logger.error(message)
+        output[:message] = message
+        raise
+      end
+
+      def rescue_strategy_for_self
+        Dynflow::Action::Rescue::Skip
+      end
+
+      private
+
+      def logger
+        action_logger
+      end
+    end
+  end
+end
--- a/lib/insights_cloud.rb
+++ b/lib/insights_cloud.rb
@@ -44,4 +44,8 @@ module InsightsCloud
   def self.enable_cloud_remediations_param
     'enable_cloud_remediations'
   end
+
+  def self.vmaas_reposcan_sync_url
+    ForemanRhCloud.iop_smart_proxy.url + '/api/vmaas-reposcan/sync'
+  end
 end
--- a/test/unit/lib/insights_cloud/async/vmaas_reposcan_sync_test.rb
+++ b/test/unit/lib/insights_cloud/async/vmaas_reposcan_sync_test.rb
@@ -1,0 +1,137 @@
+require 'test_plugin_helper'
+require 'foreman_tasks/test_helpers'
+
+class VmaasReposcanSyncTest < ActiveSupport::TestCase
+  include ForemanTasks::TestHelpers::WithInThreadExecutor
+
+  setup do
+    @repo_payload = { id: 123 }
+    @expected_url = 'https://example.com/api/v1/vmaas/reposcan/sync'
+    InsightsCloud.stubs(:vmaas_reposcan_sync_url).returns(@expected_url)
+    ForemanRhCloud.stubs(:with_iop_smart_proxy?).returns(true)
+  end
+
+  # Planning behavior
+  test 'plan plans_self when repo payload has id and IoP is available' do
+    InsightsCloud::Async::VmaasReposcanSync.any_instance.expects(:plan_self).once
+
+    ForemanTasks.sync_task(InsightsCloud::Async::VmaasReposcanSync, @repo_payload)
+  end
+
+  test 'plan does not plan_self when repo payload is missing id' do
+    payload_without_id = {}
+
+    mock_logger = mock('logger')
+    mock_logger.expects(:error).with { |msg| msg =~ /missing repository id/i }
+    InsightsCloud::Async::VmaasReposcanSync.any_instance.stubs(:logger).returns(mock_logger)
+    InsightsCloud::Async::VmaasReposcanSync.any_instance.expects(:plan_self).never
+
+    ForemanTasks.sync_task(InsightsCloud::Async::VmaasReposcanSync, payload_without_id)
+  end
+
+  test 'plan does not plan_self when IoP smart proxy is not available' do
+    ForemanRhCloud.stubs(:with_iop_smart_proxy?).returns(false)
+
+    InsightsCloud::Async::VmaasReposcanSync.any_instance.expects(:plan_self).never
+
+    ForemanTasks.sync_task(InsightsCloud::Async::VmaasReposcanSync, @repo_payload)
+  end
+
+  test 'plan skips repo_id validation when IoP smart proxy is not available' do
+    ForemanRhCloud.stubs(:with_iop_smart_proxy?).returns(false)
+    payload_without_id = {}
+
+    # Logger should not be called since IoP check returns early
+    InsightsCloud::Async::VmaasReposcanSync.any_instance.expects(:logger).never
+    InsightsCloud::Async::VmaasReposcanSync.any_instance.expects(:plan_self).never
+
+    ForemanTasks.sync_task(InsightsCloud::Async::VmaasReposcanSync, payload_without_id)
+  end
+
+  test 'plan does not plan_self when repository is nil' do
+    mock_logger = mock('logger')
+    mock_logger.expects(:error).with { |msg| msg =~ /missing repository id/i }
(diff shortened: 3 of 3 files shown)


lib/insights_cloud/async/vmaas_reposcan_sync_task.rb
```diff
@@ -0,0 +1,23 @@
+require_relative 'vmaas_reposcan_sync_task_base'
+
+module InsightsCloud
+  module Async
+    # Triggers VMaaS reposcan sync when repositories are synced
+    class VmaasReposcanSyncTask < VmaasReposcanSyncTaskBase
+      def plan(repo, *_args)
```
nofaralfasi: Keep *args to stay compatible with Katello::Repository::Sync#plan
see https://github.com/Katello/katello/blob/master/app/lib/actions/katello/repository/sync.rb#L23

lib/insights_cloud/async/vmaas_reposcan_sync_task_base.rb
```diff
@@ -0,0 +1,45 @@
...
+        trigger_vmaas_reposcan_sync
+      end
+
+      def rescue_strategy_for_self
+        Dynflow::Action::Rescue::Skip
+      end
+
+      private
+
+      def trigger_vmaas_reposcan_sync
```
nofaralfasi: Should we move this to the run method?

jeremylenz: I don't see why not. It would also make it easier to test.

lib/insights_cloud/async/vmaas_reposcan_sync_task.rb
```diff
@@ -0,0 +1,23 @@
+require_relative 'vmaas_reposcan_sync_task_base'
```
jeremylenz: In Katello, Dynflow actions are namespaced under `Actions::Katello::<model name>::<action name>`. So for instance, Actions::Katello::Organization::ManifestRefresh is in `app/lib/actions/katello/organization/manifest_refresh.rb`.

I see you've modeled this after two existing Dynflow actions in rh_cloud, but nothing about `InsightsCloud::Async` says that they're Dynflow actions. Also, I'm not even sure this action would fall under Async (see the other comment I am about to write.)

I feel like this would make more sense namespaced under `Actions::ForemanRhCloud::VmaasReposcanSync`. (Not `VmaasReposcanSyncTask`, because that's redundant) Thoughts?

lib/insights_cloud/async/vmaas_reposcan_sync_task.rb
```diff
@@ -0,0 +1,23 @@
...
+    # Triggers VMaaS reposcan sync when repositories are synced
+    class VmaasReposcanSyncTask < VmaasReposcanSyncTaskBase
+      def plan(repo, *_args)
+        return unless repo.is_a?(::Katello::Repository) && ForemanRhCloud.with_iop_smart_proxy?
+
+        plan_self
+      end
+
+      def self.subscribe
+        return unless defined?(::Katello)
```
jeremylenz: Does something break without this? The other two `def self.subscribe`s in rh_cloud don't have it.

lib/insights_cloud/async/vmaas_reposcan_sync_task.rb
```diff
@@ -0,0 +1,23 @@
...
+
+        'Actions::Katello::Repository::Sync'.constantize
```
jeremylenz: This seems to execute as a concurrence:

image

Note the blue box around VmaasReposcanTask and another blue box around the other ~12 tasks. That means this will get kicked off at the same time those other tasks are running. I don't think that's what we want -- doesn't the vulnerability service need the information in place about what new content was synced?

As far as I know, there's nothing you can pass to `def self.subscribe` to make it a sequence instead of a concurrence. 

Can you maybe try overwriting the `plan` method of `Actions::Katello::Repository::Sync` instead? I haven't tried it, but it would be something like

```
def plan
  sequence do
    super
    plan_action(InsightsCloud::Async::VmaasReposcanSyncTask)
  end
end

File: lib/insights_cloud/async/vmaas_reposcan_sync_task.rb (hunk 3 above)
Comment (ShimShtein) [reply lib/insights_cloud/async/vmaas_reposcan_sync_task.rb]: I think now that we don't have two events to follow, we don't really need to split to the base at all. I think we can merge the classes for now, and extract the base only if we get the second event to follow.

File: lib/insights_cloud/async/vmaas_reposcan_sync_task.rb (hunk 5 above)
Comment (ShimShtein) [reply lib/insights_cloud/async/vmaas_reposcan_sync_task.rb]: I am pretty sure that `super` will not be enough here. According to the [code](https://github.com/Dynflow/dynflow/blob/f8de9af18716bfd7c35c17761baad8dc1bea1dbe/lib/dynflow/action.rb#L544) it looks like it wraps all the subscribers into the same concurrent action.

I suggest one of the following solutions:
1. Use the `finalize` phase (not great, but will do the trick)
2. Use output from the original action (not sure how we can do it, but it should be possible)
3. Subscribe to `Katello::Repository::SyncHook`, which should run [after](https://github.com/Katello/katello/blob/6d5f7f7a9fa52ed8e798d136af3ca07ef7c6fb02/app/lib/actions/katello/repository/sync.rb#L61) the sync is complete.
4. Stop using the action itself and subscribe directly as an `after_sync` action: https://github.com/Katello/katello/blob/6d5f7f7a9fa52ed8e798d136af3ca07ef7c6fb02/app/models/katello/repository.rb#L843

Comment (jeremylenz) [reply lib/insights_cloud/async/vmaas_reposcan_sync_task.rb]: I don't really understand how (4) works, but those sound great! thanks @ShimShtein

Comment (ShimShtein) [reply lib/insights_cloud/async/vmaas_reposcan_sync_task.rb]: I think the easiest one would be the `SyncHook` method (3). Also we will need to do an RFE to allow executing subscriptions after the main action plan.

Comment (nofaralfasi) [reply lib/insights_cloud/async/vmaas_reposcan_sync_task.rb]: After discussing with Shim, option 3 seems to be the best choice.

File: lib/insights_cloud/async/vmaas_reposcan_sync_task.rb (hunk 3 above)
Comment (nofaralfasi) [reply lib/insights_cloud/async/vmaas_reposcan_sync_task.rb]: After discussing it with Shim, we decided to keep the file in its current location, along with the others, to avoid overcomplicating the solution.

Comment (jeremylenz) [reply lib/insights_cloud/async/vmaas_reposcan_sync_task.rb]: ok that's fine for now :+1:

**Comment by ShimShtein** [comment]
@nofaralfasi Can you fix rubocop issues please?

File: lib/insights_cloud/async/vmaas_reposcan_sync.rb [hunk 6]
```diff
@@ -0,0 +1,58 @@
...
+          logger.error("VMaaS reposcan sync: missing repository id in payload: #{repo.inspect}")
+          return
+        end
+        return unless ::ForemanRhCloud.with_iop_smart_proxy?
+
+        plan_self
+      end
+
+      # Subscribe to Katello repository sync hook action, if available
+      def self.subscribe
```
ShimShtein: Nit: can you put this method before the instance method definitions? In this case before the `plan`?

lib/insights_cloud/async/vmaas_reposcan_sync.rb
```diff
@@ -0,0 +1,58 @@
...
+module InsightsCloud
+  module Async
+    # Triggers VMaaS reposcan sync via IoP gateway when repositories are synced
+    class VmaasReposcanSync < ::Actions::EntryAction
+      include ::ForemanRhCloud::CertAuth
+
+      def plan(repo, *_args)
+        repo_id = repo.is_a?(Hash) ? (repo[:id] || repo['id']) : nil
+        unless repo_id
+          logger.error("VMaaS reposcan sync: missing repository id in payload: #{repo.inspect}")
```
ShimShtein:
```suggestion
          logger.error("VMaaS reposcan sync: missing repository id in SyncHook plan parameters: #{repo.inspect}")
```

lib/insights_cloud/async/vmaas_reposcan_sync.rb
```diff
@@ -0,0 +1,58 @@
...
+      def run
+        url = ::InsightsCloud.vmaas_reposcan_sync_url
+
+        response = execute_cloud_request(
+          method: :put,
+          url: url,
+          headers: { 'Content-Type' => 'application/json' }
+        )
+
+        logger.info("VMaaS reposcan sync triggered successfully: #{response.code}")
```
ShimShtein: If the code is not 200, I think we shouldn't say it was a success

jeremylenz: Yes, Maria always tells us never to use "successfully" in user messages :)

lib/insights_cloud/async/vmaas_reposcan_sync.rb
```diff
@@ -0,0 +1,63 @@
...
+        response = execute_cloud_request(
+          method: :put,
+          url: url,
+          headers: { 'Content-Type' => 'application/json' }
+        )
+
+        if response.code >= 200 && response.code < 300
+          logger.info("VMaaS reposcan sync triggered successfully: #{response.code}")
+        else
+          logger.error("VMaaS reposcan sync failed with status: #{response.code}")
```
ShimShtein: I suggest to log the response here and in the `rescue` statement, if the response is available.

ShimShtein (commented):
One nitpick and the tests failures are related

lib/insights_cloud/async/vmaas_reposcan_sync.rb
```diff
@@ -0,0 +1,63 @@
...
+        nil
+      end
+
+      def plan(repo, *_args)
+        repo_id = repo.is_a?(Hash) ? (repo[:id] || repo['id']) : nil
+        unless repo_id
+          logger.error("VMaaS reposcan sync: missing repository id in SyncHook plan parameters: #{repo.inspect}")
+          return
+        end
+        return unless ::ForemanRhCloud.with_iop_smart_proxy?
```
jeremylenz: This guard clause makes more sense at the top of `def plan`. No sense finding a repo we're not going to use, right?

lib/insights_cloud/async/vmaas_reposcan_sync.rb
```diff
@@ -0,0 +1,63 @@
...
+          logger.error("VMaaS reposcan sync failed with status: #{response.code}, body: #{response.body}")
+        end
+
+        response
+      rescue RestClient::ExceptionWithResponse => e
+        logger.error("VMaaS reposcan sync failed: #{e.response&.code} - #{e.response&.body}")
+        raise
+      rescue StandardError => e
+        logger.error("Error triggering VMaaS reposcan sync: #{e.message}, response: #{e.respond_to?(:response) ? e.response : nil}")
+        raise
```
jeremylenz: Instead of just logging, I think it would be useful to put these in the task output, e.g. `output[:message] = xxxx` This way it will be visible in the Dynflow console if we need to troubleshoot.

jeremylenz (commented):
Working better, I see the task gets planned in a sequence now. :+1:

lib/insights_cloud/async/vmaas_reposcan_sync.rb
nofaralfasi: I’ve added both logging and setting output[:message], so the messages appear in the Dynflow console and also in the logs. I can drop the logging if you prefer to avoid duplication.

ShimShtein (approved):
LGTM

lib/insights_cloud/async/vmaas_reposcan_sync.rb
jeremylenz: I think both are useful! let's leave them

jeremylenz (approved):

PR #1083, opened 2025-09-02
Wrap scoped search autocomplete values in quotes when they contain a pipe character

jeremylenz:

Insights rule IDs frequently (almost always) contain the pipe character, `|`. This causes problems when using scoped search to perform a search like

```
rule_id = dnf_install_update_issue|DNF_INSTALL_UPDATE_ISSUE_INFO 
```

because scoped search would take the pipe character as an "OR", and do separate searches for the keywords on each side of the pipe. This means that you don't get any results unless you put the keyword in quotes:
```
rule_id = "dnf_install_update_issue|DNF_INSTALL_UPDATE_ISSUE_INFO"
```

Which is fine, except that scoped search autocomplete _suggests_ the search without the quotes.

```
# Without quotes around the rule id
[1] pry(main)> puts InsightsHit.search_for("rule_id = dnf_install_update_issue|DNF_INSTALL_UPDATE_ISSUE_INFO").to_sql
SELECT ... FROM "insights_hits" ...  WHERE ((("insights_hits"."rule_id" = 'dnf_install_update_issue') OR ("insights_hits"."title" ILIKE '%DNF_INSTALL_UPDATE_ISSUE_INFO%' OR "hosts"."name" ILIKE '%DNF_INSTALL_UPDATE_ISSUE_INFO%')))
=> nil

# With quotes around the rule id
[2] pry(main)> puts InsightsHit.search_for("rule_id = \"dnf_install_update_issue|DNF_INSTALL_UPDATE_ISSUE_INFO\"").to_sql
SELECT "insights_hits".* FROM "insights_hits" WHERE (("insights_hits"."rule_id" = 'dnf_install_update_issue|DNF_INSTALL_UPDATE_ISSUE_INFO'))
=> nil
```

I tried various ways of modifying scoped_search to add the quotes, including external search methods and `value_translation`. But the problem there was that these receive the search keyword values _after_ scoped search already split up the values on either side of the pipe.

The solution here is to have the scoped search autocomplete builder _suggest_ the quotes. This requires redefining a method.

This will affect the behavior of scoped_search application-wide, not just for InsightsHit or foreman_rh_cloud. Ideally scoped_search itself can be updated with this change eventually, and the change would apply to all special operators and not just pipe. For now, I have it only affecting pipe.

Non-IoP
get a host with recommendations
Insights > Recommendations
Perform a search for "rule_id = "
Accept one of the autocomplete suggestions

Before: You get 0 search results. If you put the rule ID in quotes, it works.
Now: Autocomplete suggested values are already in quotes, so it works.

## Summary by Sourcery

Fix scoped_search to correctly handle Insights rule IDs containing '|' and update the InsightsHit host association to use an explicit class name

Bug Fixes:
- Remove '|' from scoped_search tokenizer regex to prevent splitting Insights rule IDs on pipe symbol

Enhancements:
- Specify class_name 'Host::Base' for the host association in InsightsHit model

### DIFF
--- a/app/models/insights_hit.rb
+++ b/app/models/insights_hit.rb
@@ -1,6 +1,6 @@
 class InsightsHit < ApplicationRecord
   include ::Authorizable
-  belongs_to :host
+  belongs_to :host, class_name: 'Host::Base'
   # since the facet is one-to-one association with a host, we can connect
   # through host_id column on both this model and facet.
   belongs_to :insights_facet, foreign_key: 'host_id', primary_key: 'host_id', counter_cache: :hits_count
--- a/lib/foreman_rh_cloud/engine.rb
+++ b/lib/foreman_rh_cloud/engine.rb
@@ -83,6 +83,27 @@ module ForemanRhCloud
         :description => N_('Configure Cloud Connector on given hosts'),
         :proxy_selector_override => ::RemoteExecutionProxySelector::INTERNAL_PROXY
       )
+
+      ScopedSearch::AutoCompleteBuilder.class_eval do
+        # Insights rule IDs always contain a pipe character.
+        # example: hardening_ssh_config_perms|OPENSSH_HARDENING_CONFIG_PERMS
+        # We need to override this method of scoped_search so that autocomplete
+        # will correctly put the value in quotes (otherwise the "|" will be
+        # interpreted as an OR.)
+        # The only change from scoped_search code is adding the | to the regex in the final map.
+        def complete_value_from_db(field, special_values, val)
+          count = 20 - special_values.count
+          completer_scope(field)
+            .where(@options[:value_filter])
+            .where(value_conditions(field.quoted_field, val))
+            .select(field.quoted_field)
+            .limit(count)
+            .distinct
+            .map(&field.field)
+            .compact
+            .map { |v| v.to_s =~ /\s|\|/ ? "\"#{v.gsub('"', '\"')}\"" : v }
+        end
+      end
     end
 
     # Ideally this code belongs to an initializer. The problem is that Katello controllers are not initialized completely until after the end of the to_prepare blocks


app/models/insights_hit.rb
```diff
@@ -1,6 +1,15 @@
...
+    # remove "|" from this regex so it works with Insights rule IDs
+    keyword << next_char while /[^=~<>\s\&\)\(,]/ =~ peek_char
+    KEYWORDS.has_key?(keyword.downcase) ? yield(KEYWORDS[keyword.downcase]) : yield(keyword)
+  end
+end
+
 class InsightsHit < ApplicationRecord
   include ::Authorizable
-  belongs_to :host
+  belongs_to :host, class_name: 'Host::Base'
```
jeremylenz, line 3: Side note: `InsightsHit.search_for(xxx).to_sql` wouldn't work before I added this. This alone took quite a bit of debugging.

app/models/insights_hit.rb
chris1984, line 1: Thanks for adding the comment explaining why the regex was changed.

To make it even clearer for the next person who comes across this, what do you think about adding an example of a rule ID before the filter, just in case we have to reference it down the road. Up to you though.

adamruzicka:
The catch with this is that this applies application-wide, which means searches will behave differently based on whether rh_cloud is enabled or not.

| Query | Before (and without rh cloud) | After | Same |
|-------|--------|-------|------|
| `foo \| bar` | `foo OR bar` | `foo OR bar` | yes |
| `foo\| bar` | `foo OR bar` | `foo\| AND bar` | no |
| `foo \|bar` | `foo OR bar` | `foo OR bar` | yes |
| `foo\|bar` | `foo OR bar` | `foo\|bar` | no |

scoped_search already wraps autocompleted values in quotes in case they contain whitespace ([link](https://github.com/wvanbergen/scoped_search/blob/3c431dff922c57a06989ddde999c71e44430bf33/lib/scoped_search/auto_complete_builder.rb#L218)). Wouldn't it be better to wrap the autocompleted values in quotes if they contain any special operators as well?

app/models/insights_hit.rb
jeremylenz: I added a comment in my new solution.

jeremylenz:
@adamruzicka updated :+1:

adamruzicka:
Why carry the fix here? Why not send it to https://github.com/wvanbergen/scoped_search ?

jeremylenz:
I would also like to have it fixed in scoped_search. This PR can be a temporary change until that is in place.

chris1984:
> Why carry the fix here? Why not send it to https://github.com/wvanbergen/scoped_search ?

I am fine with a fix here for the time being, looking at pull requests and activity in that repo, it seems a fix would sit for a while before being merged.

jeremylenz:
Opened https://github.com/wvanbergen/scoped_search/pull/230

lib/foreman_rh_cloud/engine.rb
```diff
@@ -83,6 +83,27 @@ def self.register_scheduled_task(task_class, cronline)
...
       )
+
+      ScopedSearch::AutoCompleteBuilder.class_eval do
+        # Insights rule IDs always contain a pipe character.
+        # example: hardening_ssh_config_perms|OPENSSH_HARDENING_CONFIG_PERMS
+        # We need to override this method of scoped_search so that autocomplete
+        # will correctly put the value in quotes (otherwise the "|" will be
+        # interpreted as an OR.)
+        # The only change from scoped_search code is adding the | to the regex in the final map.
+        def complete_value_from_db(field, special_values, val)
```
chris1984, line 94: What do you think of something like this to make it look cleaner on that last line?

```ruby
def complete_value_from_db(field, special_values, val)
  limit_count = 20 - special_values.count

  completer_scope(field)
    .where(@options[:value_filter])
    .where(value_conditions(field.quoted_field, val))
    .select(field.quoted_field)
    .limit(limit_count)
    .distinct
    .map(&field.field)
    .compact
    .map { |v| format_value(v) }
end

private

def format_value(value)
  if value.to_s =~ /\s|\|/
    return "\"#{value.gsub('"', '\"')}\""
  else
    return value
  end
end
```

chris1984 (commented):
Looks fine, had to use Gemini to figure out the regex 🤣 will test it tomorrow.

lib/foreman_rh_cloud/engine.rb
jeremylenz: That's certainly cleaner, and would be better if this were a permanent change. But I wanted it to look mostly the same as https://github.com/wvanbergen/scoped_search/blob/3c431dff922c57a06989ddde999c71e44430bf33/lib/scoped_search/auto_complete_builder.rb#L218

since this is temporary until scoped_search is fixed.

nofaralfasi (approved):
I tested the PR and can confirm it works as expected.

lib/foreman_rh_cloud/engine.rb
chris1984: That works for me :)

chris1984 (approved):
Code looks fine and @nofaralfasi tested it

jeremylenz:
Thanks for the reviews all!  I am going to merge this one unless I hear any objections. Will need to revert it once the scoped_search PR is fully in place.

PR #1088, opened 2025-09-11
Conditionally hide UI components and enable inventory sync in IOP mode

nofaralfasi:

* Enable inventory synchronization in local Advisor Engine environments.
* Hide cloud-specific UI elements when `use_iop_mode` is true.
* Conditionally display settings based on IOP mode.
* Show only relevant configuration options for cloud vs. local Advisor environments.

These changes allow inventory synchronization to run in IOP mode while hiding cloud-specific UI components that are not relevant in local Advisor Engine environments.

* The “Sync all inventory status” button is retained for debugging and to handle edge cases.

1. Navigate to the Inventory Upload page (Under **Administrator**).
2. **With Smart-Proxy using IoP mode enabled:**
In the top section of the page:
   * Only the "Sync all inventory status" button should be visible.
   * All other elements, including:
     * Configure Cloud Connector
     * RH Inventory Description
       should be hidden.
3. **With Smart-Proxy without IoP mode:**
   * The page should remain unchanged.

**ToDo:** Consider whether the "Connectivity test" option in the main kebab menu should also be hidden in IOP mode.

## Summary by Sourcery

Enable inventory synchronization in IOP mode and conditionally hide cloud-specific UI components using a new useIopConfig hook, and update component tests to use React Testing Library

New Features:
- Enable inventory synchronization in IOP mode by removing the skip in the inventory hosts sync job

Enhancements:
- Introduce a useIopConfig hook to toggle UI components based on IOP mode
- Hide cloud-specific UI elements (inventory settings, page description, and cloud connector button) when IOP mode is active
- Refactor PageHeader and ToolbarButtons components to conditionally render content and adjust layout

Tests:
- Migrate snapshot tests to React Testing Library using rtlHelpers.renderWithStore
- Add tests for PageHeader, ToolbarButtons, CVECountCell, and InsightsTable covering IOP and non-IOP modes and CSS class checks

Chores:
- Remove obsolete snapshot fixtures and unused mock API hook files


lib/foreman_rh_cloud/plugin.rb
```diff
@@ -13,15 +13,24 @@ def self.register
...
+              setting('allow_auto_insights_mismatch_delete', type: :boolean, description: N_('Enable automatic deletion of mismatched host records from the Red Hat cloud'), default: false, full_name: N_('Automatic mismatch deletion'))
+              setting('obfuscate_inventory_hostnames', type: :boolean, description: N_('Obfuscate host names sent to the Red Hat cloud. (If insights_minimal_data_collection is set to true, this setting is ignored because host names are not included in the report.)'), default: false, full_name: N_('Obfuscate host names'))
+              setting('obfuscate_inventory_ips', type: :boolean, description: N_('Obfuscate ipv4 addresses sent to the Red Hat cloud.  (If insights_minimal_data_collection is set to true, this setting is ignored because host IPv4 addresses are not included in the report.)'), default: false, full_name: N_('Obfuscate host ipv4 addresses.'))
+            end
+            
+            # IOP-specific settings 
+            if ForemanRhCloud.with_iop_smart_proxy?
+              setting('rhc_instance_id', type: :string, description: N_('RHC daemon id'), default: nil, full_name: N_('ID of the RHC(Yggdrasil) daemon'))
```
jeremylenz: We don't run cloud connector in IoP. Why would this be IoP-specific?

lib/foreman_rh_cloud/plugin.rb
```diff
@@ -13,15 +13,24 @@ def self.register
...
+            setting('insights_minimal_data_collection', type: :boolean, default: false, full_name: N_('Minimal data collection'), description: N_('Only include the minimum required data in inventory reports. When this is true, installed packages are excluded from the report regardless of the exclude_installed_packages setting, and host names and IPv4 addresses are excluded from the report regardless of obfuscation settings.'))
+            setting('exclude_installed_packages', type: :boolean, description: N_('Exclude installed packages from being included in inventory reports. (If insights_minimal_data_collection is set to true, this setting is ignored and installed packages are always excluded.)'), default: false, full_name: N_("Exclude installed packages"))
```
jeremylenz: Why have these available in IoP mode?

lib/foreman_rh_cloud/plugin.rb
chris1984: My pr will hide all the settings, once I make it if you want to base it off mine or I can post the code snippet here

jeremylenz: Ohh, I didn't realize this is the UI part only. I didn't expect any IoP checks in the setting definitions here. I expected that this PR would change only the UI, and yours would touch the setting definitions.

nofaralfasi: You’re right, I reverted the changes.

webpack/ForemanInventoryUpload/Components/PageHeader/__tests__/PageHeader.test.js
```diff
@@ -1,13 +1,107 @@
...
 
+// Mock the useAdvisorEngineConfig hook
+jest.mock('../../../../common/Hooks/ConfigHooks', () => ({
+  useAdvisorEngineConfig: jest.fn(),
+}));
+
+import { useAdvisorEngineConfig } from '../../../../common/Hooks/ConfigHooks';
+
+const middlewares = [thunk];
+const mockStore = configureMockStore(middlewares);
```
ShimShtein: We have a helper for this in the latest Foreman: https://github.com/theforeman/foreman/blob/5205ccb014c196e67c822c41d67991925e53fa67/webpack/assets/javascripts/react_app/common/rtlTestHelpers.js#L18, but if we want to cherry-pick it to 3.16, I suppose we should not use it.

webpack/ForemanInventoryUpload/Components/PageHeader/__tests__/PageHeader.test.js
```diff
@@ -1,13 +1,107 @@
...
+
 const fixtures = {
   'render without Props': {},
   /** fixtures, props for the component */
 };
 
 describe('PageHeader', () => {
+  beforeEach(() => {
+    // Reset mock before each test
+    useAdvisorEngineConfig.mockReset();
```
ShimShtein: You can just add the property to the test store, and then you won't have to mock the accessor

ShimShtein (commented):
A couple of small suggestions, but overall looks good.

webpack/ForemanInventoryUpload/Components/PageHeader/__tests__/PageHeader.test.js
nofaralfasi: I’m not sure I fully understand what you mean. Could you clarify?

nofaralfasi: I tried your suggestion to add the property directly to the test store instead of mocking the useAdvisorEngineConfig hook, but it failed due to complex interactions between the hook's dependency on useAPI and the existing Jest mocking infrastructure. The useAPI hook wasn't properly reading from our Redux store because of conflicting mocks in the test environment. Given these technical challenges, I think the current approach with mockUseAdvisorEngineConfig.mockReturnValue() is actually more appropriate here - it's simpler, and avoids the fragile API mocking chain. The hook-level mock effectively tests the same logic (component renders correctly based on IoP mode) without the infrastructure complexity.

nofaralfasi: I was able to replace the direct hook mocking with a store-based testing approach for `use_iop_mode`. The issue was that `useAdvisorEngineConfig` uses `useAPI` from `foremanReact`, which has a global mock that ignores our test store state. My solution: Mock `useAPI` directly and created a shared helper `setupAdvisorEngineTest(mockUseAPI, true/false)` that configures both the mock and store state consistently.

jeremylenz (approved):

ShimShtein:
I have just merged https://github.com/theforeman/foreman/pull/10654, using those new helpers we can stop mocking the `useAPI` function, and just present the correct state to the component. Something similar to 
``` js
    import { rtlHelpers } from '../../../common/rtlTestHelpers';
    const { renderWithStore } = rtlHelpers;

    test('hides inventory settings and description when in IoP mode', () => {
      renderWithStore(
          <PageHeader />,
          {
            ADVISOR_ENGINE_CONFIG: {
                response: { use_iop_mode: true },
          }
      );

      // Core components should still be present
      expect(screen.getByTestId('settings-warning')).toBeTruthy();
      expect(screen.getByTestId('page-title')).toBeTruthy();
      expect(screen.getByTestId('inventory-filter')).toBeTruthy();
      expect(screen.getByTestId('toolbar-buttons')).toBeTruthy();

      // These components should be hidden in IoP mode
      expect(screen.queryByTestId('inventory-settings')).toBeNull();
      expect(screen.queryByTestId('page-description')).toBeNull();
    });
```

I won't block the PR on it, but I think it will create more technical debt.

webpack/common/rtlTestHelpers.js
```diff
@@ -0,0 +1,56 @@
+import React from 'react';
```
ShimShtein: the original `rtlHelpers` from [core](https://github.com/theforeman/foreman/blob/develop/webpack/assets/javascripts/react_app/common/rtlTestHelpers.js) should be available to the plugins. 
@MariaAga You propbably know the best way to reference it from the plugin

webpack/ForemanInventoryUpload/Components/PageHeader/__tests__/__snapshots__/PageHeader.test.js.snap
```diff
@@ -4,13 +4,13 @@ exports[`PageHeader rendering render without Props 1`] = `
 <div
   className="inventory-upload-header"
 >
-  <ConnectedSettingsWarning />
-  <PageTitle />
+  <Component />
```
ShimShtein: Why do we still need the snaps? We should be able to remove them, if we use the RTL way

webpack/ForemanInventoryUpload/Components/PageHeader/__tests__/PageHeader.test.js
```diff
@@ -1,7 +1,32 @@
...
 import { testComponentSnapshotsWithFixtures } from '@theforeman/test';
-
+import { rtlHelpers } from '../../../../common/rtlTestHelpers';
 import PageHeader from '../PageHeader';
 
+const { renderWithStore } = rtlHelpers;
+
+// Mock child components to isolate PageHeader testing
+// This prevents child component complexity from affecting our tests
+jest.mock('../components/SettingsWarning', () => () => (
```
ShimShtein, line 23: @MariaAga what do you think about this practice?

webpack/common/rtlTestHelpers.js
MariaAga: `foremanReact/common/rtlTestHelpers.js`

webpack/ForemanInventoryUpload/Components/PageHeader/__tests__/PageHeader.test.js
MariaAga: If `SettingsWarning` is tested separetely, and there is a test that makes sure that this mock is presented when needed than its ok, so for these mocks I'm missing `expect(screen.getByTestId('toolbar-buttons')).toBeFalse();` to make sure its shown only when needed (unless its always needed, but I assume warning are not always present)

webpack/ForemanInventoryUpload/Components/PageHeader/__tests__/PageHeader.test.js
```diff
@@ -1,13 +1,114 @@
...
+));
 
 describe('PageHeader', () => {
-  describe('rendering', () =>
-    testComponentSnapshotsWithFixtures(PageHeader, fixtures));
+  describe('component behavior', () => {
+    test('renders all components when not in IoP mode', () => {
+      useAdvisorEngineConfig.mockReturnValue(false);
+
+      renderWithStore(<PageHeader />);
```
ShimShtein: Instead of using the mock, you can render the component with the augmented state:

```suggestion
      renderWithStore(
        <PageHeader />,
        {
          "ADVISOR_ENGINE_CONFIG": {
            response: {
              use_iop_mode: false,
            },
          },
        });
```
This will implicitly test the `useAdvisorEngineConfig` accessor too, which is desired.

ShimShtein: @MariaAga as a general idea: should we add a helper method to construct API responses for tests?
Just to hide the internal structure of our `useAPI` implementation.
Something that will look like:

``` js
      renderWithStore(
        <PageHeader />,
        withAPIResponse(
          'ADVISOR_ENGINE_CONFIG',
          {
            use_iop_mode: false,
          }),
      );
```
The whole tree of the API response is not something that the plugin devs should be aware of. They only know the key if the request, and the response they expect to exist in the store.

webpack/ForemanInventoryUpload/Components/PageHeader/components/ToolbarButtons/__tests__/ToolbarButtons.test.js
```diff
@@ -1,60 +1,78 @@
...
-    render(
-      <Provider store={store}>
-        <ToolbarButtons />
-      </Provider>
-    );
-    expect(screen.queryAllByText('Configure cloud connector')).toHaveLength(1);
-    expect(screen.queryAllByText('Sync all inventory status')).toHaveLength(1);
+  test('renders both buttons when subscription connection is enabled and not in IOP mode', () => {
+    useAdvisorEngineConfig.mockReturnValue(false);
+    selectSubscriptionConnectionEnabled.mockReturnValue(true);
```
ShimShtein, line 43: The same comment: you can add the setting to the store rendering function, and then you won't have to mock the selectors.
Mocking the selectors means you never test them properly, and I don't see a good reason to do it, unless the selector is really complicated.

nofaralfasi:
> Instead of using the mock, you can render the component with the augmented state:
> 
> This will implicitly test the `useAdvisorEngineConfig` accessor too, which is desired.

@ShimShtein I tried the approach you suggested, but unfortunately, it didn’t work as expected.
Here’s the error message I’m getting:
```
Unexpected key "ADVISOR_ENGINE_CONFIG" found in preloadedState argument passed to createStore. Expected to find one of the known reducer keys instead: "bookmarksPF4", "hosts", "notifications", "toasts", "passwordStrength", "breadcrumbBar", "layout", "diffModal", "editor", "templates", "settingRecords", "personalAccessTokens", "confirmModal", "router", "extendable", "auditsPage", "foremanModals", "intervals", "API". Unexpected keys will be ignored.
```

I also tried nesting it under the API key, but with that setup, the tests are still failing, so it doesn’t seem to help either.
Do you have any other suggestions on how to structure this?

webpack/ForemanInventoryUpload/Components/PageHeader/__tests__/PageHeader.test.js
MariaAga: That could be a nice addition I agree

ShimShtein:
You need to recreate the structure of the existing API redux tree:
image
Can you please try it? In the ideal world we will have a helper that will put the response exactly where we want it, but for now, we will need to do it manually

jeremylenz:
Maybe we should look at adding something like Katello's renderWithRedux: https://github.com/jeremylenz/katello/blob/b886d85ae80876855f7da333a77441354626c916/webpack/test-utils/react-testing-lib-wrapper.js#L23

nofaralfasi:
> You need to recreate the structure of the existing API redux tree: 
> Can you please try it? In the ideal world we will have a helper that will put the response exactly where we want it, but for now, we will need to do it manually

@ShimShtein I tried it with this code, but I’m still seeing errors for components that should be hidden:
```
    test('hides inventory settings and description when in IoP mode', () => {
      renderWithStore(<PageHeader />, {
        API: {
          ADVISOR_ENGINE_CONFIG: {
            payload: { url: '/api/v2/rh_cloud/advisor_engine_config' },
            response: { use_iop_mode: true },
            status: 'RESOLVED',
          },
        },
      });
```

MariaAga:
@nofaralfasi I ran the tests with 
```
    test('hides inventory settings and description when in IoP mode', () => {
      renderWithStore(<PageHeader />, {
        API: {
          ADVISOR_ENGINE_CONFIG: {
            payload: { url: '/api/v2/rh_cloud/advisor_engine_config' },
            response: { use_iop_mode: true },
            status: 'RESOLVED',
          },
        },
      });
 ```
 on my env and they pass. What error do you see?

**Comment by nofaralfasi** [comment]
> What error do you see?

```
 FAIL  ../../foreman_rh_cloud/webpack/ForemanInventoryUpload/Components/PageHeader/__tests__/PageHeader.test.js (5.887 s, 172 MB heap size)
  ● PageHeader › component behavior › hides inventory settings and description when in IoP mode

    expect(received).toBeNull()

    Received: <div data-testid="inventory-settings">InventorySettings</div>

      69 |
      70 |       // These components should be hidden in IoP mode
    > 71 |       expect(screen.queryByTestId('inventory-settings')).toBeNull();
         |                                                          ^
      72 |       expect(screen.queryByTestId('page-description')).toBeNull();
      73 |     });
      74 |

      at Object.<anonymous> (../../foreman_rh_cloud/webpack/ForemanInventoryUpload/Components/PageHeader/__tests__/PageHeader.test.js:71:58)

```

File: webpack/ForemanInventoryUpload/Components/PageHeader/__tests__/PageHeader.test.js [hunk 10]
```diff
@@ -1,13 +1,114 @@
-import { testComponentSnapshotsWithFixtures } from '@theforeman/test';
-
+import React from 'react';
+import { screen } from '@testing-library/react';
+import { rtlHelpers } from 'foremanReact/common/rtlTestHelpers';
 import PageHeader from '../PageHeader';
+import { useAdvisorEngineConfig } from '../../../../common/Hooks/ConfigHooks';
```
jeremylenz: Fyi, all these are renamed to `useIopConfig` now :) time to rebase!

webpack/ForemanInventoryUpload/Components/PageHeader/components/ToolbarButtons/ToolbarButtons.js
```diff
@@ -4,19 +4,21 @@ import SyncButton from '../SyncButton';
 import CloudConnectorButton from '../CloudConnectorButton';
 import './toolbarButtons.scss';
 import { selectSubscriptionConnectionEnabled } from '../../../InventorySettings/InventorySettingsSelectors';
+import { useIopConfig } from '../../../../../common/Hooks/ConfigHooks';
 
 const ToolbarButtons = () => {
   const subscriptionConnectionEnabled = useSelector(
     selectSubscriptionConnectionEnabled
   );
+  const useIopMode = useIopConfig();
```
jeremylenz: Variable names that start with `use` are assumed to be Hooks. For that reason I'd prefer to call it something else.

```suggestion
  const isIop = useIopConfig();
```

jeremylenz (approved):
LGTM, thanks @nofaralfasi!

Will wait a bit to merge in case any others have additional comments.

jeremylenz:
Merging. thanks @nofaralfasi !

PR #1089, opened 2025-09-12
SAT-38175 - Ignore settings in IOP setups for report generation

chris1984:

* In SAT-36556 we are hiding everything except sync inventory status in the UI.
* This pr ignores the settings for the various actions so choosing any of the things to include in the report upload does not apply regardless of the settings. This is only for IOP mode. 

* Made sure it works correctly and none of the changes are applied for non IOP setups and we keep the functionality the same
* UI PR is here https://github.com/theforeman/foreman_rh_cloud/pull/1088

* Install IOP on a devel box
* Check out PR
* Toggle various settings that would change the info in a report and generate a report and see that it's still present
* Install a devel box without IOP
* Toggle various settings that would change the info in a report and generate a report and make sure they are working correctly like `exclude_packages`, `minimal_data`, or one of the obfuscation settings.


app/controllers/foreman_inventory_upload/uploads_settings_controller.rb
```diff
@@ -2,20 +2,23 @@ module ForemanInventoryUpload
...
-        excludePackagesEnabled: Setting[:exclude_installed_packages],
-        allowAutoInsightsMismatchDelete: Setting[:allow_auto_insights_mismatch_delete],
+        hostObfuscationEnabled: Foreman.settings.find('obfuscate_inventory_hostnames') ? Setting[:obfuscate_inventory_hostnames] : false,
+        ipsObfuscationEnabled: Foreman.settings.find('obfuscate_inventory_ips') ? Setting[:obfuscate_inventory_ips] : false,
+        excludePackagesEnabled: Foreman.settings.find('exclude_installed_packages') ? Setting[:exclude_installed_packages] : false,
+        allowAutoInsightsMismatchDelete: Foreman.settings.find('allow_auto_insights_mismatch_delete') ? Setting[:allow_auto_insights_mismatch_delete] : false,
         CloudConnectorStatus: ForemanInventoryUpload::UploadsSettingsController.cloud_connector_status,
         lastSyncTask: last_successful_inventory_sync_task,
       }, status: :ok
     end
```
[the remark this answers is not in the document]
chris1984: Not needed in this case. We have a warning already for a setting not being found for normal operation outside of this mode
```ruby
def find(name)
    logger. Warn("Setting is not initialized yet, requested value for #{name} will be always nil") unless ready?
    @settings[name.to_s]
  end
```

app/controllers/insights_cloud/settings_controller.rb
```diff
@@ -5,7 +5,9 @@ def show
     end
 
     def update
-      Setting[:allow_auto_insights_sync] = settings_params
+      if Foreman.settings.find('allow_auto_insights_sync')
+        Setting[:allow_auto_insights_sync] = settings_params
+      end
       render_setting(:insightsSyncEnabled, :allow_auto_insights_sync)
     end
```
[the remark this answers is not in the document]
chris1984: Not needed in this case. We have a warning already for a setting not being found for normal operation outside of this mode
```ruby
def find(name)
    logger. Warn("Setting is not initialized yet, requested value for #{name} will be always nil") unless ready?
    @settings[name.to_s]
  end
```

app/controllers/foreman_inventory_upload/uploads_controller.rb
```diff
@@ -22,7 +22,9 @@ def download_file
 
     def enable_cloud_connector
       # Set the autoupload to true, since it's required by the feature.
-      Setting[:allow_auto_inventory_upload] = true
+      if Foreman.settings.find('allow_auto_inventory_upload')
+        Setting[:allow_auto_inventory_upload] = true
+      end
```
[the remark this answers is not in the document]
chris1984: Not needed in this case. We have a warning already for a setting not being found for normal operation outside of this mode
```ruby
def find(name)
    logger. Warn("Setting is not initialized yet, requested value for #{name} will be always nil") unless ready?
    @settings[name.to_s]
  end
```

app/controllers/foreman_inventory_upload/uploads_controller.rb
```diff
@@ -22,7 +22,9 @@ def download_file
 
     def enable_cloud_connector
```
ShimShtein, line 23: This method should not be called in IoP use case at all

lib/foreman_rh_cloud/plugin.rb
```diff
@@ -12,16 +12,18 @@ def self.register
...
-            setting('include_parameter_tags', type: :boolean, description: N_('Should import include parameter tags from Foreman?'), default: false, full_name: N_('Include parameters in insights-client reports'))
-            setting('rhc_instance_id', type: :string, description: N_('RHC daemon id'), default: nil, full_name: N_('ID of the RHC(Yggdrasil) daemon'))
-            setting('insights_minimal_data_collection', type: :boolean, default: false, full_name: N_('Minimal data collection'), description: N_('Only include the minimum required data in inventory reports for uploading to Red Hat cloud. When this is true, installed packages are excluded from the report regardless of the exclude_installed_packages setting, and host names and IPv4 addresses are excluded from the report regardless of obfuscation settings.'))
+          unless ForemanRhCloud.with_iop_smart_proxy?
+            category(:rh_cloud, N_('Insights')) do
+              setting('allow_auto_inventory_upload', type: :boolean, description: N_('Enable automatic upload of your host inventory to the Red Hat cloud'), default: true, full_name: N_('Automatic inventory upload'))
```
ShimShtein: @ekohl I think we should start handling conditional settings properly here. 
I would suggest adding two more parameters to setting definition, like this:
```suggestion
              setting('allow_auto_inventory_upload', type: :boolean, description: N_('Enable automatic upload of your host inventory to the Red Hat cloud'), default: true, full_name: N_('Automatic inventory upload'), disabled: -> { ForemanRhCloud.with_iop_smart_proxy? }, disabled_value: true)
```
and then in the `[]` method return the disabled value if the disabled callback returns true. Of course, in the UI we can filter out the disabled settings.
WDYT?

ShimShtein (commented):
Let's hear @ekohl's opinion about adding functionality into the settings infrastructure, before we merge this code.
IMHO it should be implemented on the core level.

nofaralfasi (commented):
Tested on both IoP and non-IoP environments, and I can confirm everything works as expected.
That said, I agree with Shim’s suggestion. It would be cleaner to extend the settings infrastructure on Foreman core to handle conditional settings directly.

lib/foreman_rh_cloud/plugin.rb
```diff
@@ -12,16 +12,18 @@ def self.register
...
-            setting('include_parameter_tags', type: :boolean, description: N_('Should import include parameter tags from Foreman?'), default: false, full_name: N_('Include parameters in insights-client reports'))
-            setting('rhc_instance_id', type: :string, description: N_('RHC daemon id'), default: nil, full_name: N_('ID of the RHC(Yggdrasil) daemon'))
-            setting('insights_minimal_data_collection', type: :boolean, default: false, full_name: N_('Minimal data collection'), description: N_('Only include the minimum required data in inventory reports for uploading to Red Hat cloud. When this is true, installed packages are excluded from the report regardless of the exclude_installed_packages setting, and host names and IPv4 addresses are excluded from the report regardless of obfuscation settings.'))
+          unless ForemanRhCloud.with_iop_smart_proxy?
```
ekohl: This is a runtime thing. Implementing it this way means you need to restart Foreman to get these settings. I strongly dislike that. IMHO the whole idea is that you can add a Smart Proxy to discover new functionality, but also remove it. :-1: from me on implementing it this way

ekohl (commented):
I agree with @ShimShtein that this should be avoided and implemented in a proper way.

The disabled setting concept is an interesting one that could be an option, but I haven't fully thought it through.

lib/foreman_rh_cloud/plugin.rb
chris1984: How should I go about solving it then?

chris1984: About your comment on thinking it through, should I wait for your response or start adding what Shim suggested to the setting infrastructure in foreman core?

jeremylenz: We already check the existence of the IoP smart proxy in 2 other places in plugin.rb (when we decide how to build the main nav items, and when registering the Ping extension). IMO this is just a consequence of the design decision to define IoP as "is there an IoP smart proxy?" rather than what we did previously, which was to look at the installer flag. cc @ehelms

app/controllers/foreman_inventory_upload/uploads_controller.rb
```diff
@@ -22,7 +22,9 @@ def download_file
...
       # Set the autoupload to true, since it's required by the feature.
-      Setting[:allow_auto_inventory_upload] = true
+      if Foreman.settings.find('allow_auto_inventory_upload')
```
jeremylenz: Would prefer all these `Foreman.settings.find` to be replaced by a wrapper method whose name shows its intent, e.g. 
```
def set_non_iop_setting(setting_name, value)
  if Foreman.settings.find(setting_name)
    Setting[setting_name] = value
  end
end
```

ShimShtein: Let's first try and hide this whole mess into the settings infrastructure as I have suggested.

ekohl: As @ShimShtein said, let's avoid it. IMHO we should always have the settings defined. If you don't you'll end up with orphaned database entries. While it's only a debug statement, it does log something to clean up your DB:
https://github.com/theforeman/foreman/blob/8d05fc4561bea27643c572caec2c972b69a06975/app/registries/setting_registry.rb#L185-L186

lib/foreman_rh_cloud/plugin.rb
ekohl: > IMO this is just a consequence of the design decision to define IoP as "is there an IoP smart proxy?" rather than what we did previously, which was to look at the installer flag.

The reason to avoid an installer flag is that you can at runtime add functionality. It makes the application more dynamic, but this should play nicer into our containerization strategy.

If you don't need to redeploy your application then it's easier for the admin. Remember that there is an increasing demand for a high available solution. Restarting application servers is downtime. If you have multiple application servers behind a load balancer you need to restart them all. That can also lead to inconsistency.

ekohl:
We have more cases where there are settings that are useless if a certain condition is true. In https://github.com/theforeman/foreman/blob/develop/config/initializers/f_foreman_settings_notification.rb we have `rss_enable` and `rss_url`. The URL is simply not used if RSS is disabled.

When reading this PR I actually went through the thought process: should we hide/disable `rss_url` if `rss_enable` is false? My conclusion was that we shouldn't because a user workflow may be that they first want to have all settings correct before it's enabled. In the case of IoP it may mean they first want to ensure that no data is shared with the cloud before enabling the feature. Hiding the settings beforehand makes this impossible. So I think you're trying to solve a problem that doesn't really exist IMHO.

jeremylenz:
> In the case of IoP it may mean they first want to ensure that no data is shared with the cloud before enabling the feature. Hiding the settings beforehand makes this impossible. So I think you're trying to solve a problem that doesn't really exist IMHO.

So are we going to add functionality to Settings, or are we going to leave them alone? If we leave them alone, at a minimum we still need to update setting descriptions and behavior. It still wouldn't make sense to send reports to Vuln with no package profiles.

app/controllers/foreman_inventory_upload/uploads_controller.rb
jeremylenz: I like the idea of adding functionality to Settings; I think it's clear we need it and are working around its absence.

jeremylenz:
@chris1984 Ewoud prefers not changing setting definitions or functionality, so I vote to move forward this way

1. Add to the description of each relevant setting "Ignored when using local Insights" <-- will be translated to "Ignored when using local Red Hat Lightspeed"
2. Remove the IoP checks from plugin.rb
3. Keep the IoP checks when deciding about what to include in the reports

app/controllers/foreman_inventory_upload/uploads_controller.rb
chris1984: Fixed here `before_action :require_non_iop_smart_proxy, only: [:enable_cloud_connector]`

jeremylenz (approved):
I ran a report on my IoP box with all the settings turned on

```
[2] pry(main)> Setting[:obfuscate_inventory_hostnames] = true
=> true
[3] pry(main)> Setting[:obfuscate_inventory_hostnames]
=> true
[4] pry(main)> Setting[:obfuscate_inventory_ips] = true
=> true
[5] pry(main)> Setting[:insights_minimal_data_collection] = true
=> true
[6] pry(main)> Setting[:exclude_installed_packages] = true
=> true
[7] pry(main)> exit
[vagrant@ip-10-0-168-217 foreman]$ organization_id=1 bundle exec rake rh_cloud_inventory:report:generate
```

And it correctly ignored them. The report had unobfuscated hostnames and installed packages.

works for me! ACK 👍

PR #1098, opened 2025-09-19
SAT-37817 - Hide Recommendations page for non-registered hosts

chris1984:
* Remove the dependency on host-specific response data
* This approach is consistent with how other components in the codebase handle this configuration and ensures that all hosts (with or without insights_facet) properly determine which advisor engine to display based on the global system settings.
* Now the Insights/Recommendations tab will be hidden if either:
  * The host is not a RHEL host (existing logic), OR
  * The host doesn't have an insights_facet (new logic)

* This means that for a host to show the Recommendations tab, it must be:
  * A RHEL host (Red Hat Enterprise Linux), AND
  * Have an insights_facet (indicated by the presence of insights_attributes in the response)

* Made sure it still works on non-iop systems

* Setup a devel box with IOP
* Click on the devel box hostname since that should not have a subscription/insights_facet and see that it pulls up the non-iop recommendations page
Screenshot 2025-09-19 at 14-53-42 ip-10-0-168-128 rhos-01 prod psi rdu2 redhat com

* Check out PR
* Clear your browser cache
* Visit the same host and see we have don't have a recommendations page

### DIFF
--- a/webpack/InsightsHostDetailsTab/NewHostDetailsTab.js
+++ b/webpack/InsightsHostDetailsTab/NewHostDetailsTab.js
@@ -131,28 +131,15 @@ const IopInsightsTabWrapped = props => (
 );
 
 const InsightsTab = props => {
-  const { response } = props;
-  const isLocalAdvisorEngine =
-    // eslint-disable-next-line camelcase
-    response?.insights_attributes?.use_iop_mode;
+  const isLocalIop = useAdvisorEngineConfig();
 
-  return isLocalAdvisorEngine ? (
+  return isLocalIop ? (
     <IopInsightsTabWrapped {...props} />
   ) : (
     <NewHostDetailsTab {...props} />
   );
 };
 
-InsightsTab.propTypes = {
-  response: PropTypes.shape({
-    insights_attributes: {
-      use_iop_mode: PropTypes.bool,
-    },
-  }),
-};
-
-InsightsTab.defaultProps = {
-  response: {},
-};
+InsightsTab.defaultProps = {};
 
 export default InsightsTab;
--- a/webpack/ForemanRhCloudFills.js
+++ b/webpack/ForemanRhCloudFills.js
@@ -4,7 +4,11 @@ import { translate as __ } from 'foremanReact/common/I18n';
 import InventoryAutoUploadSwitcher from './ForemanInventoryUpload/SubscriptionsPageExtension/InventoryAutoUpload';
 import NewHostDetailsTab from './InsightsHostDetailsTab/NewHostDetailsTab';
 import { InsightsTotalRiskChartWrapper } from './InsightsHostDetailsTab/InsightsTotalRiskChartWrapper';
-import { isNotRhelHost, vulnerabilityDisabled } from './ForemanRhCloudHelpers';
+import {
+  isNotRhelHost,
+  vulnerabilityDisabled,
+  hasNoInsightsFacet,
+} from './ForemanRhCloudHelpers';
 import CVEsHostDetailsTabWrapper from './CVEsHostDetailsTab/CVEsHostDetailsTab';
 
 const fills = [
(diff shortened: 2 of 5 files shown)


webpack/InsightsHostDetailsTab/NewHostDetailsTab.js
```diff
@@ -131,28 +131,17 @@ const IopInsightsTabWrapped = props => (
...
 
 const InsightsTab = props => {
-  const { response } = props;
-  const isLocalAdvisorEngine =
-    // eslint-disable-next-line camelcase
-    response?.insights_attributes?.use_iop_mode;
+  const isLocalAdvisorEngine = useAdvisorEngineConfig();
 
-  return isLocalAdvisorEngine ? (
+  return isLocalAdvisorEngine === true ? (
```
jeremylenz: I agree with AI here. Is this `===` really needed? If we're sure that `isLocalAdvisorEngine` returns a Boolean (and we should be, it's our code..) then a simple `isLocalAdvisorEngine ?` should be fine here, unless I am missing something.

webpack/InsightsHostDetailsTab/NewHostDetailsTab.js
```diff
@@ -131,28 +131,17 @@ const IopInsightsTabWrapped = props => (
...
 };
 
-InsightsTab.propTypes = {
-  response: PropTypes.shape({
-    insights_attributes: {
-      use_iop_mode: PropTypes.bool,
-    },
-  }),
-};
+InsightsTab.propTypes = {};
```
jeremylenz: You can delete these propTypes

jeremylenz:
> * Consider renaming `isLocalAdvisorEngine` to something like `isIopModeEnabled` to better reflect that it’s driven by your `useAdvisorEngineConfig` hook.

Not in the scope of this PR, but I've been wanting to do this. It's no longer just "advisor," so we should call it `useIopConfig` and `isIop` or some such.

chris1984:
> > * Consider renaming `isLocalAdvisorEngine` to something like `isIopModeEnabled` to better reflect that it’s driven by your `useAdvisorEngineConfig` hook.
> 
> Not in the scope of this PR, but I've been wanting to do this. It's no longer just "advisor," so we should call it `useIopConfig` and `isIop` or some such.

Changed the names, let me know what you think? Also updated the changes you and AI wanted.

jeremylenz (commented):
This works, but thinking about it more ..

- I don't think an unregistered host will _ever_ have recommendations.
- For Vulnerabilities, we just don't show the tab at all.

I think we should just not show the tab at all for unregistered hosts. Thoughts?

chris1984:
> This works, but thinking about it more ..
> 
> * I don't think an unregistered host will _ever_ have recommendations.
> * For Vulnerabilities, we just don't show the tab at all.
> 
> I think we should just not show the tab at all for unregistered hosts. Thoughts?

Makes sense, updated the pr to:
Now the Insights/Recommendations tab will be hidden if either:

The host is not a RHEL host (existing logic), OR
The host doesn't have an insights_facet (new logic)

This means that for a host to show the Recommendations tab, it must be:

A RHEL host (Red Hat Enterprise Linux), AND
Have an insights_facet (indicated by the presence of insights_attributes in the response)

jeremylenz (changes requested):
I checked out the PR
My host has insights_attributes in the response, but I don't see a Recommendations tab:

Image

chris1984:
> I checked out the PR My host has insights_attributes in the response, but I don't see a Recommendations tab:
> 
> Image

@jeremylenz updated, let me know what you think. I tested it on a host that didn't have an `insights_facet` and one that did, and it worked as intended now.

jeremylenz (approved):
Works well

thanks @chris1984 !

ACK 👍

PR #1102, opened 2025-09-26
Generate single-host reports without shell commands or rake task

jeremylenz:

As part of the IoP Vulnerability service, package profile upload triggers a single-host inventory report upload. Previously this was accomplished the same way as other inventory reports, meaning it was run in a shell process which ran a rake task, and report generation output was captured for display in the web UI. However, with hundreds or thousands of hosts it seems this caused large performance bottlenecks.

This change removes some of that overhead by extracting the report generation to a new Dynflow task and running it directly. This means no shell process and no rake task, hopefully eliminating that bloat and improving performance.

Both this and BulkGenerateApplicability are triggered by package profile upload and can run concurrently. But since inventory reports only require an updated package profile and not updated applicability calculation, I think this is fine.

I had to fix a line in slice.rb since my host didn't have a content source

yum -y upgrade jq or yum -y downgrade jq
Watch Foreman logs and verify that `ForemanInventoryUpload::Async::SingleHostReportJob` completes with success
Inspect Dynflow console inputs and outputs and make sure everything looks ok
Extract the report tar and make sure it looks ok.

## Summary by Sourcery

Support generating single-host inventory reports via a new async job pipeline, replace shell/rake-based approach, and improve yum repo URL handling.

New Features:
- Introduce SingleHostReportJob to orchestrate asynchronous single-host report generation and upload
- Add GenerateSingleHostReport action to render archived report for a specific host
- Update host report endpoint to invoke SingleHostReportJob instead of the generic GenerateReportJob

Enhancements:
- Provide fallback to SmartProxy primary pulp_content_url for yum repository reporting when content_source is unavailable

### DIFF
--- a/lib/foreman_inventory_upload/async/create_missing_insights_facets.rb
+++ b/lib/foreman_inventory_upload/async/create_missing_insights_facets.rb
@@ -1,0 +1,29 @@
+module ForemanInventoryUpload
+  module Async
+    class CreateMissingInsightsFacets < ::Actions::EntryAction
+      def plan(organization_id)
+        plan_self(organization_id: organization_id)
+      end
+
+      def run
+        organization = ::Organization.find(input[:organization_id])
+        hosts_without_facets = ::ForemanInventoryUpload::Generators::Queries.for_org(organization, hosts_query: 'null? insights_uuid')
+        facet_count = hosts_without_facets.count
+        hosts_without_facets.each do |batch|
+          facets = batch.pluck(:id, 'katello_subscription_facets.uuid').map do |host_id, uuid|
+            {
+              host_id: host_id,
+              uuid: uuid,
+            }
+          end
+          # We don't need to validate the facets here as we create the necessary fields.
+          # rubocop:disable Rails/SkipsModelValidations
+          InsightsFacet.upsert_all(facets, unique_by: :host_id) unless facets.empty?
+          # rubocop:enable Rails/SkipsModelValidations
+        end
+        output[:result] = facet_count.zero? ? _("There were no missing Insights facets") : format(_("Missing Insights facets created: %s"), facet_count)
+        Rails.logger.info output[:result]
+      end
+    end
+  end
+end
--- a/lib/foreman_inventory_upload/async/generate_host_report.rb
+++ b/lib/foreman_inventory_upload/async/generate_host_report.rb
@@ -1,0 +1,20 @@
+module ForemanInventoryUpload
+  module Async
+    class GenerateHostReport < ::Actions::EntryAction
+      def plan(base_folder, organization_id, filter)
+        plan_self(
+          base_folder: base_folder,
+          organization_id: organization_id,
+          filter: filter
+        )
+        input[:target] = File.join(base_folder, ForemanInventoryUpload.facts_archive_name(input[:organization_id], input[:filter]))
+      end
+
+      def run
+        archived_report_generator = ForemanInventoryUpload::Generators::ArchivedReport.new(input[:target])
+        archived_report_generator.render(organization: input[:organization_id], filter: input[:filter])
+        output[:result] = "Generated #{input[:target]} for organization id #{input[:organization_id]}"
+      end
+    end
+  end
+end
--- a/lib/foreman_inventory_upload/async/single_host_report_job.rb
+++ b/lib/foreman_inventory_upload/async/single_host_report_job.rb
@@ -1,0 +1,20 @@
+module ForemanInventoryUpload
+  module Async
+    class SingleHostReportJob < HostInventoryReportJob
+      def plan(base_folder, organization_id, host_id)
+        input[:host_id] = host_id
+        super(base_folder, organization_id, "id=#{input[:host_id]}")
+      end
+
+      def hostname(host_id)
+        host = ::Host.find_by(id: host_id)
+        host&.name
+      end
+
+      def humanized_name
+        hostname_result = hostname(input[:host_id])
+        hostname_result.present? ? format(_("Single-host report job for host %s"), hostname_result) : _("Single-host report job")
+      end
+    end
+  end
+end
--- a/app/controllers/concerns/insights_cloud/package_profile_upload_extensions.rb
+++ b/app/controllers/concerns/insights_cloud/package_profile_upload_extensions.rb
@@ -15,11 +15,10 @@ module InsightsCloud
       logger.debug("Generating host-specific report for host #{@host.name}")
 
       ForemanTasks.async_task(
-        ForemanInventoryUpload::Async::GenerateReportJob,
+        ForemanInventoryUpload::Async::SingleHostReportJob,
         ForemanInventoryUpload.generated_reports_folder,
         @host.organization_id,
-        false,
-        "id=#{@host.id}"
+        @host.id
       )
 
       # in IoP case, the hosts are identified by the sub-man ID, and we can assume they already
--- a/lib/foreman_inventory_upload/async/generate_report_job.rb
+++ b/lib/foreman_inventory_upload/async/generate_report_job.rb
@@ -5,7 +5,7 @@ module ForemanInventoryUpload
         "report_for_#{label}"
       end
 
-      def plan(base_folder, organization_id, disconnected, hosts_filter = nil)
+      def plan(base_folder, organization_id, disconnected = false, hosts_filter = nil)
         sequence do
           super(
             GenerateReportJob.output_label("#{organization_id}#{hosts_filter.empty? ? nil : "[#{hosts_filter.to_s.parameterize}]"}"),
--- a/lib/foreman_inventory_upload/async/host_inventory_report_job.rb
+++ b/lib/foreman_inventory_upload/async/host_inventory_report_job.rb
@@ -1,0 +1,39 @@
+module ForemanInventoryUpload
+  module Async
+    class HostInventoryReportJob < ::Actions::EntryAction
+      def plan(base_folder, organization_id, hosts_filter = "", upload = true)
+        sequence do
+          plan_action(
+            GenerateHostReport,
+            base_folder,
+            organization_id,
+            hosts_filter
+          )
+          if upload
+            plan_action(
+              QueueForUploadJob,
+              base_folder,
+              ForemanInventoryUpload.facts_archive_name(organization_id, hosts_filter),
+              organization_id
+            )
+          end
+
+          if ForemanRhCloud.with_iop_smart_proxy?
+            plan_action(
+              CreateMissingInsightsFacets,
+              organization_id
+            )
+          end
+        end
+      end
+
+      def humanized_name
+        _("Host inventory report job")
+      end
+
+      def organization_id
+        input[:organization_id]
+      end
+    end
+  end
+end
--- a/lib/foreman_inventory_upload/generators/slice.rb
+++ b/lib/foreman_inventory_upload/generators/slice.rb
@@ -256,7 +256,7 @@ module ForemanInventoryUpload
 
         @stream.array_field('yum_repos') do
           host.content_facet.bound_repositories.each_with_index do |repo, index|
-            report_yum_repo(host.content_source.load_balancer_pulp_content_url, repo)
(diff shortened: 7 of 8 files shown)


jeremylenz:
@adamruzicka I can't tag you for review, but have a look :)

jeremylenz:
I didn't realize that `QueueForUploadJob` calls `UploadReportJob` which also uses ShellProcess and uploads via uploader.sh. I'll have to think about what to do about that.

parthaa:
I do like the approach here. My only issue is we cant see the Shell output in Administer -> Inventory Upload (which is not a big deal imo).
We should use the same approach for Upload (may be in a separate PR.)

parthaa:
where are you accounting for this logic
https://github.com/theforeman/foreman_rh_cloud/blob/develop/lib/tasks/rh_cloud_inventory.rake#L52-L66

Added by https://github.com/theforeman/foreman_rh_cloud/commit/15f9c55836781f2946d7167718b92bd62423f80d

parthaa:
Also I dont like duplicating the logic for both rake and actions. Can you dry that part up. For example you can have this part of the logic common to both your action ad the rake task https://github.com/theforeman/foreman_rh_cloud/blob/6ba56e733a4c83d7c990049853e5f750bf323dc1/lib/tasks/rh_cloud_inventory.rake#L43-L65

lib/foreman_inventory_upload/async/generate_host_report.rb
```diff
@@ -0,0 +1,20 @@
...
+    class GenerateSingleHostReport < ::Actions::EntryAction
+      def plan(base_folder, organization_id, filter)
+        plan_self(
+          base_folder: base_folder,
+          organization_id: organization_id,
+          filter: filter,
+        )
+        input[:target] = File.join(base_folder, ForemanInventoryUpload.facts_archive_name(input[:organization_id], input[:filter]))
+        archived_report_generator = ForemanInventoryUpload::Generators::ArchivedReport.new(input[:target])
+        archived_report_generator.render(organization: input[:organization_id], filter: input[:filter])
```
adamruzicka, line 15: Assuming this is what takes the majority of time, move it to `#run` please

adamruzicka (commented):
Definitely a step in the right direction

> I didn't realize that `QueueForUploadJob` calls `UploadReportJob` which also uses ShellProcess and uploads via uploader.sh. I'll have to think about what to do about that.

Ideally yes. Firing up a shell with curl is a relatively cheap thing to do, so I wouldn't expect any major performance gains, but it would be the more correct thing to do.

jeremylenz:
> where are you accounting for this logic https://github.com/theforeman/foreman_rh_cloud/blob/develop/lib/tasks/rh_cloud_inventory.rake#L52-L66

```
next unless ForemanRhCloud.with_iop_smart_proxy?
```

Seems I misread that `unless` as an `if`. Will update

jeremylenz:
> We should use the same approach for Upload (may be in a separate PR.)

Agree, let's do the upload in a followup.

lib/foreman_inventory_upload/async/generate_host_report.rb
jeremylenz: I thought we weren't supposed to do database reads or writes in `#run`.  `ArchivedReport#render` calls a bunch of `ForemanInventoryUpload::Generators::Queries` which does database lookups.

adamruzicka: > I thought we weren't supposed to do database reads or writes in `#run`

No, why wouldn't you?

The thing is, plan is (usually, there are exceptions) done on the client, while run is performed by the workers. If you were to do everything in plan, you could just leave dynflow out of the equation altogether.

jeremylenz: I think what I remembered is that plan and finalize run in a transaction, so I thought that meant no database stuff in run (which is why I thought run was basically useless.) Good to know that I can actually do stuff there.

lib/foreman_inventory_upload/async/create_missing_insights_facets.rb
```diff
@@ -0,0 +1,28 @@
+module ForemanInventoryUpload
+  module Async
+    class CreateMissingInsightsFacets < ::Actions::EntryAction
+      def plan(organization_id)
+        plan_self(organization_id: organization_id)
+      end
+
+      def finalize
```
ShimShtein: Why finalize, and not run?
As long as the task runs in sequence to other tasks, it will be OK to run it i the run phase

lib/foreman_inventory_upload/async/single_host_report_job.rb
```diff
@@ -0,0 +1,47 @@
+module ForemanInventoryUpload
+  module Async
+    class SingleHostReportJob < HostInventoryReportJob
```
ShimShtein, line 3: If we want an inheritance model, we can just overload the `plan` method:
``` ruby
      def plan(base_folder, organization_id, host_id)
         input[:host_id] = host_id
         super(base_folder, organization_id,  "id=#{input[:host_id]}")
      end
```
That should do the trick

lib/foreman_inventory_upload/async/generate_host_report.rb
```diff
@@ -0,0 +1,23 @@
+module ForemanInventoryUpload
+  module Async
+    class GenerateHostReport < ::Actions::EntryAction
+      def plan(base_folder, organization_id, filter)
+        plan_self(
+          base_folder: base_folder,
+          organization_id: organization_id,
+          filter: filter
+        )
+        input[:target] = File.join(base_folder, ForemanInventoryUpload.facts_archive_name(input[:organization_id], input[:filter]))
```
ShimShtein, line 10: why not putting it into the `plan_self`? this way you will have it as an input for free

lib/foreman_inventory_upload/async/generate_host_report.rb
```diff
@@ -0,0 +1,23 @@
...
+      end
+
+      def run
+        archived_report_generator = ForemanInventoryUpload::Generators::ArchivedReport.new(input[:target])
+        archived_report_generator.render(organization: input[:organization_id], filter: input[:filter])
+      end
+
+      def finalize
+        output[:result] = "Generated #{input[:target]} for organization id #{input[:organization_id]}"
```
ShimShtein, line 16: You can move this also to the end of the `run`

ShimShtein (commented):
As a concept I like it. There are small changes that will make it a bit more adapted to the foreman tasks paradigm, but in general it looks good.

lib/foreman_inventory_upload/async/generate_host_report.rb
jeremylenz: I tried this, but it broke things and I'm not sure why.
```
2025-09-30T12:01:29 [D|bac|] Copying report_for_1.tar.xz to /var/lib/foreman/red_hat_inventory/uploads/
2025-09-30T12:01:29 [E|bac|] No such file or directory @ rb_file_s_rename - (/home/vagrant/foreman/report_for_1.tar.xz, /var/lib/foreman/red_hat_inventory/uploads/report_for_1.tar.xz) (Errno::ENOENT)
```
It was making a file called `report_for_.tar.xz` instead. I think just setting the input explicitly is fine imo.

lib/foreman_inventory_upload/async/create_missing_insights_facets.rb
jeremylenz: Moved back to `#run`.

jeremylenz:
@adamruzicka @parthaa @ShimShtein Updated :+1:

parthaa:
basically 'rh_cloud_inventory::generate'   was never meant to upload (it was to only generate). So my suggested patch 

```patch
$ git diff
diff --git a/lib/foreman_inventory_upload/async/host_inventory_report_job.rb b/lib/foreman_inventory_upload/async/host_inventory_report_job.rb
index c4cb9e9..d6bd97f 100644
--- a/lib/foreman_inventory_upload/async/host_inventory_report_job.rb
+++ b/lib/foreman_inventory_upload/async/host_inventory_report_job.rb
@@ -1,7 +1,7 @@
 module ForemanInventoryUpload
   module Async
     class HostInventoryReportJob < ::Actions::EntryAction
-      def plan(base_folder, organization_id, hosts_filter = "")
+      def plan(base_folder, organization_id, hosts_filter = "", upload = true)
         sequence do
           plan_action(
             GenerateHostReport,
@@ -10,12 +10,14 @@ module ForemanInventoryUpload
             hosts_filter
           )
 
-          plan_action(
-            QueueForUploadJob,
-            base_folder,
-            ForemanInventoryUpload.facts_archive_name(organization_id, hosts_filter),
-            organization_id
-          )
+          if upload
+            plan_action(
+              QueueForUploadJob,
+              base_folder,
+              ForemanInventoryUpload.facts_archive_name(organization_id, hosts_filter),
+              organization_id
+            )
+          end
 
           if ForemanRhCloud.with_iop_smart_proxy?
             plan_action(
index 212d31f..b3bb10c 100644
@@ -45,7 +45,8 @@ namespace :rh_cloud_inventory do
             ForemanInventoryUpload::Async::HostInventoryReportJob,
             base_folder,
             organization_id,
-            filter
+            filter,
+            false
           )
         end
         puts "Check the Uploading tab for report uploading status." if Setting[:subscription_connection_enabled]
```

jeremylenz:
updated & squashed

parthaa:
One more suggested patch. I hit some errors when testing the rake rh_cloud_inventory calls (not connected to changes in this PR but related to IoP in the same file.)

```patch
diff --git a/lib/foreman_inventory_upload/async/generate_report_job.rb b/lib/foreman_inventory_upload/async/generate_report_job.rb
index 86bf91f..c474415 100644
--- a/lib/foreman_inventory_upload/async/generate_report_job.rb
+++ b/lib/foreman_inventory_upload/async/generate_report_job.rb
@@ -5,7 +5,7 @@ module ForemanInventoryUpload
         "report_for_#{label}"
       end
 
-      def plan(base_folder, organization_id, disconnected, hosts_filter = nil)
+      def plan(base_folder, organization_id, disconnected = false, hosts_filter = nil)
         sequence do
           super(
             GenerateReportJob.output_label("#{organization_id}#{hosts_filter.empty? ? nil : "[#{hosts_filter.to_s.parameterize}]"}"),
index 616001d..41bc1fc 100644
@@ -9,14 +9,12 @@ namespace :rh_cloud_inventory do
       else
         organizations = [Organization.where(:id => ENV['organization_id']).first]
       end
-      disconnected = ForemanRhCloud.with_iop_smart_proxy?
       User.as_anonymous_admin do
         organizations.each do |organization|
           ForemanTasks.async_task(
             ForemanInventoryUpload::Async::GenerateReportJob,
             ForemanInventoryUpload.generated_reports_folder,
-            organization.id,
-            disconnected
+            organization.id
           )
           puts "Generated and uploaded inventory report for organization '#{organization.name}'"
         end
@@ -57,8 +55,7 @@ namespace :rh_cloud_inventory do
       base_folder = ENV['target'] || ForemanInventoryUpload.generated_reports_folder
       organization_id = ENV['organization_id']
       report_file = ForemanInventoryUpload.facts_archive_name(organization_id)
-      disconnected = ForemanRhCloud.with_iop_smart_proxy?
-      ForemanTasks.sync_task(ForemanInventoryUpload::Async::QueueForUploadJob, base_folder, report_file, organization_id, disconnected)
+      ForemanTasks.sync_task(ForemanInventoryUpload::Async::QueueForUploadJob, base_folder, report_file, organization_id)
       puts "Uploaded #{report_file}"
     end
   end
```

jeremylenz:
updated

parthaa:
I 'd like tests for some of these changes in a follow up PR. I am happy with the current state from my testing. I tested

- Inventory Upload => Generate report
- `organization_id=1 target=/var/lib/foreman/red_hat_inventory//generated_reports brake rh_cloud_inventory:report:generate`
- `organization_id=1 brake rh_cloud_inventory:report:upload`
- `organization_id=1 brake rh_cloud_inventory:report:generate_upload`

Going to approve. Please add tests in a follow up.

parthaa (approved):

chris1984:
🍏

PR #1119, opened 2025-10-16
Don't skip authentication when adding smart proxy filters

jeremylenz:
Requires https://github.com/theforeman/foreman/pull/10727

Pass `:ensure_session_expiry => true` to `add_smart_proxy_filters`. This should prevent sessions from living past their expiry time.

## Summary by Sourcery

Enforce session expiry in Smart Proxy filters to prevent authentication bypass when adding filters

Bug Fixes:
- Add ensure_session_expiry option to smart proxy filters for OrganizationsController to enforce authentication session expiry
- Add ensure_session_expiry option to smart proxy filters for RepositoriesController to enforce authentication session expiry


jeremylenz:
Closed because we went with a different approach in https://github.com/theforeman/foreman/pull/10727

PR #1120, opened 2025-10-21
Delete host from HBI on unregister

jeremylenz:

Extends Katello's `RegistrationManager#unregister_host` method to clean up the host from HBI inventory as well.

see inline comment

```
sudo -u postgres psql -d inventory_db -c "select display_name, id, org_id from hbi.hosts"
```
or `select count(*) from hbi.hosts`

Note the hosts that are present. Now unregister a host (`subscription-manager unregister`)

Check again - the host should no longer be in the hbi inventory.

Also test with deleting a host - the same should be the case.

Duplicate hosts should not get created.

### DIFF
--- a/app/controllers/concerns/foreman_rh_cloud/registration_manager_extensions.rb
+++ b/app/controllers/concerns/foreman_rh_cloud/registration_manager_extensions.rb
@@ -13,7 +13,7 @@ module ForemanRhCloud
       host.reload
 
       # Only delete from HBI in IoP mode (hosted mode uses async job for cleanup)
-      hbi_host_destroy(host) if ForemanRhCloud.with_iop_smart_proxy? && !organization_destroy && host.insights_facet.try(:uuid)
+      hbi_host_destroy(host) if ForemanRhCloud.with_iop_smart_proxy? && !organization_destroy && host.insights_facet&.uuid&.presence
       host.insights&.destroy!
       super(host, options)
     end
@@ -33,7 +33,7 @@ module ForemanRhCloud
       Rails.logger.warn(_("Attempted to destroy HBI host %s, but host does not exist in HBI") % uuid)
     rescue StandardError => e
       # TODO: Improve error handling - don't break registration if HBI delete fails
-      Rails.logger.error(_("Failed to destroy HBI host %s: %s") % [uuid, e.message])
+      Rails.logger.error(format(_("Failed to destroy HBI host %s: %s"), uuid, e.message))
     end
   end
 end
--- a/app/controllers/concerns/foreman_rh_cloud/registration_manager_extensions.rb
+++ b/app/controllers/concerns/foreman_rh_cloud/registration_manager_extensions.rb
@@ -7,26 +7,33 @@ module ForemanRhCloud
     end
 
     def unregister_host(host, options = {})
-    organization_destroy = options.fetch(:organization_destroy, false)
-    
-    hbi_host_destroy(host) if !organization_destroy && host.insights_facet.try(:uuid)
-    host.insights_facet&.destroy!
-    super(host, options)
-  end
-  
-  def hbi_host_destroy(host)
-    uuid = host.insights_facet.uuid
+      organization_destroy = options.fetch(:organization_destroy, false)
+
+      # Reload to ensure we have fresh association data
+      host.reload
+
+      # Only delete from HBI in IoP mode (hosted mode uses async job for cleanup)
+      hbi_host_destroy(host) if ForemanRhCloud.with_iop_smart_proxy? && !organization_destroy && host.insights_facet.try(:uuid)
+      host.insights&.destroy!
+      super(host, options)
+    end
+
+    def hbi_host_destroy(host)
+      uuid = host.insights_facet.uuid
       logger.debug "Unregistering host #{uuid} from HBI"
       execute_cloud_request(
-          organization: host.organization,
-          method: :delete,
-          url: ForemanInventoryUpload.host_by_id_url(uuid),
-          headers: {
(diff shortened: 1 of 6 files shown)


app/controllers/concerns/foreman_rh_cloud/registration_manager_extensions.rb
```diff
@@ -0,0 +1,39 @@
...
+    end
+
+    def unregister_host(host, options = {})
+      organization_destroy = options.fetch(:organization_destroy, false)
+
+      # Reload to ensure we have fresh association data
+      host.reload
+
+      # Only delete from HBI in IoP mode (hosted mode uses async job for cleanup)
+      hbi_host_destroy(host) if ForemanRhCloud.with_iop_smart_proxy? && !organization_destroy && host.insights_facet&.uuid&.presence
```
jeremylenz, line 16: Re: `organization_destroy` - This line is modeled on the [`candlepin_consumer_destroy` line](https://github.com/Katello/katello/blob/master/app/services/katello/registration_manager.rb#L135) in Katello's registration_manager. Katello cleans up Candlepin consumers in bulk during organization destroy, but foreman_rh_cloud currently doesn't have any corresponding cleanup attached to destroying organizations. The consequence would be that if you destroy an org, orphaned hosts will remain in HBI. (Or, if we removed the check here, possibly thousands of API calls to HBI during org destroy.) I'm thinking we should maybe add bulk HBI cleanup in a followup. Thoughts @parthaa @ShimShtein ?

app/controllers/concerns/foreman_rh_cloud/registration_manager_extensions.rb
```diff
@@ -0,0 +1,39 @@
...
+      host.insights&.destroy!
```
chris1984, line 17: Should this be `host.insights_facet&.destroy!?`

test/unit/lib/foreman_rh_cloud/registration_manager_extensions_test.rb
```diff
@@ -0,0 +1,192 @@
...
+        assert_nil InsightsFacet.find_by(id: @facet_id)
+      end
+    end
+
+    context 'integration' do
+      test 'should be properly prepended to RegistrationManager' do
+        assert_includes Katello::RegistrationManager.singleton_class.ancestors,
+          ForemanRhCloud::RegistrationManagerExtensions
+      end
+
```
chris1984, line 172: Should we add a test to not call HBI delete during org delete?
```ruby
test 'should NOT call HBI delete during organization destruction' do
ForemanRhCloud.stubs(:with_iop_smart_proxy?).returns(true)
Katello::RegistrationManager.expects(:execute_cloud_request).never
Katello::RegistrationManager.unregister_host(@host, organization_destroy: true)
end
```

chris1984 (commented):
Just a few questions, overall looks good. Learned about `singleton` from Claude and how it connects/extends Katello. Will test tomorrow and post testing results.

app/controllers/concerns/foreman_rh_cloud/registration_manager_extensions.rb
jeremylenz: `insights` is an alias for `insights_facet`; they're the same

test/unit/lib/foreman_rh_cloud/registration_manager_extensions_test.rb
jeremylenz: I mean, right now we don't call HBI delete during org delete. But I don't think that's the _correct_ behavior, and if we added a test wouldn't it be like doubling down on the incorrect behavior?

chris1984: Thinking this over for a bit, I agree, marking the comment as resolved.

chris1984 (approved):
Tested and works fine. Code looks good. Will defer to Partha or Shim for a final ack based on the open question.

Registered a system and did subman unregister:
```bash
[root@ip-10-0-168-251 ~]# sudo -u postgres psql -d inventory_db -c "select display_name, id, org_id from hbi.hosts"
                  display_name                   |                  id                  |        org_id
-------------------------------------------------+--------------------------------------+----------------------
 ip-10-0-167-31.rhos-01.prod.psi.rdu2.redhat.com | 9e9f0b04-bddc-4403-b462-01d59926f460 | Default_Organization
(1 row)

[root@ip-10-0-167-31 ~]# subscription-manager unregister
Unregistering from: ip-10-0-168-251.rhos-01.prod.psi.rdu2.redhat.com:443/rhsm
System has been unregistered.

[root@ip-10-0-168-251 ~]# sudo -u postgres psql -d inventory_db -c "select display_name, id, org_id from hbi.hosts"
 display_name | id | org_id
--------------+----+--------
(0 rows)
```
Registered a system and deleted it from the UI:
```shell
[root@ip-10-0-168-251 ~]# sudo -u postgres psql -d inventory_db -c "select display_name, id, org_id from hbi.hosts"
                  display_name                   |                  id                  |        org_id
-------------------------------------------------+--------------------------------------+----------------------
 ip-10-0-167-31.rhos-01.prod.psi.rdu2.redhat.com | 0110b331-ff5b-4ea5-b435-6e5f53f2112b | Default_Organization
(1 row)

[root@ip-10-0-168-251 ~]# sudo -u postgres psql -d inventory_db -c "select display_name, id, org_id from hbi.hosts"
 display_name | id | org_id
--------------+----+--------
(0 rows)
```

parthaa (approved):
APJ. Create a follow up for the HBI Org Destroy

PR #1127, opened 2025-11-14
Add TaskProgress components and remove obsolete UI

jeremylenz:
### Inventory Upload Page Redesign - PR 2 of 3

**User Impact:**
This PR removes the old terminal-style progress output UI and replaces it with modern PatternFly components that show task progress, status, and history in a clear, user-friendly format.

**What Changed:**
- New TaskProgress component displays current task state with progress bars and status
- New TaskHistory component shows recent task results in a clean list
- Dashboard simplified to use these new components
- Backend now queries ForemanTasks directly instead of using ProgressOutput intermediary
- All obsolete UI code, Redux state, and test files removed

**Technical Details:**
- Added API endpoints: `/api/v2/rh_cloud/inventory_upload/tasks/{current,history}`
- AccountsController returns task objects with backward-compatible status strings
- All 363 tests passing

Co-Authored-By: Claude <noreply@anthropic.com>


qcjames53 (commented):
Hey Jeremy, nice job on this PR! The new inventory upload page seems to be really solid and `/api/v2/rh_cloud/inventory_upload/tasks/current` and `/api/v2/rh_cloud/inventory_upload/tasks/history` work great when testing manually. Big fan of how the current org is auto-populated in the search field. I have a few suggested changes / questions:

I'd prefer there to be a margin between the text and the settings window. Maybe put this text in an box clearly separated from the rest of form? I'm happy to chat about this but it seems odd to put this description front and center instead of the web form?
Image

Are customers going to be okay with the output of the inventory status sync? Should this be a more permanent part of the page rather than a notification?
Image

The three obfuscation settings remain enabled after toggling back to 'Analytics' from 'Minimal'. I love how minimal flips them on and disables the field. How hard would it be for them to retain their previous values when going back to `Analytics` though? I notice they maintain state through refresh so if this is too tricky I think it's fine to nix.
Image

qcjames53:
I was trying to find examples but the best I could come up with was the about page: table on the left and lots of text on the right. Could you implement something like that here? Or if not, could the text be limited to ~80ch?

jeremylenz:
@qcjames53 Sorry, I should have been more clear that this redesign only affects the __bottom half__ of the page - the Generate & Upload Report section. We do have mocks for a redesign of the top half (which look MUCH better than this) but that's coming at a later date. :smile:

app/controllers/foreman_inventory_upload/api/tasks_controller.rb
```diff
@@ -0,0 +1,78 @@
+module ForemanInventoryUpload
+  module Api
+    class TasksController < ::Api::V2::BaseController
```
qcjames53, line 3: I think there needs to be some form of `before_action` auth check here to make sure the user has inv upload or tasks permissions?

app/controllers/foreman_inventory_upload/accounts_controller.rb
qcjames53, line 2: Same thing. I think this controller needs an authorization check outside of User.current.my_organizations.

app/controllers/foreman_inventory_upload/accounts_controller.rb
```diff
@@ -30,9 +35,89 @@ def index
...
+      return nil unless task
+
+      if task.state == 'stopped'
+        # Mimic old ProgressOutput format: "pid 12345 exit 0"
+        exit_code = task.result == 'success' ? 0 : 1
+        "pid #{Process.pid} exit #{exit_code}"
+      else
+        task.state
+      end
+    end
```
qcjames53, line 65: Are you using this method anywhere?

app/controllers/foreman_inventory_upload/accounts_controller.rb
```diff
@@ -6,15 +6,20 @@ def index
 
       accounts = Hash[
         labels.map do |id, label|
-          generate_report_status = status_for(id, ForemanInventoryUpload::Async::GenerateReportJob)
-          upload_report_status = status_for(id, ForemanInventoryUpload::Async::UploadReportDirectJob)
+          generate_task = latest_task_for(id, ForemanInventoryUpload::Async::HostInventoryReportJob)
+
+          # Check sub-action completion status
+          generated_status = sub_action_status(generate_task, 'GenerateHostReport')
+          uploaded_status = sub_action_status(generate_task, 'UploadReportDirectJob')
```
qcjames53, line 13: Claude says these are no reasons to keep these two completion statuses anymore since the new UI doesn't use them. Is that true? Can we remove these?

app/controllers/foreman_inventory_upload/reports_controller.rb
qcjames53, line 7: Like the others, I think this needs a before_action to validate auth and sanitize org_id + disconnected.

qcjames53: Claude says this action is obsolete and can be entirely removed since it isn't referenced in UI. Does anything outside of the Web UI depend on this?

qcjames53 (changes requested):
I made it through all the changed controllers. Lots of concerns about auth. For what it's worth, loading the `/foreman_rh_cloud/inventory_upload` webpage is allowed and just throws some 403s:

Image

qcjames53:
> @qcjames53 Sorry, I should have been more clear that this redesign only affects the **bottom half** of the page - the Generate & Upload Report section. We do have mocks for a redesign of the top half (which look MUCH better than this) but that's coming at a later date. 😄

Understood, thanks :)

qcjames53:
Hey @jeremylenz! I was thinking about this over the weekend and perhaps it would be better to fix all the foreman_rh_cloud controller authentication issues as a part of a separate bug fix? You didn't introduce any of these vulnerabilities so I don't think it should be on you to fix them. I'd still very much appreciate if you could add input validation though, if you  don't mind.

jeremylenz:
Added input validation for organization_id, limit, and disconnected parameters as requested. Also backported deletions from part3 (GenerateReportJob → HostInventoryReportJob transition) to reduce part3 PR size by 52%.

jeremylenz:
@qcjames53 thanks for the reviews! Updated. To address your comments about unused code I pulled in some of the changes from part3 to part2.
```
Impact on PR sizes:

  - Part 2: Increased by +324 lines (but all deletions - easy to review)
  - Part 3: Will decrease by -464 lines (52% reduction!) 🎉
```

app/controllers/foreman_inventory_upload/accounts_controller.rb
jeremylenz: This will be used in Part 3 which isn't raised yet.

app/controllers/foreman_inventory_upload/accounts_controller.rb
jeremylenz: The UI will use them once the Generated and Uploaded indicators start working, in Part 3.

qcjames53:
Awesome! Yeah, I forgot that might be the case for unused code since this was a multi-parter. I can probably put a few hours into this review tomorrow afternoon depending on how other things go; you can expect movement then :)

app/controllers/foreman_inventory_upload/api/tasks_controller.rb
```diff
@@ -0,0 +1,110 @@
...
+        tasks = ForemanTasks::Task
+                .active
+                .for_action_types(action_types)
+                .with_duration
+
+        if organization_id.present?
+          tasks = tasks.joins(:links)
+                       .where(foreman_tasks_links: {
+                         resource_type: 'Organization',
+              resource_id: organization_id,
```
qcjames53: Please fix indentation.

app/controllers/foreman_inventory_upload/api/tasks_controller.rb
```diff
@@ -0,0 +1,110 @@
...
+      # GET /foreman_inventory_upload/api/tasks/current
+      # Returns current running/active tasks for inventory operations
+      def current
+        organization_id = validate_organization_id if params[:organization_id].present?
+        return if performed?
+
+        action_types = [
+          'ForemanInventoryUpload::Async::HostInventoryReportJob',
+          'ForemanInventoryUpload::Async::UploadReportDirectJob',
+        ]
```
qcjames53: This constant happens twice in this file. I think it should be refactored.

app/controllers/foreman_inventory_upload/api/tasks_controller.rb
```diff
@@ -0,0 +1,110 @@
...
+                .for_action_types(action_types)
+                .with_duration
+                .order('started_at DESC')
+                .limit(limit)
+
+        if organization_id.present?
+          tasks = tasks.joins(:links)
+                       .where(foreman_tasks_links: {
+                         resource_type: 'Organization',
+              resource_id: organization_id,
```
qcjames53: Please fix indentation.

app/controllers/foreman_inventory_upload/api/tasks_controller.rb
```diff
@@ -0,0 +1,110 @@
...
+        {
+          id: task.id,
+          label: task.label,
+          action: task.action,
+          state: task.state,
+          result: task.result,
+          progress: task.progress,
+          started_at: task.started_at,
+          ended_at: task.ended_at,
+          duration: task.try(:duration)&.to_f,
```
qcjames53: Claude says this should be `duration: task.duration&.to_f` since `.try(:duration)` is older syntax.

app/controllers/foreman_inventory_upload/uploads_controller.rb
qcjames53, line 1: Good deletion.

config/routes.rb
qcjames53, line 1: Confirmed good.

lib/foreman_inventory_upload/async/generate_all_reports_job.rb
qcjames53, line 1: Looks good to me.

lib/foreman_inventory_upload/async/host_inventory_report_job.rb
qcjames53, line 1: This one looks great to me.

lib/foreman_inventory_upload/async/upload_report_direct_job.rb
qcjames53, line 1: This job looks good to me as well. Claude is super happy, too.

test/controllers/insights_cloud/api/cloud_request_controller_test.rb
```diff
@@ -29,8 +29,7 @@ class CloudRequestControllerTest < ActionController::TestCase
       mock_composer = mock('composer')
       ::JobInvocationComposer.expects(:for_feature).with do |feature, host_ids, params|
         feature == :rh_cloud_connector_run_playbook &&
-        host_ids.first == host1.id &&
-        host_ids.last == host2.id
+        host_ids.sort == [host1.id, host2.id].sort
```
qcjames53, line 32: This is a good add.

test/jobs/host_inventory_report_job_test.rb
```diff
@@ -10,18 +10,17 @@ class HostInventoryReportJobTest < ActiveSupport::TestCase
   let(:hosts_filter) { '' }
   let(:upload) { true }
 
+  setup do
+  end
```
qcjames53: ???

test/jobs/host_inventory_report_job_test.rb
qcjames53, line 1: The existing tests seem great but Claude says you're missing the following:
- Coverage for `report_file_path`
- Coverage for `organization`
- Coverage for `resource_locks` (this one is debatable imho)

Please advise.

test/jobs/single_host_report_job_test.rb
qcjames53, line 1: Looks good and all tests pass!

test/jobs/upload_report_direct_job_test.rb
qcjames53, line 1: These removals seem fine to me. All tests passing.

webpack/ForemanInventoryUpload/Components/AccountList/Components/ListItem/ListItem.js
qcjames53, line 1: Big file but it looks good to me.

webpack/ForemanInventoryUpload/Components/AccountList/AccountList.js
qcjames53, line 1: Claude had a few comments on this file but personally I don't agree. It looks pretty solid to me.

webpack/ForemanInventoryUpload/Components/Dashboard/__tests__/Dashboard.test.js
qcjames53, line 1: The test passes. I'm thinking this test needs some additional logic to verify the contents of the TaskProgress component match on title and task type (+ others if you want).

webpack/ForemanInventoryUpload/Components/Dashboard/Dashboard.js
```diff
@@ -1,149 +1,42 @@
...
-  restartProcess: PropTypes.func,
-  downloadReports: PropTypes.func,
-  pollingProcessID: PropTypes.number,
   account: PropTypes.shape({
-    generate_report_status: PropTypes.string,
-    upload_report_status: PropTypes.string,
-    report_file_paths: PropTypes.arrayOf(PropTypes.string),
+    id: PropTypes.number,
+    generate_task: PropTypes.shape({
+      id: PropTypes.string,
```
qcjames53, line 38: Almost missed this. This should be a number, no?

webpack/ForemanInventoryUpload/Components/Dashboard/Dashboard.js
```diff
@@ -1,149 +1,42 @@
...
-          },
-        ]}
-        toggleFullScreen={this.handleToggleFullScreen}
-        showFullScreen={showFullScreen}
-        terminalProps={this.props[activeTab]}
-      />
-    );
-  }
-}
+const Dashboard = ({ account, onTaskStart }) => (
```
qcjames53: We need error handling here for the account object returning bad values for `.generate_task` or `.id`. This also needs a unit test since a bad account obj causes a runtime error.

webpack/ForemanInventoryUpload/Components/NavContainer/NavContainer.js
qcjames53, line 1: These deletions seem fine to me!

webpack/ForemanInventoryUpload/Components/TaskHistory/TaskHistory.js
```diff
@@ -0,0 +1,131 @@
...
+        ))}
+      </DataList>
+    </div>
+  );
+};
+
+TaskHistory.propTypes = {
+  tasks: PropTypes.arrayOf(
+    PropTypes.shape({
+      id: PropTypes.string.isRequired,
```
qcjames53, line 117: Again, shouldn't this be a number?

webpack/ForemanInventoryUpload/Components/TaskHistory/TaskHistory.js
```diff
@@ -0,0 +1,131 @@
...
+          </DataListItem>
+        ))}
+      </DataList>
+    </div>
+  );
+};
+
+TaskHistory.propTypes = {
+  tasks: PropTypes.arrayOf(
+    PropTypes.shape({
```
qcjames53, line 116: Claude had the following to say. What are your thoughts on this? I don't know enough about `.shape` to advise:
```
The backend returns additional properties in task_json that aren't in PropTypes:
  - label (string)
  - action (string)
  - state (string)
  - ended_at (string)
  - progress (number)
  - humanized (object)

  If these aren't used, that's fine, but the shape should reflect what's actually received for documentation purposes.
```

webpack/ForemanInventoryUpload/Components/TaskProgress/TaskProgress.js
```diff
@@ -0,0 +1,359 @@
...
+          )}
+        </Flex>
+      </CardBody>
+    </Card>
+  );
+};
+
+TaskProgress.propTypes = {
+  task: PropTypes.shape({
+    id: PropTypes.string.isRequired,
```
qcjames53, line 315: Again, this should be a number.

webpack/ForemanInventoryUpload/Components/TaskProgress/TaskProgress.js
```diff
@@ -0,0 +1,359 @@
...
+      dispatch(
+        addToast({
+          sticky: true,
+          type: 'error',
+          message: error.message,
+        })
+      );
+    } finally {
+      setIsLoading(false);
+    }
```
qcjames53: It looks like you're using both `addToast` and `tfm.toastNotifications`. Can `addToast` be used for both of these please?

webpack/ForemanInventoryUpload/Components/TaskProgress/TaskProgress.js
```diff
@@ -0,0 +1,359 @@
...
+    result: PropTypes.string,
+    progress: PropTypes.number,
+    started_at: PropTypes.string,
+    ended_at: PropTypes.string,
+    duration: PropTypes.number,
+    report_file_path: PropTypes.string,
+  }),
+  title: PropTypes.string,
+  emptyMessage: PropTypes.string,
+  organizationId: PropTypes.oneOfType([PropTypes.string, PropTypes.number]),
```
qcjames53: This is weird. Why both?

webpack/ForemanInventoryUpload/Components/TaskProgress/TaskProgress.js
```diff
@@ -0,0 +1,359 @@
...
+              <FlexItem>
+                <Button
+                  ouiaId="generate-report-button-task-view"
+                  variant="secondary"
+                  onClick={() => handleGenerateReport(true)}
+                  isDisabled={areButtonsDisabled}
+                >
+                  {__('Generate report')}
+                </Button>
+              </FlexItem>
```
qcjames53: Claude caught that this was almost completely duplicated code compared to lines 130-164. Can you combine this with empty state and refactor please?

webpack/ForemanInventoryUpload/Components/TaskProgress/TaskProgress.js
```diff
@@ -0,0 +1,359 @@
...
+  const formatDuration = seconds => {
+    if (
+      seconds === null ||
+      seconds === undefined ||
+      typeof seconds !== 'number' ||
+      Number.isNaN(seconds) ||
+      seconds < 0
+    ) {
+      return '';
+    }
```
qcjames53, line 215: Claude wants you to refactor to `if (!seconds || seconds < 0) return '';`. Please double check all the types work out here but it sounds reasonable to me.

webpack/ForemanInventoryUpload/Components/TaskProgress/TaskProgress.js
```diff
@@ -0,0 +1,359 @@
...
+      {title && <CardTitle>{title}</CardTitle>}
+      <CardBody>
+        <Progress
+          value={progressValue}
+          title={displayStateLabel}
+          variant={displayVariant}
+          measureLocation="outside"
+          aria-label="task-progress"
+        />
+        <DescriptionList isHorizontal className="task-progress-details">
```
qcjames53: This should have an ouiaId.

webpack/ForemanInventoryUpload/Components/TaskProgress/TaskProgress.js
qcjames53, line 1: This was quite the file to wrap my brain around. When you implement these fixes, please pass it through Claude to get its opinion. There are a few edge cases that still need ironed out here.

qcjames53 (changes requested):
Damn this PR got hands

webpack/ForemanInventoryUpload/Components/Dashboard/Dashboard.js
jeremylenz: I think ForemanTask::Task id's are string uuids.

webpack/ForemanInventoryUpload/Components/TaskHistory/TaskHistory.js
jeremylenz: Oh, `.shape` is just what you're supposed to use in PropTypes instead of `PropTypes.object`. It allows you to specify the keys of the object and what type each value is.

webpack/ForemanInventoryUpload/Components/TaskProgress/TaskProgress.js
jeremylenz: Ugh I bet the test uses number and real life uses a string, I'll check

test/jobs/host_inventory_report_job_test.rb
qcjames53: Sorry for the bad comment. Do we need to keep this?

webpack/ForemanInventoryUpload/Components/Dashboard/Dashboard.js
qcjames53: You're right!

jeremylenz:
@qcjames53 Updated :+1:

app/controllers/foreman_inventory_upload/api/tasks_controller.rb
qcjames53: This isn't addressed yet :(

Can you require `:view_foreman_rh_cloud` for this controller?

app/controllers/foreman_inventory_upload/reports_controller.rb
qcjames53: Can you address this please?

webpack/ForemanInventoryUpload/Components/TaskProgress/TaskProgress.js
qcjames53: This still needs to be addressed.

webpack/ForemanInventoryUpload/Components/TaskProgress/TaskProgress.js
qcjames53: This still needs to be addressed.

webpack/ForemanInventoryUpload/Components/TaskProgress/TaskProgress.js
qcjames53: This still needs to be addressed? What did you find out?

qcjames53 (changes requested):
There are a few outstanding comments that need to be addressed. The other requested changes seem good to me; all tests passing locally.

qcjames53:
@jeremylenz I'm unable to un-resolve the five comments above but the issues are still outstanding.

app/controllers/foreman_inventory_upload/reports_controller.rb
jeremylenz: I'm adding authorization following Foreman's pattern to all the controllers. (Note that you won't see a `before_action` in the code here, since Foreman's Api::V2::BaseController has the `before_action :authorize`.)

webpack/ForemanInventoryUpload/Components/TaskProgress/TaskProgress.js
jeremylenz: We intentionally kept the more defensive checks because:
  1. `!seconds` would be true for 0, but 0 is a valid duration (0 seconds)
  2. We want to explicitly handle null, undefined, non-numbers, and `NaN` separately from the actual value of 0

  The current implementation is more defensive and handles edge cases better. With the simplified version `!seconds`, we would incorrectly return '' for seconds = 0, when we should
  actually return '0s'.

webpack/ForemanInventoryUpload/Components/TaskProgress/TaskProgress.js
jeremylenz: Seems it works fine as a Number.

webpack/ForemanInventoryUpload/Components/TaskProgress/TaskProgress.js
jeremylenz: ouiaId added 👍

jeremylenz:
@qcjames53 Updated. I _think_ I got 'em all this time. (either addressed or replied to.) Let me know if it works for you

qcjames53:
Thanks! I'll take a look tomorrow!

app/controllers/foreman_inventory_upload/reports_controller.rb
qcjames53: Awesome, thank you :)

webpack/ForemanInventoryUpload/Components/TaskProgress/TaskProgress.js
qcjames53: Gotcha. Thank you for the rationale :)

qcjames53 (approved):
I gave this the final once-over and the PR looks good to me! The new ForemanTasks additions, `/api/v2/rh_cloud/inventory_upload/tasks/current`, and `/api/v2/rh_cloud/inventory_upload/tasks/history` all work well and seem robust. The new components in the UI for the inventory upload dashboard looks pretty snazzy to me as well. As I understand it we're certain that page is getting cleaned up before we ship a major version. 

I'm sure I missed MULTIPLE issues but we can always go back and fix bugs later if need be. We're looking solid, though. Thank you for all the work you put into this, Jeremy.

chris1984:
Testing wise it works great, tested with `subscription_connection_enabled` set to false and `generate and upload` button is disabled. I see that Generating and Uploading tabs are no longer doing anything, are we able to remove that out of this pr or the next one? I could see users confused why nothing is showing up there. Starting to look at the code

jeremylenz:
> I see that Generating and Uploading tabs are no longer doing anything, are we able to remove that out of this pr or the next one?

Are those still there? If so they will definitely be removed in part 3, which is the followup PR that I am opening right after this is merged. Or if you prefer I can just combine them now. but I don't want to ruin @qcjames53's life

chris1984:
> > I see that Generating and Uploading tabs are no longer doing anything, are we able to remove that out of this pr or the next one?
> 
> Are those still there? If so they will definitely be removed in part 3, which is the followup PR that I am opening right after this is merged. Or if you prefer I can just combine them now. but I don't want to ruin @qcjames53's life

Let's push it to part 3 that you will open up next, I don't want to throw anything new into the mix since it's already beefy

chris1984 (approved):
Looked through all the code and most of the stuff that jumped out is what @qcjames53 commented on. I am sure I missed a few things but with this pr having hands and being beefy we can fix things in part 3 like the tabs I mentioned. Tested and works well and looks a hell of a lot better than the old page. ACK

ColeHiggins2 (commented):
late but looks good to me !

jeremylenz:
Part 3: https://github.com/theforeman/foreman_rh_cloud/pull/1130

PR #1130, opened 2025-12-08
Polish inventory upload UI: update status field names and improve testing

jeremylenz:

**User Impact:**
- **Status labels updated for clarity**: The status indicators in the organization accordion view now show "Generated" and "Uploaded" instead of "Generating" and "Uploading" to better reflect their meaning as historical status indicators rather than ongoing actions.

**Developer/Code Quality Improvements:**
- Updated backend field names for consistency: `generated_status` and `uploaded_status` (previously `generate_report_status` and `upload_report_status`)
- Enhanced test coverage with updated snapshots across all AccountList components
- Fixed ESLint configuration for proper linting
- Improved code consistency between frontend and backend

This PR builds on #1127 and focuses on polishing the UI:

1. **Field naming consistency**: The backend was already returning `generated_status` and `uploaded_status`, but the frontend was using the old field names. This PR updates all frontend components to use the correct field names that match the backend.

2. **Backwards compatibility**: This is a new UI (from #1127), so there are no backwards compatibility concerns.

3. **Test coverage**: All component snapshots have been updated to reflect the new field names, ensuring tests accurately validate the component behavior.

1. Navigate to Configure > Inventory Upload in Foreman
2. Verify the organization accordion view displays properly
3. Check that status indicators show "Generated" and "Uploaded" labels
4. Run tests: `npm test` (JavaScript) and `bundle exec rake test:foreman_rh_cloud` (Ruby)
5. Run linting: `npm run lint`

All tests should pass and the UI should render correctly in both regular and IoP modes.

## Summary by Sourcery

Align inventory upload backend job behavior and frontend UI with new status semantics while improving robustness and IoP handling.

New Features:
- Show separate "Generated" and "Uploaded" status labels in the inventory account list using the new generated_status and uploaded_status fields.
- Display a tooltip explaining why upload is disabled when subscription connection is turned off in the inventory upload header.
- Treat IoP mode as always allowed to generate/upload regardless of subscription connection setting.

Bug Fixes:
- Avoid crashes when duration values in task history are null, invalid, or negative by treating them as not available.
- Ensure the inventory upload generate/upload job uses the correct organization as its Dynflow action subject and exposes a predictable output label.
- Prevent confusing behavior when SSL certificate or key files are missing by validating readability and raising clear errors instead of silently failing.
- Allow re-running uploads by overwriting existing files in the done folder with a logged warning instead of failing on existing destinations.
- Correctly handle content disconnection logic so IoP mode bypasses subscription connection checks both in the backend and UI button state.

Enhancements:
- Add a resource lock to the inventory upload Dynflow action to better coordinate concurrent operations.
- Increase test coverage for the upload job, inventory upload header, and IoP/content-connection combinations, including new error scenarios for certificate handling.
- Rename and standardize account status field usage across fixtures, components, and controller tests to match backend naming and semantics.
- Tidy up remediation helper comments and disable/enable spellcheck linting in a more focused way.

Tests:
- Extend Ruby job tests to cover action_subject behavior, new output labels, IoP/content-disconnected logic, certificate error handling, and file move behavior.
- Expand JavaScript tests for the TabHeader component to assert button enabled/disabled states and labels across subscription connection and IoP combinations.
- Update React component fixtures and snapshots for account list and status components to reflect the new generated/uploaded fields and labels.


lib/foreman_inventory_upload/async/upload_report_direct_job.rb
```diff
@@ -102,7 +109,10 @@ def upload_file(cer_path)
       def move_to_done_folder
         FileUtils.mkdir_p(ForemanInventoryUpload.done_folder)
         done_file = ForemanInventoryUpload.done_file_path(File.basename(filename))
-        FileUtils.mv(filename, done_file)
+        if File.exist?(done_file)
+          logger.warn("Destination file #{done_file} already exists. Overwriting with new report.")
```
chris1984, line 113: Thanks for adding this, I never got around to adding wording that we are just going to overwrite your report, just doing it blindly was a bad practice.

test/jobs/upload_report_direct_job_test.rb
```diff
@@ -28,43 +28,38 @@ class UploadReportDirectJobTest < ActiveSupport::TestCase
       },
     }
     Organization.any_instance.stubs(:owner_details).returns(@cert_data)
-
-    # Clear task output
-    TaskOutputLine.delete_all
```
chris1984, line 33: Since we are no longer showing the shell process and we are removing it from the tests, should we do a migration to remove those two tables?

`task_output_lines`
`task_output_statuses`

webpack/ForemanInventoryUpload/Components/TaskHistory/TaskHistory.js
```diff
@@ -49,7 +49,15 @@ const TaskHistory = ({ tasks, title }) => {
   };
 
   const formatDuration = seconds => {
```
chris1984, line 51: Could it be made cleaner? I asked Gemini and it came with this:
```javascript
const formatDuration = seconds => {
  // Check if it's NOT a finite number OR if it's negative
  if (!Number.isFinite(seconds) || seconds < 0) {
    return __('N/A');
  }
  
  // ... rest of function
};
```

Why this works:

Number.isFinite(null) returns false

Number.isFinite(undefined) returns false

Number.isFinite('123') returns false (it does not coerce strings)

Number.isFinite(NaN) returns false

chris1984 (commented):
Left a few comments, will get it tested too

lib/foreman_inventory_upload/async/upload_report_direct_job.rb
qcjames53, line 1: All the actual changes in this file look good to me. I was able to verify the new log message works great.

jeremylenz:
@chris1984 good suggestions, updated :+1:

test/jobs/upload_report_direct_job_test.rb
qcjames53, line 1: I feel like with so many changes that I should have found something to complain about here. Both Claude and myself think this is well written and that your coverage is good. I mean, you even test for the log message when overwriting the old org report :)

All tests pass.

db/migrate/20251209163012_drop_task_output_tables.foreman_rh_cloud.rb
qcjames53, line 1: Thanks for pointing out these tables, Chris. I had Claude think about these drops for a while and this should be a really clean migration for everyone. There's a reference to ProgressOutput on `app/controllers/foreman_inventory_upload/accounts_controller.rb:59` that we could consider rewording since it won't really make sense in the future.

qcjames53 (changes requested):
I'm about halfway through my review but noticed that these two files still contain 'generate_report_status' + 'upload_report_status'. Can you update them and rebuild the snaps please?
- `webpack/ForemanInventoryUpload/Components/AccountList/AccountListActions.js`
- `webpack/ForemanInventoryUpload/Components/AccountList/Components/ListItem/ListItem.fixtures.js`

webpack/ForemanInventoryUpload/Components/TabHeader/__tests__/TabHeader.test.js
qcjames53, line 1: All tests pass. You have coverage for all combinations of `useIopConfig` and `subscriptionConnectionEnabled`.

qcjames53 (changes requested):
Everything else looked great to me, nice job. Just the comment above needs addressed.

I want to confirm the inventory upload UI is still going to have the paragraph in the header reworked in the near future? The margins between components are sadly still pretty unappealing.

jeremylenz:
> I want to confirm the inventory upload UI is still going to have the paragraph in the header reworked in the near future? The margins between components are sadly still pretty unappealing.

Yep, the top half of the page redesign has mockups, we just need to plan that work in somewhere.

app/controllers/foreman_inventory_upload/accounts_controller.rb
```diff
@@ -56,7 +56,7 @@ def task_status_string(task)
       return nil unless task
 
       if task.state == 'stopped'
-        # Mimic old ProgressOutput format: "pid 12345 exit 0"
+        # Format status as "pid 12345 exit 0" for UI compatibility
```
jeremylenz: >  There's a reference to ProgressOutput on app/controllers/foreman_inventory_upload/accounts_controller.rb:59 that we could consider rewording since it won't really make sense in the future.

Actually, I want to go even further here. The old "pid xxxxx exit 0" format no longer makes any sense. So I'm going to simplify this.

jeremylenz: Removed `task_status_string` method entirely. We don't need it anymore.

jeremylenz:
@qcjames53 updated :+1:

qcjames53 (approved):
This looks great to me, Jeremy. Thanks for the changes!

chris1984 (commented):
Did another review and the code looks good, will start to test and ack once that is done. Thanks for addressing my comments :)

chris1984 (approved):
Tested and works great, nice to see the generating and uploading tab responding to the new changes. ACK

jeremylenz:
Thanks!

Before merging I'll wait for final automation ack from @ColeHiggins2.

ColeHiggins2 (approved):
Looks good to me!

PR #1131, opened 2025-12-09
Fix IoP host UUID data consistency issues

jeremylenz:
# Fix IoP host UUID data consistency issues

  ## User Impact

  In IoP (Insights on Premise) mode, hosts that were previously registered to Red Hat Hybrid Cloud Console could display incorrect/stale UUIDs throughout the application. This caused:

  - **Data inconsistencies** when viewing host details and recommendations
  - **Incorrect identification** during host unregistration
  - **Mismatched records** between Foreman and the local Insights inventory

  This fix ensures IoP hosts always use their subscription manager UUID (the correct identifier for IoP environments) instead of stale cloud-assigned UUIDs. Hosts are automatically
  synchronized when they upload package profiles.

Note: This does _not_ remove the UUID field from insights facets, or allow nils there. The insights UUID is still kept in sync just as before, as a "just in case" kind of thing. But I've tried to move the actual _decision points_ to look at subscription facet. This gets us closer to a single source of truth.

  ## Technical Changes

  - Added `insights_uuid` method that returns the appropriate UUID based on operating mode:
    - **IoP mode**: Always uses subscription facet UUID (current/correct)
    - **Cloud mode**: Uses insights facet UUID (cloud-assigned)
  - Added `ensure_iop_insights_uuid` method to automatically sync mismatched UUIDs
  - Updated all call sites to use the new `insights_uuid` method
  - Comprehensive test coverage with 14 new test cases

  ## Testing steps

  1. Get an IoP host with recommendations
  2. Note that recommendations display BOTH on the Insights > Recommendations page, and on the host details > Recommendations tab
  3. In rails console, run
```
myhost.subscription_facet.uuid
myhost.insights_facet.uuid
```
Note that they are the same.
  5. In rails console, run
```
myhost.insights_facet.update(uuid: "foo")
```
Now, the uuids are different. This simulates a stale uuid from a host previously registered to Hosted insights.

6. Refresh

#### Before checking out the PR:
Note that recommendations still display in Insights > Recommendations, and show the affected host. BUT they do NOT show up on the host details > Recommendations tab.
You may notice a failed API request to "foo" on the browser Network tab

#### After checking out the PR:
Recommendations will display on BOTH pages
No failed API requests

### DIFF
--- a/app/controllers/concerns/foreman_rh_cloud/registration_manager_extensions.rb
+++ b/app/controllers/concerns/foreman_rh_cloud/registration_manager_extensions.rb
@@ -19,7 +19,7 @@ module ForemanRhCloud
     end
 
     def hbi_host_destroy(host)
-      uuid = host.insights_facet.uuid
+      uuid = host.insights_uuid
       logger.debug "Unregistering host #{uuid} from HBI"
       execute_cloud_request(
         organization: host.organization,
--- a/app/controllers/concerns/foreman_rh_cloud/registration_manager_extensions.rb
+++ b/app/controllers/concerns/foreman_rh_cloud/registration_manager_extensions.rb
@@ -13,7 +13,7 @@ module ForemanRhCloud
       host.reload
 
       # Only delete from HBI in IoP mode (hosted mode uses async job for cleanup)
-      hbi_host_destroy(host) if ForemanRhCloud.with_iop_smart_proxy? && !organization_destroy && host.insights_facet&.uuid&.presence
+      hbi_host_destroy(host) if ForemanRhCloud.with_iop_smart_proxy? && !organization_destroy && host.insights_uuid.presence
       host.insights&.destroy!
       super(host, options)
     end
--- a/lib/insights_cloud/async/connector_playbook_execution_reporter_task.rb
+++ b/lib/insights_cloud/async/connector_playbook_execution_reporter_task.rb
@@ -128,8 +128,8 @@ module InsightsCloud
         hosts_state = Hash[job_invocation.targeting.hosts.map do |host|
           next unless host.insights&.uuid
           [
-            host.insights.uuid,
-            task_status(job_invocation.sub_task_for_host(host), host.insights.uuid),
+            host.insights_uuid,
+            task_status(job_invocation.sub_task_for_host(host), host.insights_uuid),
           ]
         end.compact]
 
--- a/lib/insights_cloud/async/connector_playbook_execution_reporter_task.rb
+++ b/lib/insights_cloud/async/connector_playbook_execution_reporter_task.rb
@@ -126,7 +126,7 @@ module InsightsCloud
 
       def invocation_status
         hosts_state = Hash[job_invocation.targeting.hosts.map do |host|
-          next unless host.insights&.uuid
+          next unless host.insights_uuid
           [
             host.insights_uuid,
             task_status(job_invocation.sub_task_for_host(host), host.insights_uuid),
--- a/app/controllers/concerns/insights_cloud/package_profile_upload_extensions.rb
+++ b/app/controllers/concerns/insights_cloud/package_profile_upload_extensions.rb
@@ -21,6 +21,8 @@ module InsightsCloud
         @host.id
       )
 
+      @host.ensure_iop_insights_uuid
+
       # in IoP case, the hosts are identified by the sub-man ID, and we can assume they already
       # exist in the local inventory. This will also handle facet creation for new hosts.
       return if @host.insights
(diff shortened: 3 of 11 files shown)


app/controllers/concerns/foreman_rh_cloud/registration_manager_extensions.rb
nofaralfasi: The guard checks `insights_facet&.uuid&.presence`, but `hbi_host_destroy` uses `insights_uuid` (which returns `subscription_facet.uuid` in IoP mode).
Question: Should the guard also use `insights_uuid`?

lib/insights_cloud/async/connector_playbook_execution_reporter_task.rb
```diff
@@ -126,10 +126,10 @@ def report_url
 
       def invocation_status
         hosts_state = Hash[job_invocation.targeting.hosts.map do |host|
-          next unless host.insights&.uuid
+          next unless host.insights_uuid
```
nofaralfasi, line 129: In non-IoP mode, this is functionally equivalent. In IoP mode, a host without `subscription_facet` would now be skipped (returns nil), whereas before it would use `insights_facet.uuid`.
Is this the intended behavior? If a host has `insights_facet` but no `subscription_facet` in IoP mode, should it be excluded from playbook reporting?

nofaralfasi (commented):
LGTM. tested and works as expected.
One potential bug to investigate: the guard condition in `registration_manager_extensions.rb:16` may cause hosts with `subscription_facet.uuid` but no `insights_facet.uuid` to skip HBI unregistration in IoP mode.

nofaralfasi:
Question about [InsightsHit#host_uuid](https://github.com/theforeman/foreman_rh_cloud/blob/develop/app/models/insights_hit.rb#L27):
I noticed that `app/models/insights_hit.rb` still accesses `insights_facet.uuid` directly.
Was this intentionally left unchanged? I'm trying to understand if there's a reason this doesn't need updating to use insights_uuid, or if it was just missed.
I see that in practice the UUIDs should stay in sync because:
- `ensure_iop_insights_uuid` syncs them on package profile upload
- `HitsUploader#update_facets` updates the facet UUID when hits are uploaded

So the current code probably works fine in most cases. But for consistency with the pattern established in this PR (always reading from the authoritative source via `insights_uuid`), should we update this too?

app/controllers/concerns/foreman_rh_cloud/registration_manager_extensions.rb
jeremylenz: Good point! Updating.

lib/insights_cloud/async/connector_playbook_execution_reporter_task.rb
jeremylenz: Actually I think this is not an issue, because this task is only relevant to the Cloud Connector which isn't a thing in non-IoP.

jeremylenz:
> I noticed that `app/models/insights_hit.rb` still accesses `insights_facet.uuid` directly. Was this intentionally left unchanged? I'm trying to understand if there's a reason this doesn't need updating to use insights_uuid, or if it was just missed. I see that in practice the UUIDs should stay in sync because:
> 
> * `ensure_iop_insights_uuid` syncs them on package profile upload
> * `HitsUploader#update_facets` updates the facet UUID when hits are uploaded
> 
> So the current code probably works fine in most cases. But for consistency with the pattern established in this PR (always reading from the authoritative source via `insights_uuid`), should we update this too?

I'm on the fence here. On one hand, I see what you're saying about consistency and source of truth. But on the other hand,

- InsightsHit is a hosted-only model, so these codepaths should never be touched in IoP
- if something is out of sync and `ForemanRhCloud.with_iop_smart_proxy?` is incorrect, we could incorrectly try finding a non-IoP host via its subscription-manager id.
- in the course of building IoP, I have tried to adhere to the principle "Don't change the non-IoP code / behavior / experience at all, if you can help it." But hey, I've changed it in some other places that only affect hosted. Like I said.. on the fence.

Let me know your thoughts.

nofaralfasi:
> I'm on the fence here. On one hand, I see what you're saying about consistency and source of truth. But on the other hand,
> 
> * InsightsHit is a hosted-only model, so these codepaths should never be touched in IoP
> * if something is out of sync and `ForemanRhCloud.with_iop_smart_proxy?` is incorrect, we could incorrectly try finding a non-IoP host via its subscription-manager id.
> * in the course of building IoP, I have tried to adhere to the principle "Don't change the non-IoP code / behavior / experience at all, if you can help it." But hey, I've changed it in some other places that only affect hosted. Like I said.. on the fence.
> 
> Let me know your thoughts.

That makes sense - Since InsightsHit is hosted-only, and if these codepaths aren't touched in IoP, then adding the IoP branching logic there would violate the "minimal change to non-IoP code" principle and could introduce unnecessary risk.

No changes needed here - thanks for the context!

nofaralfasi (approved):

PR #1132, opened 2025-12-11
SAT-38427 - Implement CSV exports on the hosts overview

lfu:
Add missing Manage columns to pagelet for CSV export.

Requires https://github.com/theforeman/foreman/pull/10793

## Summary by Sourcery

Add CSV export support for RH Cloud host overview columns related to insights recommendations and CVE counts.

New Features:
- Enable CSV export for the insights recommendations count column on the RH Cloud hosts overview.
- Expose total CVEs as a new column on the RH Cloud hosts overview with CSV export support.

### DIFF
--- a/lib/foreman_rh_cloud/plugin.rb
+++ b/lib/foreman_rh_cloud/plugin.rb
@@ -142,7 +142,7 @@ module ForemanRhCloud
         extend_page 'hosts/_list' do |context|
           context.with_profile :cloud, _('RH Cloud'), default: true do
             add_pagelet :hosts_table_column_header, key: :insights_recommendations_count, label: _('Recommendations'), sortable: true, width: '12%', class: 'hidden-xs ellipsis', priority: 100,
-                        export_data: CsvExporter::ExportDefinition.new(:insights_recommendations_count, callback: ->(host) { host&.insights_hits&.count })
+                        export_data: CsvExporter::ExportDefinition.new(:insights_recommendations_count, label: _('Recommendations'), callback: ->(host) { ForemanRhCloud.with_iop_smart_proxy? ? 'N/A' : host&.insights_hits&.count })
             add_pagelet :hosts_table_column_content, key: :insights_recommendations_count, callback: ->(host) { hits_counts_cell(host) }, class: 'hidden-xs ellipsis text-center', priority: 100
           end
         end


lib/foreman_rh_cloud/plugin.rb
```diff
@@ -142,8 +142,11 @@ def self.register
         extend_page 'hosts/_list' do |context|
           context.with_profile :cloud, _('RH Cloud'), default: true do
             add_pagelet :hosts_table_column_header, key: :insights_recommendations_count, label: _('Recommendations'), sortable: true, width: '12%', class: 'hidden-xs ellipsis', priority: 100,
-                        export_data: CsvExporter::ExportDefinition.new(:insights_recommendations_count, callback: ->(host) { host&.insights_hits&.count })
+                        export_data: CsvExporter::ExportDefinition.new(:insights_recommendations_count, label: _('Recommendations'), callback: ->(host) { host&.insights_hits&.count })
```
jeremylenz: This won't work with IoP. Insights hits are not stored in our database. Rather, we make an API call to IoP services from the frontend: https://github.com/theforeman/foreman_rh_cloud/blob/083b91d62b519c7921a314fb0bcbfaefb92607a8/webpack/ForemanColumnExtensions/index.js#L31-L32

Not sure what that means for CSV export. Perhaps you could make the same call from the backend. Or if it's too much, we could disable this column for export, but I hope we don't have to.

lib/foreman_rh_cloud/plugin.rb
```diff
@@ -142,8 +142,11 @@ def self.register
...
             add_pagelet :hosts_table_column_content, key: :insights_recommendations_count, callback: ->(host) { hits_counts_cell(host) }, class: 'hidden-xs ellipsis text-center', priority: 100
+            add_pagelet :hosts_table_column_header, key: :cves_count, label: _('Total CVEs'), width: '12%', class: 'hidden-xs ellipsis',
+                        export_data: CsvExporter::ExportDefinition.new(:cves_count, label: _('Total CVEs'), callback: ->(host) { host&.insights_attributes&.dig('cves_count') })
```
jeremylenz: `cves_count` is not a valid insights attribute. Just like recommendations, for CVEs we make an API call: https://github.com/jeremylenz/foreman_rh_cloud/blob/083b91d62b519c7921a314fb0bcbfaefb92607a8/webpack/InsightsVulnerabilityHostIndexExtensions/CVECountCell.js#L21-L22

jeremylenz: Note that this column is only "relevant" for IoP; it is not available with hosted.

jeremylenz (changes requested):
Thanks for opening the PR @lfu!

lfu:
CSV exports are synchronous batch operations in one request. Seems not a fit solution for IoP export which would make hundreds/thousands of sequential API calls during CSV generation depending on the number of hosts.

How about we keep the CSV export for hosted only until there is a better solution for IoP export?

jeremylenz:
> How about we keep the CSV export for hosted only until there is a better solution for IoP export?

That sounds good to me. In that case, we will not be able to export the CVE column.

lfu:
CVE data comes from external services rather than the local database.
Current CSV export works with data from local database in a synchronous call.

jeremylenz (approved):
Thanks @lfu!

jeremylenz:
please rebase

PR #1138, opened 2026-01-08
Evaluate IoP status at runtime, not application start, and register Ping extension always

jeremylenz:

1. Evaluate `ForemanRhCloud.with_iop_smart_proxy?` in a lambda for ForemanContext app metadata registration. This ensures the UI responds immediately when switching between IoP and non-IoP.
2. Register the foreman_rh_cloud Ping extension regardless of IoP. This ensures that the advisor and vulnerability nodes will immediately appear when IoP is turned on, and disappear when turned off.

AI slop below, but I'm gonna hand-write the manual testing steps.

**IoP Metadata Runtime Evaluation:**
- Wrapped `with_iop_smart_proxy?` check in lambda for plugin metadata registration
- Fixes stale metadata when IoP smart proxy is created/destroyed without server restart
- Depends on Foreman core PR: https://github.com/theforeman/foreman/pull/10814

**Simplified URL Caching:**
- Removed complex IoP state tracking from `base_url`, `cert_base_url`, and `legacy_insights_url`
- These methods now rely on `env_or_on_premise_url` which is already IoP-aware
- Cleaner, more maintainable code that's easier to understand

**Ping Extension Improvements:**
- Removed conditional ping extension registration (always register, return `{}` when no IoP)
- Added nil safety to `iop_smart_proxy_url` with safe navigation operator
- Early return in `ping` method when no IoP proxy exists

- **Performance**: The lambda evaluation relies on ActiveRecord query cache, so performance is unchanged despite dynamic evaluation
- **Backward compatibility**: Always registering ping/status extensions ensures consistent behavior
- **Foreman core dependency**: This PR depends on the runtime Proc evaluation feature in Foreman core (PR #10814)
- **User impact**: When an IoP smart proxy is created or destroyed, the plugin metadata and URLs will now correctly reflect the current state without requiring a server restart

**Unit Tests:**
- 4 tests for plugin metadata lambda evaluation
- 10 tests for URL method dynamic behavior with IoP mode changes
- All 336 foreman_rh_cloud tests passing

**Manual Testing:**
Check out this PR and the Foreman PR --> https://github.com/theforeman/foreman/pull/10814

1. Start Foreman with IoP smart proxy
2. Infrastructure > About: Backend system status should show "advisor" and "vulnerability"
3. Main nav: Insights should have 'Recommendations' and 'Vulnerability' sub-items.
4. Both pages should load Scalprum components (you can tell because you see the "loading..." text)
5. Host details page should have Recommendations and Vulnerability tabs
6. Recommendations tab should load the Scalprum component

8. Now, go to Infrastructure > Smart Proxies and DELETE the IoP smart proxy
9. The following changes should be IMMEDIATE:
10. Infrastructure > About: Backend system status will NOT have "advisor" and "vulnerability"
11. Main nav: Insights should have 'Recommendations' and 'Inventory Upload'
12. Host details page should have only 'Recommendations', NOT 'Vulnerabilities'
13. Should NOT load Scalprum components anywhere

Now, Infrastructure > Smart Proxies > Create a new smart proxy with url `https://localhost:24443 `
Everything should immediately go back to how it was in steps 1-6.

### DIFF
--- a/lib/foreman_rh_cloud/plugin.rb
+++ b/lib/foreman_rh_cloud/plugin.rb
@@ -127,10 +127,8 @@ module ForemanRhCloud
 
         register_custom_status InventorySync::InventoryStatus
         register_custom_status InsightsClientReportStatus
-        if ForemanRhCloud.with_iop_smart_proxy?
-          register_ping_extension { ForemanRhCloud::Ping.ping }
-          register_status_extension { ForemanRhCloud::Ping.status }
-        end
+        register_ping_extension { ForemanRhCloud::Ping.ping }
+        register_status_extension { ForemanRhCloud::Ping.status }
 
         describe_host do
           overview_buttons_provider :insights_host_overview_buttons
@@ -153,7 +151,7 @@ module ForemanRhCloud
         end
 
         ::Foreman::Plugin.app_metadata_registry.register(:foreman_rh_cloud, {
-          iop: ForemanRhCloud.with_iop_smart_proxy?,
+          iop: -> { ForemanRhCloud.with_iop_smart_proxy? },
         })
 
         extend_template_helpers ForemanRhCloud::TemplateRendererHelper
--- a/app/models/foreman_rh_cloud/ping.rb
+++ b/app/models/foreman_rh_cloud/ping.rb
@@ -7,7 +7,7 @@ module ForemanRhCloud
       include ForemanRhCloud::CertAuth
 
       def iop_smart_proxy_url
-        @iop_smart_proxy_url ||= ForemanRhCloud.iop_smart_proxy&.url
+        ForemanRhCloud.iop_smart_proxy&.url
       end
 
       def service_urls
--- a/app/models/foreman_rh_cloud/ping.rb
+++ b/app/models/foreman_rh_cloud/ping.rb
@@ -7,7 +7,7 @@ module ForemanRhCloud
       include ForemanRhCloud::CertAuth
 
       def iop_smart_proxy_url
-        @iop_smart_proxy_url ||= ForemanRhCloud.iop_smart_proxy.url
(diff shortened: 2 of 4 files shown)


jeremylenz:
rebased.

lib/foreman_rh_cloud/plugin.rb
```diff
@@ -153,7 +151,7 @@ def self.register
         end
 
         ::Foreman::Plugin.app_metadata_registry.register(:foreman_rh_cloud, {
-          iop: ForemanRhCloud.with_iop_smart_proxy?,
+          iop: -> { ForemanRhCloud.with_iop_smart_proxy? },
```
chris1984, line 154: Had to ask AI what this is and it makes sense now as a proc which goes along with the foreman change

chris1984 (approved):
Thanks for cleaning up the urls and now with the ping conditional gone, we can work on adding the status to hammer ping. Tests look good, helped explain the proc lambda change you made a little better seeing it in action in the test case. ACK

nofaralfasi (approved):
LGTM!
The approach is solid: wrapping the IoP check in a lambda ensures runtime evaluation instead of stale startup values. The defensive return {} in Ping.ping and removing the URL memoization are the right calls.
Nice comprehensive test coverage too.
I haven't tested this locally, but the code looks correct.

nofaralfasi:
> LGTM! The approach is solid: wrapping the IoP check in a lambda ensures runtime evaluation instead of stale startup values. The defensive return {} in Ping.ping and removing the URL memoization are the right calls. Nice comprehensive test coverage too. I haven't tested this locally, but the code looks correct.

**Update:**
I tested this together with https://github.com/theforeman/foreman/pull/10814 and it works as expected.

PR #1141, opened 2026-01-15
Display 'Analysis disabled' in CVE column

jeremylenz:

Hey, we've got some upgrades to the __Total CVEs__ column for IoP-enabled systems!

image

1. Display "Analysis disabled" for hosts that have vulnerability analysis disabled
2. Instead of a QuestionIcon, display '—' like we do for other columns. This will show for unregistered hosts, or when data is missing.
3. Change the weight of the Total CVEs column so that 'Content view environments' doesn't show up between Recommendations and Total CVEs
4. use propsToCamelCase so we can stop disabling that silly eslint rule

everything is well considered!!

This code is :sparkles: artisan :sparkles: and no AI was used in its creation

Hosts > All Hosts > Manage Columns > add the Total CVEs column

```js
// CVECountCell.js line 36
// Change this 
if (optOut === true) {
// to this
if (true) {
```

Alternatively, you could figure out the proper curl command to send IoP a `[PATCH /api/vulnerability/v1/systems/opt_out](https://console.redhat.com/docs/api/vulnerability/v1#operations-default-manager\.system_handler\.PatchBulkSystemsOptOut\.patch)`

You should now see "Analysis disabled" for your host.
Additionally, you should see '—' and NOT a QuestionIcon for unregistered hosts.
You should continue to see a number of vulnerabilities (with link) for hosts with vulnerabilities.

### DIFF
--- a/webpack/InsightsVulnerabilityHostIndexExtensions/__tests__/CVECountCell.test.js
+++ b/webpack/InsightsVulnerabilityHostIndexExtensions/__tests__/CVECountCell.test.js
@@ -1,5 +1,6 @@
 import React from 'react';
 import { screen } from '@testing-library/react';
+import '@testing-library/jest-dom';
 import { rtlHelpers } from 'foremanReact/common/rtlTestHelpers';
 import { API } from 'foremanReact/redux/API';
 import { CVECountCell } from '../CVECountCell';
@@ -54,10 +55,10 @@ describe('CVECountCell', () => {
         },
       },
     });
-    expect(screen.getByRole('img', { hidden: true })).toBeTruthy();
+    expect(screen.getByText('—')).toBeInTheDocument();
   });
 
-  it('renders UnknownIcon when IoP is not enabled', () => {
+  it('renders — when IoP is not enabled', () => {
     renderWithStore(<CVECountCell hostDetails={hostDetailsMock} />, {
       router: {
         location: {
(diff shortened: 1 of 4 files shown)


webpack/InsightsVulnerabilityHostIndexExtensions/__tests__/CVECountCell.test.js
```diff
@@ -105,10 +106,10 @@ describe('CVECountCell', () => {
       },
     });
     // Should render UnknownIcon when CVE API fails
```
nofaralfasi: This comment should be updated to match the new behavior (rendering '—').

nofaralfasi (commented):
Tested the code locally and it works as expected - hosts with vulnerability analysis disabled now correctly show "Analysis disabled" instead of the CVE count, and the em-dash replacement for unknown states is consistent with the rest of the UI.
One suggestion: consider adding a test case for the new optOut feature to ensure it's covered.

jeremylenz:
@nofaralfasi updated :+1:

nofaralfasi:
> One suggestion: consider adding a test case for the new optOut feature to ensure it's covered.

@jeremylenz What do you think about adding tests for this?

jeremylenz:
oops, I saw that and then forgot :smile: working on it!

jeremylenz:
@nofaralfasi test added. :+1:

nofaralfasi (approved):
LGTM. The test refactoring is a nice improvement over the previous approach, and the opt-out feature now has proper coverage.

ColeHiggins2 (approved):

PR #1142, opened 2026-01-16
Add vulnerability analysis actions to host table row kebab menu

jeremylenz:
## Summary
- Add actions to All Hosts table row kebab for enable and disable vulnerability analysis (IoP-only)
- Check current status and only show the relevant action
- No actions shown for unregistered hosts or hosts without system records

## Implementation
Uses Redux update pattern:
- CVECountCell updates HOSTS Redux data when CVE data loads
- Kebab actions read opt_out from host object or CVE cache as fallback
- After toggle, both CVE cache and HOSTS data are updated to trigger re-renders

## Behavior
- Initial load: No actions shown until CVE data loads (~500ms)
- After CVE loads: Shows only "Enable" OR "Disable" based on opt_out status
- After toggle: Both CVE column and kebab action update immediately
- Empty data: No actions shown if host not in vulnerability system

---
**Note:** This branch is built on top of [another PR (Display 'Analysis disabled' in CVE column)
](https://github.com/theforeman/foreman_rh_cloud/pull/1141)

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -54,7 +54,8 @@
           "theforeman",
           "tooltip",
           "unmount",
-          "vnd"
+          "vnd",
+          "vuln"
         ],
         "minLength": 3
       }
--- a/webpack/ForemanColumnExtensions/index.js
+++ b/webpack/ForemanColumnExtensions/index.js
@@ -79,9 +79,10 @@ const hostsIndexColumnExtensions = [
     categoryName: insightsCategoryName,
     categoryKey: 'insights',
     isSorted: false,
-    // eslint-disable-next-line camelcase
     // Show column unless IoP is explicitly disabled (handles undefined during loading)
-    isRelevant: contextData => contextData?.metadata?.foreman_rh_cloud?.iop !== false,
+    isRelevant: contextData =>
+      // eslint-disable-next-line camelcase
+      contextData?.metadata?.foreman_rh_cloud?.iop !== false,
   },
 ];
 
--- a/webpack/ForemanColumnExtensions/index.js
+++ b/webpack/ForemanColumnExtensions/index.js
@@ -80,7 +80,8 @@ const hostsIndexColumnExtensions = [
     categoryKey: 'insights',
     isSorted: false,
     // eslint-disable-next-line camelcase
-    isRelevant: contextData => contextData?.metadata?.foreman_rh_cloud?.iop,
+    // Show column unless IoP is explicitly disabled (handles undefined during loading)
+    isRelevant: contextData => contextData?.metadata?.foreman_rh_cloud?.iop !== false,
   },
 ];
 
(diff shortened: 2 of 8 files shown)


jeremylenz:
@nofaralfasi Updated and also simplified the Redux code a bit, see what you think

jeremylenz:
By the way, if you're sorting the table and seeing weird behavior, check https://github.com/theforeman/foreman/pull/10830

nofaralfasi:
> By the way, if you're sorting the table and seeing weird behavior, check [theforeman/foreman#10830](https://github.com/theforeman/foreman/pull/10830)

Something isn’t working with the changes from PR [#10830](https://github.com/theforeman/foreman/pull/10830).
After enabling the analysis using the new row kebab action, the Total CVEs column doesn’t update immediately. It only shows the correct value after refreshing the page.
This issue does not occur on the Foreman develop branch.

nofaralfasi (approved):
LGTM! Tested locally and the enable/disable actions work as expected - the kebab menu correctly shows only the relevant action based on the current `opt_out` state, and toggling updates both the CVE column and the action immediately.
The approach of reusing the CVE column's API data for the kebab actions makes sense and avoids duplicate requests.

jeremylenz:
> After enabling the analysis using the new row kebab action, the Total CVEs column doesn’t update immediately. It only shows the correct value after refreshing the page.

This one was tricky!
```
Root Cause

  When enabling vulnerability analysis, cve_count was null (host not scanned yet), but CVECountCell only checked for undefined, causing it to render <Link>null</Link> which appeared as a blank cell.

  Changes Made

  1. CVECountCell.js - Fixed null handling

  - Changed cveCount === undefined to cveCount == null to handle both null and undefined
  - This ensures "—" is displayed when CVE count isn't available yet

  2. VulnerabilityAnalysisActions.js - Added refetch on enable

  - When disabling: Optimistically updates cache to show "Analysis disabled" immediately
  - When enabling: Skips optimistic update and dispatches a GET request to fetch fresh CVE data
  - Added safety check for missing data in updateData

  3. ForemanColumnExtensions/index.js - Fixed loading state

  - Changed isRelevant from contextData?.metadata?.foreman_rh_cloud?.iop to !== false
  - This shows the column during metadata loading instead of hiding it

  4. Tests Updated

  - Added test for null cve_count scenario
  - Added test for zero CVE count (edge case)
  - Split tests for enable vs disable actions
  - Added test for refetch behavior when enabling
  - All 173 tests passing ✅

  The fix ensures:
  - ✅ Disabling shows "Analysis disabled" immediately (optimistic)
  - ✅ Enabling shows "—" briefly, then fetches and displays actual CVE count
  - ✅ No more blank cells
  - ✅ Column always visible (no flash during metadata load)
```

Try it now..

nofaralfasi:
> > After enabling the analysis using the new row kebab action, the Total CVEs column doesn’t update immediately. It only shows the correct value after refreshing the page.
> 
> This one was tricky!
> Try it now..

Fixed the issue! Works well now. Thanks @jeremylenz!

jeremylenz:
Will wait for @ColeHiggins2 ack to merge.

chris1984 (approved):
Tested as well and works great, code looks fine.

chris1984:
Since Cole is out this week, merging per Justin so we make feature freeze from dev side, automation can be written after based on a release inside a stream build

PR #1148, opened 2026-02-02
SAT-38683 - Support any location

lfu:
When setting "Any location" on the location dropdown menu, accessing either Remediations or Vulnerabilities menu should work.

Get an IoP server.
Change the location to "Any location" and visit either the Remediations or Vulnerabilities menu.
Data should be displayed.

## Summary by Sourcery

Allow Insights Cloud requests and tag updates to work when a user selects Any location (nil location).

New Features:
- Support forwarding Insights Cloud UI requests when the session has no location selected (Any location).

Bug Fixes:
- Ensure tag generation works when location is nil by using a wildcard location tag and relaxing host filtering on location.

Tests:
- Add controller tests for forwarding requests with nil and non-nil locations, and a unit test for tag generation with a wildcard location.

### DIFF
--- a/app/controllers/insights_cloud/ui_requests_controller.rb
+++ b/app/controllers/insights_cloud/ui_requests_controller.rb
@@ -2,7 +2,7 @@ module InsightsCloud
   class UIRequestsController < ::ApplicationController
     layout false
 
-    before_action :ensure_org, :ensure_loc, :only => [:forward_request]
+    before_action :ensure_org, :find_location, :only => [:forward_request]
 
     # The method that "proxies" requests over to Cloud
     def forward_request
@@ -93,9 +93,8 @@ module InsightsCloud
       return render_message 'Organization not found or invalid', :status => 400 unless @organization
     end
 
-    def ensure_loc
+    def find_location
       @location = Location.current
-      return render_message 'Location not found or invalid', :status => 400 unless @location
     end
 
     def base_url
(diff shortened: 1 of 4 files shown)


app/controllers/insights_cloud/ui_requests_controller.rb
```diff
@@ -95,7 +95,6 @@ def ensure_org
 
     def ensure_loc
       @location = Location.current
-      return render_message 'Location not found or invalid', :status => 400 unless @location
     end
```
jeremylenz: Should remove this entire method and callback, if we're not going to use it.

jeremylenz (commented):
Thanks @lfu!

General question - Why only Location and not Org as well? What would it look like to fix both?

Haven't tested yet but one nit below

app/controllers/insights_cloud/ui_requests_controller.rb
lfu: This method is still required for location setting. 
Just it does not reject any location.

jeremylenz: ah, gotcha. Then it should be renamed `find_location` for consistency with similar methods.

LadislavVasina1 (approved):
ACK, from the QE side, with packit build, Recommendation and Vulnerability pages are not displaying errors anymore, while the Any Location is selected, as they were previously.

Agree with the question raised by Jeremy about the "Any Organization" functionality.

nofaralfasi (approved):
Tested the changes and verified expected behavior
The PR works as expected and the code looks good.

nofaralfasi:
> Agree with the question raised by Jeremy about the "Any Organization" functionality.

I believe we decided not to implement this for now and to track it separately as its own issue.
cc to @jeremylenz.

PR #1159, opened 2026-02-25
SAT-42791 - Pass docs links through LinksController

ofedoren:

Wrapping a static link to subscription service docs via foremanUrl, so it goes through LinksController.

This is needed so we can properly redirect this link to RHOKP.

 - Use related PR in foreman_theme_satellite: https://github.com/RedHatSatellite/foreman_theme_satellite/pull/313
 - Spin RHOKP
 - Change Settings > satellite_documentation_url to the RHOKP
 - See if the links are being redirected there instead of docs.redhat.com

## Summary by Sourcery

Route the subscriptions service documentation link through Foreman’s LinksController instead of using a static URL.

New Features:
- Add helper to generate the subscriptions service docs URL via the Foreman links endpoint.

Enhancements:
- Update the inventory upload page description to use the new helper-based subscriptions docs link for redirect support.

### DIFF
--- a/webpack/ForemanInventoryUpload/Components/PageHeader/components/PageDescription/PageDescription.js
+++ b/webpack/ForemanInventoryUpload/Components/PageHeader/components/PageDescription/PageDescription.js
@@ -7,6 +7,8 @@ import { getDocsURL } from 'foremanReact/common/helpers';
 import { FormattedMessage } from 'react-intl';
 import { selectSubscriptionConnectionEnabled } from '../../../InventorySettings/InventorySettingsSelectors';
 
+import { getSubscriptionServiceDocsUrl } from '../../../../ForemanInventoryHelpers';
+
 export const PageDescription = () => {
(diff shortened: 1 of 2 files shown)


ofedoren:
> * The new `getSubscriptionServiceDocsUrl` helper encodes a different docs path (`html-single/.../index`) than the original link (`html/.../index`); if this change is not intentional, consider aligning the URL to preserve the original target and avoid unexpected redirects.

This is intentional, we need all the links to point to html-single, this one was missed.

> * Since the subscription service docs URL is now abstracted into `getSubscriptionServiceDocsUrl`, consider scanning for any remaining hard-coded links to the same documentation and routing them through this helper for consistency.

I didn't find any that applicable.

adamruzicka (approved):
LGTM, but let's hold this open until the counterpart in f-theme-satellite is also approved.

adamruzicka:
Thank you @ofedoren !

PR #1160, opened 2026-02-25
Release 14.0.1

chris1984:
## Summary by Sourcery

Bump the project version to 14.0.1 for both the Ruby gem and the npm package metadata.

Build:
- Update Ruby gem version constant to 14.0.1.
- Update npm package.json version to 14.0.1.


jeremylenz (commented):
The major version bump is so it aligns with the new Sat y version, right?

chris1984:
> The major version bump is so it aligns with the new Sat y version, right?

Yeah, and we just did branching for 6.19 as well which is version 13.y.z, so figured now is the time of year again to bump the X number to match the new 6.y version

jeremylenz (approved):

PR #1164, opened 2026-03-03
Fix CVE details page 400 error by respecting no-tags patterns

nofaralfasi:
The scope_request? method was incorrectly filtering to only consider patterns with tag_name defined, which caused the CVE details endpoint pattern (api/vulnerability/v1/cves/[^/]+$) to be excluded.

This resulted in the more general pattern (api/vulnerability/v1/.*) being selected instead, which adds tags to the request. Since the CVE details API endpoint does not support tags per the OpenAPI spec, this caused a 400 Bad Request error:
"Extra query parameter(s) tags not in spec"

The fix removes the tag_name filter so all matching patterns are considered, then the most specific pattern (longest regex) is selected. For the CVE details endpoint, this correctly returns nil for tag_name, meaning no tags are added to the request.

Fixes the "Cannot read properties of undefined (reading 'synopsis')" error that occurred when clicking on any CVE from the Vulnerabilities page or host details Vulnerabilities tab.

Scenario 1:
Navigation Path:
Insights → Vulnerability → Click on any CVE

Scenario 2:
Navigation Path:
Hosts → All Hosts → client.example.com → Vulnerabilities → Click on any CVE

Expected Result:
Clicking on any CVE should open the Vulnerability details page and display full CVE information, including synopsis, description, severity, affected packages, etc.

## Summary by Sourcery

Bug Fixes:
- Fix 400 errors and frontend failures on CVE detail pages by correctly handling tag-less vulnerability CVE endpoints in scoped request matching.

### DIFF
--- a/app/services/foreman_rh_cloud/insights_api_forwarder.rb
+++ b/app/services/foreman_rh_cloud/insights_api_forwarder.rb
@@ -224,14 +224,21 @@ module ForemanRhCloud
     def scope_request?(original_request, path)
       return nil unless original_request.get?
 
-      # Only consider patterns that define tag_name - this ensures patterns without
-      # tag_name (permission-only entries) cannot override tag-supporting patterns
-      matching_patterns = SCOPED_REQUESTS.select { |pattern| pattern[:tag_name] && pattern[:test].match?(path) }
+      # Find patterns that match this path AND are relevant for GET requests.
+      # A pattern is relevant if it either:
+      # - Has tag_name defined (supports scoping)
+      # - Has GET permissions defined (explicitly handles GET, even without tags)
+      # This ensures patterns like api/vulnerability/v1/cves/[^/]+$ (GET without tags)
+      # take precedence over general patterns, while POST-only patterns are ignored.
+      matching_patterns = SCOPED_REQUESTS.select do |pattern|
+        pattern[:test].match?(path) && (pattern[:tag_name] || pattern.dig(:permissions, 'GET'))
+      end
       return nil if matching_patterns.empty?
 
-      # Choose the most specific pattern by regex source length for consistency
-      # with required_permission_for behavior
+      # Choose the most specific pattern by regex source length
       request_pattern = matching_patterns.max_by { |pattern| pattern[:test].source.length }
+
+      # Return the tag_name (may be nil if the most specific pattern doesn't support tags)
       request_pattern[:tag_name]
     end
 


jeremylenz (commented):
I tested this and it fixes the issue.

Will leave it to @vkrizan to see if he agrees with bot's comments here. also the test failure seems related :)

nofaralfasi:
@sourcery-ai review

jeremylenz:
> * The new `pattern.dig(:permissions, 'GET')` check assumes permission keys are strings; if any entries use symbol keys (e.g. `:GET`), this will silently skip them—consider normalizing or supporting both.

If `.with_indifferent_access` is called on these at any point, we can use either string or hash keys and not worry about it.

app/services/foreman_rh_cloud/insights_api_forwarder.rb
```diff
@@ -224,14 +224,25 @@ def original_headers(original_request)
...
-      # tag_name (permission-only entries) cannot override tag-supporting patterns
-      matching_patterns = SCOPED_REQUESTS.select { |pattern| pattern[:tag_name] && pattern[:test].match?(path) }
+      # Find patterns that match this path AND are relevant for GET requests.
+      # A pattern is relevant if it either:
+      # - Has tag_name defined (supports scoping)
+      # - Has GET permissions defined (explicitly handles GET, even without tags)
+      # This ensures patterns like api/vulnerability/v1/cves/[^/]+$ (GET without tags)
+      # take precedence over general patterns, while POST-only patterns are ignored.
+      matching_patterns = SCOPED_REQUESTS.select do |pattern|
+        pattern[:test].match?(path) && (pattern[:tag_name] || pattern.dig(:permissions, 'GET'))
```
vkrizan, line 234: I see that this combines two unrelated things: 1) that it needs to have a tag definition; or 2) it needs to have a GET permissions.

Could you elaborate on the requirement for testing the existence of a tag definition? Could there be just one requirement for `GET` permissions definition? Or is this kept for a backwards compatibility?

app/services/foreman_rh_cloud/insights_api_forwarder.rb
```diff
@@ -224,14 +224,25 @@ def original_headers(original_request)
...
       return nil if matching_patterns.empty?
 
-      # Choose the most specific pattern by regex source length for consistency
-      # with required_permission_for behavior
-      request_pattern = matching_patterns.max_by { |pattern| pattern[:test].source.length }
+      # Choose the most specific pattern by regex source length.
+      # Tie-break by preferring tag-supporting patterns when specificity is equal,
+      # to avoid order-dependent behavior in SCOPED_REQUESTS.
+      request_pattern = matching_patterns.max_by do |pattern|
+        [pattern[:test].source.length, pattern[:tag_name] ? 1 : 0]
```
vkrizan: This matching condition for preferring matches by tag definition could be simplified just to the rule of longest match on path. Could you elaborate on the need for that tag preference over simplification?

vkrizan (commented):
Thanks for this fix. I've left some comments for possible simplification of the matching criteria.

app/services/foreman_rh_cloud/insights_api_forwarder.rb
nofaralfasi: Let me explain why I beleieve both conditions are needed.
Looking at the SCOPED_REQUESTS patterns, there are different categories:
**Category A**: Patterns with `tag_name` but NO permissions defined:
These patterns exist purely for tagging - they don't enforce permissions but do add tags to GET requests. Without the `tag_name` check, these would be excluded.

**Category B:** Patterns with GET permissions but NO `tag_name`:
These patterns explicitly say "this GET endpoint does NOT support tags". Without the GET permissions check, these would be excluded (the bug we're fixing).

If we only used `pattern.dig(:permissions, 'GET')`:
- `api/inventory/.* and api/tasks/.*` would be excluded (breaking change)
- GET requests to those endpoints would no longer get tags

So yes, it combines two concepts, but both are required to:

- Fix the CVE details bug (needs GET check)
- Maintain backwards compatibility for tagging-only patterns (needs `tag_name` check)

app/services/foreman_rh_cloud/insights_api_forwarder.rb
nofaralfasi: You're right - the tie-breaking logic for preferring tag-supporting patterns adds unnecessary complexity.
The simple "longest match wins" rule is sufficient.

nofaralfasi:
> > * The new `pattern.dig(:permissions, 'GET')` check assumes permission keys are strings; if any entries use symbol keys (e.g. `:GET`), this will silently skip them—consider normalizing or supporting both.
> 
> If `.with_indifferent_access` is called on these at any point, we can use either string or hash keys and not worry about it.

I checked and all permission entries in SCOPED_REQUESTS currently use string keys ('GET', 'POST', etc.), so `pattern.dig(:permissions, 'GET')` should be safe here.
This is also consistent with how `required_permission_for` accesses permissions using string keys.
That said, if you think it's worth adding a safeguard for future-proofing (like supporting both or adding a comment), I'm happy to do that.

jeremylenz:
> That said, if you think it's worth adding a safeguard for future-proofing (like supporting both or adding a comment), I'm happy to do that.

If it works as-is I'm fine without it.

app/services/foreman_rh_cloud/insights_api_forwarder.rb
vkrizan: Thanks for the explanation!

vkrizan (approved):

jeremylenz (approved):
Thanks @nofaralfasi!

ACK 👍

PR #1032, opened 2025-07-17
Add assert single commit action

chris1984:
## Summary by Sourcery

CI:
- Add non-blocking 'Assert Single Commit' GitHub Actions workflow to enforce exactly one commit per pull request

PR #1167, opened 2026-03-04
Fix single-commit check for stable branch PRs

zjhuntin:
## Summary

The single-commit workflow added in #1032 enforces one commit per PR to ensure squashed commits before merging. However, it was hardcoded to compare against `develop`, which breaks for PRs targeting stable branches like `foreman_3_18`.

This change uses `github.base_ref` to dynamically compare against the PR's actual target branch.

## Problem

PRs like #1166 (Release 13.2.3) targeting `foreman_3_18` fail the commit count check because the stable branch has diverged from `develop`.

## Solution

Replace hardcoded `develop` with `${{ github.base_ref }}` so the check works for PRs targeting any branch.

## Summary by Sourcery

CI:
- Adjust the single-commit check workflow to fetch and create a local branch from github.base_ref and use it as the reference for the commit count check.

### DIFF
--- a/.github/workflows/single-commit.yml
+++ b/.github/workflows/single-commit.yml
@@ -13,9 +13,9 @@ jobs:
       uses: actions/checkout@v3
       with:
         fetch-depth: 30
-    - name: Checkout develop
-      run: git fetch origin develop
-    - name: create local develop branch
-      run: git branch develop origin/develop
+    - name: Checkout base branch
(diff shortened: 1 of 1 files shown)


chris1984 (approved):

jeremylenz:
Awesome, thanks @zjhuntin!

Seems https://github.com/Katello/katello/blob/master/.github/workflows/pull_request.yml will need the same fix.

PR #1178, opened 2026-03-17
Fix InsightsClientReportStatus when path has no trailing slash

jeremylenz:
## Summary
Fixes a bug where `InsightsClientReportStatus` shows "N/A" instead of "Reporting" after successful insights-client uploads.

## Root Cause
The `upload_success?` method in `MachineTelemetriesController` was checking for paths containing `/redhat_access/r/insights/uploads/` (with trailing slash). This failed to match upload requests with paths like `/redhat_access/r/insights/uploads` (without a UUID suffix). `insights-client --test-connection` makes requests to this path.

## The Fix
- Changed from `.include?('/redhat_access/r/insights/uploads/')` to `.start_with?('/redhat_access/r/insights/uploads')`
- Removed trailing slash requirement
- Now correctly matches both path formats:
  - `/redhat_access/r/insights/uploads` ✓
  - `/redhat_access/r/insights/uploads/UUID` ✓

## Impact
Instead of remaining "N/A", hosts will now correctly show "Reporting" status both after a successful insights-client upload, and after a run of `insights-client --test-connection`.

## Test Plan
- [x] Added debug logging to identify the issue
- [x] Confirmed path mismatch in CI logs
- [ ] Verify status updates correctly in robottelo test: `test_insights_registration_with_capsule[rhel9-ipv6]`

### DIFF
--- a/app/controllers/insights_cloud/api/machine_telemetries_controller.rb
+++ b/app/controllers/insights_cloud/api/machine_telemetries_controller.rb
@@ -118,13 +118,10 @@ module InsightsCloud::Api
     end
 
     def update_host_insights_status
-      logger.info("update_host_insights_status called - @host: #{@host&.name}, upload_success?: #{upload_success?}, request.path: #{request.path}, @cloud_response.code: #{@cloud_response&.code}")
       return unless upload_success?
-      logger.info("Updating insights status for host #{@host.name}")
       # create insights status if it wasn't there in the first place and refresh its reporting date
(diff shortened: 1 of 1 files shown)


nofaralfasi (approved):
Following the Slack discussion, the code looks good to me. I believe running the automation for this PR should be sufficient.

jeremylenz:
Closing as this is not the correct approach.
* The host status checks if the host is uploading reports
* `insights-client --test-connection` does not upload a report
* therefore, it should not be expected that it updates the host status.

PR #1179, opened 2026-03-17
SAT-42831 - Fix stale recommendations/CVEs data when switching hosts on All Hosts page

lfu:
Add `key` props to `ScalprumComponent` instances so they remount when navigating between different hosts, instead of keeping stale internal state from the previously viewed host.
All hosts -> Manage columns -> select recommendations as a column and save

Now All hosts page shows the number of recommendations for each host.

Click on the number of recommendation for host A -> Click all hosts again -> Click the number of recommendation for host B

**Before**
It shows the recommendations for host A instead of host B
Same happens for CVEs as well
(I could not see the behavior with my testing environments - both stream iop devel box and 6.19 iop devel box)

**After**
Correct recommendations/CVEs for each host should be displayed

## Summary by Sourcery

Ensure host details Scalprum components remount when switching between hosts to prevent stale vulnerability and recommendations data.

Bug Fixes:
- Fix stale CVEs and recommendations data shown when navigating between different hosts on the All Hosts page and host details tabs.

Tests:
- Add tests to verify that the CVEs host details Scalprum component unmounts and remounts when the system ID changes.

### DIFF
--- a/webpack/CVEsHostDetailsTab/CVEsHostDetailsTab.js
+++ b/webpack/CVEsHostDetailsTab/CVEsHostDetailsTab.js
@@ -10,7 +10,12 @@ const CVEsHostDetailsTab = ({ systemId }) => {
   const module = './SystemDetailTable';
   return (
     <div className="rh-cloud-insights-vulnerability-host-details-component vulnerability">
-      <ScalprumComponent scope={scope} module={module} systemId={systemId} />
+      <ScalprumComponent
+        key={systemId}
+        scope={scope}
+        module={module}
+        systemId={systemId}
+      />
     </div>
   );
 };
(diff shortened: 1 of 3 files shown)


jeremylenz (commented):
Thanks @lfu! 

Fixes the original issue, https://redhat.atlassian.net/browse/SAT-42831

It also addresses https://redhat.atlassian.net/browse/SAT-38703 - but I still see odd behavior when switching to a host that does not have a Vulnerabilties tab. However, this fix fixes switching between 2 hosts that both have the tab.

Lint needs fixing

lfu:
I don't see the odd behavior with both my stream and 6.19 devel boxes.
So just fixed the issue by logic :smile:

jeremylenz (approved):

PR #1182, opened 2026-03-23
Improve inventory stats with USER_OMITTED status and enhanced UX

jeremylenz:
### What are the changes introduced in this pull request?
First, some background.

Insights hosts have host statuses:
* `InsightsClientReportStatus` - Tracks whether `insights-client` is regularly reporting for this host.
* `InventoryStatus`- Tracks whether this host has a record in the hosted Inventory service. (These records come from inventory upload reports sent from rh_cloud. They will also be created if insights-client sends a report.)

And host parameters:
* `host_registration_insights` - Decides two things:
  * whether insights-client is installed on this host as part of the generated global registration script
  * whether requests from insights-client get forwarded to the cloud or blocked for this host
* `host_registration_insights_inventory` - Decides whether the host's information is included in inventory upload reports.

Users can set these params individually. On the Inventory Upload page, if you click "Sync all inventory status," you get a toast notification with a tally of
* all registered hosts
* Hosts uploaded to the hosted inventory ("synced")
* "disconnected" hosts (those that are not present in the hosted inventory)

There were a few problems with this -
* Hosts that were excluded on purpose, via the host parameter, were not accounted for in this tally. ("Total registered hosts" was just "disconnected + synced".)
* Once the toast was gone, you could review the same information by looking at the Inventory status. But that was not highlighted.

So, the improvements here are

* Introduce a third category in the notification, USER_OMITTED, to account for the "user-omitted" hosts.
* Also introduce a USER_OMITTED __status value__ for both `InventoryStatus` AND `InsightsClientReportStatus`.
* Update and improve the wording of both the notification and the host status descriptions.
* Side bonus improvement: Mark `InventoryStatus` as not relevant for IoP setups. In IoP, this status is ignored because we use single-host reports and data is never sent to the cloud.
* Also added scoped_search support.

### Considerations taken when implementing this change?
It took quite a while to get this framework in my head. Also, the code that updates host status is a bit complicated, so I hope I didn't do anything wrong. But I tried to stay within the existing frameworks.

### What are the testing steps for this pull request?
Setup: NON-IoP (hosted). Get 3 registered hosts.

It helped me to have 3 host groups, one host in each. This way you can add the "Host group" column on the All Hosts page, and immediately know which host you're dealing with.

'Yes' hostgroup - both host params set to true
'Mixed' hostgroup - `host_registration_insights` set to false, `host_registration_insights_inventory` set to true (this one will show 'Disconnected' status for ~24h or until hosted HBI gives it a uuid)
'No' hostgroup - both host params set to false (this one should show the new USER_OMITTED status)

Now -
Insights > Inventory upload > Sync all inventory status

1. New toast notification should have correct tallies
2. Look! I added links! Click them. isn't that cool? 😄
3. Inspect each host status and make sure it's as expected. user_omitted status will show as OK / green.
4. Try searching hosts for `insights_client_report_status = user_omitted` or `insights_inventory_sync_status = user_omitted` - these should work.

### DIFF
--- a/lib/inventory_sync/async/inventory_full_sync.rb
+++ b/lib/inventory_sync/async/inventory_full_sync.rb
@@ -17,6 +17,11 @@ module InventorySync
         @subscribed_hosts_ids = Set.new(affected_host_ids)
         @omitted_ids = Set.new(user_omitted_host_ids)
 
+        # Remove user-omitted hosts from subscribed set. In normal operation, affected_host_ids
+        # already excludes user-omitted hosts via for_slice, but this handles edge cases like
+        # a host transitioning from uploaded to user-omitted between syncs.
+        @subscribed_hosts_ids.subtract(@omitted_ids)
+
         InventorySync::InventoryStatus.transaction do
           InventorySync::InventoryStatus.where(host_id: @subscribed_hosts_ids).delete_all
           yield
--- a/lib/inventory_sync/async/inventory_full_sync.rb
+++ b/lib/inventory_sync/async/inventory_full_sync.rb
@@ -24,6 +24,7 @@ module InventorySync
 
         InventorySync::InventoryStatus.transaction do
           InventorySync::InventoryStatus.where(host_id: @subscribed_hosts_ids).delete_all
+          InventorySync::InventoryStatus.where(host_id: @omitted_ids).delete_all
           yield
           add_missing_hosts_statuses(@subscribed_hosts_ids) # any remaining hosts after yield are disconnected
           add_user_omitted_host_statuses(@omitted_ids)
--- a/lib/inventory_sync/async/inventory_full_sync.rb
+++ b/lib/inventory_sync/async/inventory_full_sync.rb
@@ -100,12 +100,13 @@ module InventorySync
       def user_omitted_host_ids
         param_name = InsightsCloud.enable_client_param_inventory
 
-        # Load parameters (not hosts) and filter using Foreman::Cast.to_bool
-        Parameter.where(
-          name: param_name,
-          reference_id: Host.unscoped.where(organization: organizations).select(:id),
-          type: 'HostParameter'
-        ).select { |param| ::Foreman::Cast.to_bool(param.value) == false }.map(&:reference_id)
+        # Use search_for to respect parameter inheritance (global, org, hostgroup, host)
+        # This matches the same logic used by for_slice, ensuring consistency
+        Host.unscoped
+            .where(organization: organizations)
+            .joins(:subscription_facet)
+            .search_for("params.#{param_name} = f")
+            .pluck(:id)
       end
     end
   end
--- a/lib/inventory_sync/async/inventory_full_sync.rb
+++ b/lib/inventory_sync/async/inventory_full_sync.rb
@@ -57,14 +57,6 @@ module InventorySync
         InventorySync::InventoryStatus.create(status_hashes)
         updated_ids = status_hashes.map { |hash| hash[:host_id] }
         @subscribed_hosts_ids.subtract(updated_ids)
-        # also refresh the InsightsClientReportStatus so hosts that are user-omitted will not show as irrelevant
-        # Eager load associations to avoid N+1 queries: parameters used in to_status, host_statuses used by get_status
-        ::Host.where(id: updated_ids)
-              .includes(:parameters, :host_statuses)
-              .each do |host|
-          insights_client_report_status = host.get_status(::InsightsClientReportStatus)
-          insights_client_report_status.refresh!
-        end
       end
 
       def add_missing_hosts_statuses(hosts_ids)
--- a/lib/inventory_sync/async/inventory_full_sync.rb
+++ b/lib/inventory_sync/async/inventory_full_sync.rb
@@ -23,9 +23,7 @@ module InventorySync
           add_missing_hosts_statuses(@subscribed_hosts_ids) # any remaining hosts after yield are disconnected
           add_user_omitted_host_statuses(@omitted_ids)
           host_statuses[:disconnect] += @subscribed_hosts_ids.size
-          logger.debug("Disconnected hosts: #{@subscribed_hosts_ids.map { |id| Host.find(id).name }}")
           host_statuses[:user_omitted] += @omitted_ids.size
-          logger.debug("User-omitted hosts: #{@omitted_ids.map { |id| Host.find(id).name }}")
         end
 
         logger.debug("Synced hosts count: #{host_statuses[:sync]}")
@@ -57,7 +55,7 @@ module InventorySync
         # also refresh the InsightsClientReportStatus so hosts that are user-omitted will not show as irrelevant
         updated_ids.each do |host_id|
           host = ::Host.find_by(id: host_id)
-          next unless host.present?
+          next if host.blank?
           insights_client_report_status = host.get_status(::InsightsClientReportStatus)
           insights_client_report_status.refresh!
         end
--- a/lib/inventory_sync/async/inventory_full_sync.rb
+++ b/lib/inventory_sync/async/inventory_full_sync.rb
@@ -50,9 +50,17 @@ module InventorySync
       private
 
       def update_hosts_status(status_hashes)
+        # create Inventory statuses
         InventorySync::InventoryStatus.create(status_hashes)
         updated_ids = status_hashes.map { |hash| hash[:host_id] }
         @subscribed_hosts_ids.subtract(updated_ids)
+        # also refresh the InsightsClientReportStatus so hosts that are user-omitted will not show as irrelevant
+        updated_ids.each do |host_id|
+          host = ::Host.find_by(id: host_id)
+          next unless host.present?
+          insights_client_report_status = host.get_status(::InsightsClientReportStatus)
+          insights_client_report_status.refresh!
+        end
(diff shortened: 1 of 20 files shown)


lib/inventory_sync/async/inventory_full_sync.rb
```diff
@@ -15,16 +15,22 @@ def plan(organization)
...
       def setup_statuses
         @subscribed_hosts_ids = Set.new(affected_host_ids)
+        @omitted_ids = Set.new(user_omitted_host_ids)
 
         InventorySync::InventoryStatus.transaction do
           InventorySync::InventoryStatus.where(host_id: @subscribed_hosts_ids).delete_all
           yield
-          add_missing_hosts_statuses(@subscribed_hosts_ids)
+          add_missing_hosts_statuses(@subscribed_hosts_ids) # any remaining hosts after yield are disconnected
+          add_user_omitted_host_statuses(@omitted_ids)
```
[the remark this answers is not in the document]
jeremylenz: In regular mode, affected_host_ids uses Queries.for_slice which already filters out user-omitted hosts via search_for("not
  params.#{InsightsCloud.enable_client_param_inventory} = f") (line 54 in queries.rb). This means @subscribed_hosts_ids and @omitted_ids are mutually exclusive - no overlap.

  ▎ In IoP mode, even if there were overlap, the inventory status is irrelevant (relevant? returns false, showing "N/A") since IoP uses single-host reports instead of inventory sync.

lib/inventory_sync/async/inventory_full_sync.rb
```diff
@@ -44,9 +53,17 @@ def rescue_strategy_for_self
...
         InventorySync::InventoryStatus.create(status_hashes)
         updated_ids = status_hashes.map { |hash| hash[:host_id] }
         @subscribed_hosts_ids.subtract(updated_ids)
+        # also refresh the InsightsClientReportStatus so hosts that are user-omitted will not show as irrelevant
+        updated_ids.each do |host_id|
+          host = ::Host.find_by(id: host_id)
+          next if host.blank?
+          insights_client_report_status = host.get_status(::InsightsClientReportStatus)
+          insights_client_report_status.refresh!
+        end
```
lfu: This does a Host.find_by + get_status + refresh! per host inside a loop. For large syncs (thousands of hosts), this could be very slow. Consider batching or at minimum eager-loading the status association.

lib/inventory_sync/async/inventory_full_sync.rb
```diff
@@ -44,9 +53,17 @@ def rescue_strategy_for_self
...
       def update_hosts_status(status_hashes)
+        # create Inventory statuses
         InventorySync::InventoryStatus.create(status_hashes)
         updated_ids = status_hashes.map { |hash| hash[:host_id] }
         @subscribed_hosts_ids.subtract(updated_ids)
+        # also refresh the InsightsClientReportStatus so hosts that are user-omitted will not show as irrelevant
+        updated_ids.each do |host_id|
+          host = ::Host.find_by(id: host_id)
+          next if host.blank?
+          insights_client_report_status = host.get_status(::InsightsClientReportStatus)
```
lfu: If `InsightsClientReportStatus` is about whether the `insights-client` on the host is sending reports back to Satellite (not directly to Red Hat), it should not be affected by `InventorySync` in my opinion.

lfu:
I'm really confused by these parameters and their combination :confused: 
> 'Mixed' hostgroup - host_registration_insights set to false, host_registration_insights_inventory set to true (this one will show 'Disconnected' status for ~24h or until hosted HBI gives it a uuid)

Is that correct? 
Or it should be like the following?

| host_registration_insights | host_registration_insights_inventory | Meaning |
|----------|-----------|---------|
| true | true | Full cloud integration |
| true | false (global-level)| Report to foreman_rh_cloud, but don't upload to cloud inventory - Disconnected|
| true | false (host-level) | Report to foreman_rh_cloud, but don't upload to cloud inventory - User omitted|
| false | X | Fully excluded |

lib/inventory_sync/async/inventory_full_sync.rb
jeremylenz: You're right that it seems weird to update the `InsightsClientReportStatus` during `InventoryFullSync`. But let me explain why I put that here.

### Factor 1
A host status is updated in the db like
```rb
my_status.refresh!
```

That does two things:
1. calls the `.to_status` method of the status, which is where the actual logic happens for what status the host should be
2. saves the result of that status calculation to the db.

So I went looking for where the `InsightsClientReportStatus` is updated. Turns out it's only updated in one place, `MachineTelemetriesController`: https://github.com/jeremylenz/foreman_rh_cloud/blob/67416d833d7e3aaa7a59df720ea4463d52047aea/app/controllers/insights_cloud/api/machine_telemetries_controller.rb#L14

That's the controller that forwards requests from insights-client to the cloud. 

### Factor 2

Host statuses can implement a `def relevant?` method, if they so choose. If `status.relevant?` comes back false, the host status is displayed as "N/A" in the UI and is not used to calculate the global host status.

In the case of `InsightsClientReportStatus`, we have

```rb
# prevent creation of the status on global refresh, but show it if the record already exists
  def relevant?(_options = {})
    persisted?
  end
```

`persisted?` just means "is the record saved in the db?"

This means that

1. when a host is first registered, its `InsightsClientReportStatus` will not be `relevant?` because it will not yet be `persisted?` in the database.
2. Normally, when `insights-client` makes the first request to the cloud, it will hit `MachineTelemetriesController` which will run the `after_action :update_host_insights_status`, and the status will now be relevant.
3. However, if the user sets the param to false, that means we need to set the host status to USER_OMITTED. So, when do we write that status?

* `MachineTelemetriesController` has a `before_action :ensure_telemetry_enabled_for_consumer, :only => [:forward_request]` which blocks all requests if the param is set to false. So if the param is false, the status will always be irrelevant / "N/A" and we will never hit the `after_action :update_host_insights_status`. So we would never have an opportunity to set it to USER_OMITTED.
* I guess another possibility would be to have some sort of hook on the host parameter change, but I didn't look into that much because it seems like it might be brittle and unreliable and I don't know if any such hooks exist.

So what I ended up with is updating the `InsightsClientReportStatus` during the full sync here. I know it's not the best, but I couldn't think of a better idea.

jeremylenz:
> it should be like the following?

All we know is what parameter the host has. I don't think it makes sense to figure out if the parameter is inherited or not, and change behavior based on that.

In my experience testing it this week, the flow for `host_registration_insights` set to false, `host_registration_insights_inventory` set to true is:

1. Host is uploaded to HBI with the first report - HBI does not know about the host, so it shows "Disconnected" (my understanding is that Disconnected status is shown for any hosts that rh_cloud knows about, but are not included in the hosted HBI response).
2. At some point later (24 hours?) HBI assigns a UUID to the host.
3. During the next report upload, rh_cloud sees the host in the response and sees that it now has a UUID assigned by HBI. An insights_facet is created for the host with that UUID.
4. Now, the host status is changed to SYNC / "Connected."

lib/inventory_sync/async/inventory_full_sync.rb
jeremylenz: Updated to add eager loading. :+1:

lib/inventory_sync/async/inventory_full_sync.rb
jeremylenz: I asked Claude about it and it suggests modifying `ensure_telemetry_enabled_for_consumer` to update the status even when blocking the request. I think that's a great idea, let me try it..

jeremylenz:
@lfu Updated. `InsightsClientReportStatus` is no longer coupled to inventory syncs. Instead, I am now updating it in 2 places:

1. MachineTelemetriesController - I moved the status update to _before_ allowing or blocking the request (rather than after). This way if a host has the `host_registration_insights` param off, but still somehow has `insights-client` installed and attempting to report, the host will get the proper USER_OMITTED `InsightsClientReportStatus`.
2. For hosts that have the `host_registration_insights` param off from the start (which is 99% of cases), `insights-client` will never be installed at all, which means it will never hit `MachineTelemetriesController` and the status would be N/A forever. To avoid this, I also added a hook so we can refresh the `InsightsClientReportStatus` on each regular subscription-manager host checkin.

jeremylenz:
Updated to fix bugs:

1. hostgroup inheritance was not being respected for user-omitted count
2. "0" was getting rendered on screen incorrectly for user-omitted

lfu:
Got this after registered 4 hosts each in a hostgroup corresponding to the 4 combination of `host_registration_insights=true/false` with `host_registration_insights_inventory=true/false`.
And 2 hosts showed up in console dot.com.
inv22

Ran `Generate report and uplod` then `Sync all inventory status` and got this.
inv40

lfu:
Got this after changing param data type from `string` to `boolean` in hostgroup.
inv112

and after report:
inv202

lfu (approved):
LGTM 👍 
Thanks @jeremylenz

lfu:
Both host insights status and inventory status are aligned with the parameter setting.
Thanks @jeremylenz

lfu (approved):

LadislavVasina1 (approved):
ACK from my side, [automation](https://github.com/SatelliteQE/robottelo/pull/21165) for this test which exercises various host statuses passes.

PR #1183, opened 2026-03-23
Fix blank screen when switching to host without Insights tabs

nofaralfasi:
When using breadcrumb switcher to navigate from a registered host (with Recommendations/Vulnerabilities tabs) to an unregistered host, redirect to Overview tab instead of showing a loading spinner.

This PR fixes a navigation bug when using the breadcrumb host switcher to navigate from a registered host (with Recommendations/Vulnerabilities tabs) to an unregistered host (without those tabs). Previously, users would see a blank screen with a loading spinner. Now, they are automatically redirected to the Overview tab.

Changes:

- Added redirect logic to `CVEsHostDetailsTabWrapper` and `InsightsTab` components that detects when the current host doesn't support the tab and redirects to `/Overview`
- Uses the same visibility helper functions (`vulnerabilityDisabled`, `isNotRhelHost`, `hasNoInsightsFacet`) already used in the `hideTab` metadata
- Added corresponding test cases

Consistency: Reused the existing helper functions from `ForemanRhCloudHelpers.js` that are already used for the `hideTab` metadata, ensuring the redirect logic matches the tab visibility logic exactly

1. Setup: Have two types of hosts available:

-   Type 1: Unregistered host without Recommendations/Vulnerabilities tabs (e.g., the Satellite base host)
-   Type 2: Registered RHEL host with Recommendations & Vulnerabilities tabs (IoP environment)

2. Test the fix:

- Navigate to a Type 2 host and select the Recommendations tab
-  Use the breadcrumb switcher to switch to a Type 1 host
-   Expected: You are redirected to the Overview tab
-   Repeat for the Vulnerabilities tab

3. Regression test:

- Navigate to a Type 2 host, Recommendations or Vulnerabilities tab
- Use breadcrumb switcher to switch to another Type 2 host
- Expected: You remain on the same tab with the new host's data

## Summary by Sourcery

Ensure host details Insights and Vulnerabilities tabs redirect to Overview when unsupported for the selected host.

Bug Fixes:
- Fix blank screen by redirecting from unsupported Insights and Vulnerabilities tabs to the Overview tab when switching to hosts without the necessary capabilities.

Tests:
- Add tests validating redirect behavior for Insights and Vulnerabilities tabs based on host OS, Insights facet presence, and vulnerability enablement.

### DIFF
--- a/webpack/InsightsHostDetailsTab/NewHostDetailsTab.js
+++ b/webpack/InsightsHostDetailsTab/NewHostDetailsTab.js
@@ -1,6 +1,7 @@
 import React, { useEffect, useState } from 'react';
 import PropTypes from 'prop-types';
 import { useDispatch, useSelector } from 'react-redux';
+import { useHistory } from 'react-router-dom';
 import SearchBar from 'foremanReact/components/SearchBar';
 import { translate as __ } from 'foremanReact/common/I18n';
 import { ScalprumComponent, ScalprumProvider } from '@scalprum/react-core';
@@ -25,6 +26,7 @@ import { useIopConfig } from '../common/Hooks/ConfigHooks';
 import { generateRuleUrl } from '../InsightsCloudSync/InsightsCloudSync';
 import { createProviderOptions } from '../common/ScalprumModule/ScalprumContext';
 import { useInsightsPermissions } from '../common/Hooks/PermissionsHooks';
+import { isNotRhelHost, hasNoInsightsFacet } from '../ForemanRhCloudHelpers';
 
 // Hosted Insights advisor
 const NewHostDetailsTab = ({ hostName, router }) => {
@@ -165,7 +167,24 @@ const IopInsightsTabWrapped = props => {
 };
 
 const InsightsTab = props => {
+  const { response } = props;
+  const history = useHistory();
   const isIop = useIopConfig();
+  const isHostDataLoaded = Boolean(response?.id);
+  const shouldHideTab =
+    isHostDataLoaded &&
+    (isNotRhelHost({ hostDetails: response }) ||
+      hasNoInsightsFacet({ response, hostDetails: response }));
+
+  useEffect(() => {
+    if (shouldHideTab && history) {
+      history.replace('/Overview');
+    }
+  }, [shouldHideTab, history]);
+
+  if (shouldHideTab) {
+    return null;
+  }
 
   return isIop ? (
     <IopInsightsTabWrapped {...props} />
@@ -174,6 +193,12 @@ const InsightsTab = props => {
   );
 };
 
-InsightsTab.defaultProps = {};
+InsightsTab.propTypes = {
+  response: PropTypes.object,
+};
+
+InsightsTab.defaultProps = {
+  response: {},
+};
 
 export default InsightsTab;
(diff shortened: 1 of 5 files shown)


nofaralfasi:
@sourcery-ai review

webpack/InsightsHostDetailsTab/NewHostDetailsTab.js
```diff
@@ -174,6 +193,12 @@ const InsightsTab = props => {
   );
 };
 
-InsightsTab.defaultProps = {};
+InsightsTab.propTypes = {
+  response: PropTypes.object,
```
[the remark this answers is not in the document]
jeremylenz: I thought we had a lint rule that prohibits `PropTypes.object` in favor of `.shape`.. in any case, I too prefer `.shape`

webpack/InsightsHostDetailsTab/NewHostDetailsTab.js
```diff
@@ -165,7 +171,24 @@ const IopInsightsTabWrapped = props => {
...
+
+  useEffect(() => {
+    if (shouldHideTab && history) {
+      history.replace(OVERVIEW_TAB_PATH);
+    }
+  }, [shouldHideTab, history]);
+
+  if (shouldHideTab) {
+    return null;
+  }
```
lfu, line 184: Please extract the duplicated redirect logic into a single custom hook:
```
// Add this to the existing ForemanRhCloudHelpers.js file

import { useEffect } from 'react';
import { useHistory } from 'react-router-dom';

export const OVERVIEW_TAB_PATH = '/Overview';

/**
 * Redirects to Overview tab when the current tab should be hidden
 * @param {boolean} shouldRedirect - Whether to redirect (e.g., host loaded AND tab should hide)
 */
export const useTabRedirect = (shouldRedirect) => {
  const history = useHistory();
  
  useEffect(() => {
    if (shouldRedirect) {
      history.replace(OVERVIEW_TAB_PATH);
    }
  }, [shouldRedirect, history]);
  
  return shouldRedirect;
};
```
Then
```
import { useTabRedirect, isNotRhelHost, hasNoInsightsFacet } from '../ForemanRhCloudHelpers';

const NewHostDetailsTab = ({ response = {} }) => {
   const shouldHide = useTabRedirect(
    response?.id && (isNotRhelHost() || hasNoInsightsFacet())
  );
  
  if (shouldHide) return null;

  // ... rest of component stays the same
};
```

nofaralfasi: Implemented the `useTabRedirect` custom hook as suggested. That said, I'm not fully convinced it's necessary here - with only 2 components using this pattern, the ~5 lines of duplicated code might be more readable inline than adding a layer of abstraction. Happy to revert to the inline version if reviewers prefer explicit over DRY in this case.

webpack/InsightsHostDetailsTab/__tests__/NewHostDetailsTab.test.js
```diff
@@ -2,9 +2,20 @@ import React from 'react';
 import { render } from '@testing-library/react';
 import '@testing-library/jest-dom';
 import { Provider } from 'react-redux';
+import { MemoryRouter } from 'react-router-dom';
 import configureMockStore from 'redux-mock-store';
 import thunk from 'redux-thunk';
 import NewHostDetailsTab from '../NewHostDetailsTab';
+import { OVERVIEW_TAB_PATH } from '../../ForemanRhCloudHelpers';
+
+const mockHistoryReplace = jest.fn();
```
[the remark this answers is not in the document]
nofaralfasi: Already addressed - there's `jest.clearAllMocks()` in afterEach (line 99-101) which clears `mockHistoryReplace` after each test.

webpack/InsightsHostDetailsTab/NewHostDetailsTab.js
jeremylenz: I actually do like it; it makes it a bit more explicit what's happening vs. `history.replace(OVERVIEW_TAB_PATH);`

lfu (approved):
LGTM 👍

PR #1189, opened 2026-03-26
Fix report path not returned when subscription connection is disabled

chris1984:

* When subscription_connection_enabled is FALSE, the generated report stays in generated_reports/ folder since QueueForUploadJob is skipped. The report_file_paths method only searched uploads/ and done/ folders, so the UI received an empty array and couldn't show the download path.
* Add generated_reports/ to the search locations in report_file_paths and add tests covering all three folder locations.
* Fixed wording on the upload section

#### Root cause

* (lib/foreman_inventory_upload.rb:33-36):

* When subscription_connection_enabled is FALSE:
  1. The report generates successfully into generated_reports/ folder
  2. QueueForUploadJob is skipped (line 17 of generate_report_job.rb), so the file never moves to uploads/
  3. report_file_paths (line 33-36) only searches in uploads/ and done/ — not in generated_reports/
  4. The UI gets an empty array and can't show the report path

* Spin up a non iop Satellite 6.18.4 latest snap
* Turn the `subscription_connection_enabled` to false in Settings
* Try to do an inventory upload and see nothing shows up
Screenshot_26-3-2026_11217_ip-10-0-198-98 rhos-01 prod psi rdu2 redhat com
* `cd /usr/share/gems/gems/foreman_rh_cloud-12.2.17/`
* `wget https://patch-diff.githubusercontent.com/raw/theforeman/foreman_rh_cloud/pull/1189.patch
* `patch -p1 < 1189.patch`
* Ignore the errors about the test files and hit y it should have patched the main files
* Restart services with `foreman-maintain service restart`
* Try to do an inventory upload and see it shows the report location and text now:
Screenshot_26-3-2026_113646_ip-10-0-198-98 rhos-01 prod psi rdu2 redhat com

### DIFF
--- a/lib/foreman_inventory_upload.rb
+++ b/lib/foreman_inventory_upload.rb
@@ -32,7 +32,11 @@ module ForemanInventoryUpload
 
   def self.report_file_paths(organization_id)
     filename = facts_archive_name(organization_id)
-    Dir[ForemanInventoryUpload.uploads_file_path(filename), ForemanInventoryUpload.done_file_path(filename)]
+    Dir[
+      ForemanInventoryUpload.generated_reports_file_path(filename),
+      ForemanInventoryUpload.uploads_file_path(filename),
+      ForemanInventoryUpload.done_file_path(filename)
+    ]
   end
 
   def self.generated_reports_folder
@@ -42,6 +46,10 @@ module ForemanInventoryUpload
         'generated_reports/'
       )
     )
+  end
+
+  def self.generated_reports_file_path(filename)
+    File.join(ForemanInventoryUpload.generated_reports_folder, filename)
   end
 
   def self.outputs_folder
--- a/lib/foreman_inventory_upload.rb
+++ b/lib/foreman_inventory_upload.rb
@@ -32,6 +32,13 @@ module ForemanInventoryUpload
 
   def self.report_file_paths(organization_id)
     filename = facts_archive_name(organization_id)
+    # Report files start in generated
+    # They are then MOVED (not copied) to uploads, then done.
+    # When they are moved to the new folder, they overwrite any file with the same name.
+    # If it's a generate-only, it will be in generated
+    # Failed or incomplete uploads will be in uploads
+    # Completed uploads will be in done
+    # The ordering here ensures we get the correct file path every time.
     Dir[
       ForemanInventoryUpload.generated_reports_file_path(filename),
       ForemanInventoryUpload.uploads_file_path(filename),
--- a/lib/tasks/rh_cloud_inventory.rake
+++ b/lib/tasks/rh_cloud_inventory.rake
@@ -46,6 +46,10 @@ namespace :rh_cloud_inventory do
             filter,
             false # don't upload; the user ran report:generate and not report:generate_upload
           )
+          unless Setting[:subscription_connection_enabled]
+            report_path = File.join(base_folder, ForemanInventoryUpload.facts_archive_name(organization_id, filter))
+            puts "Generated report at #{report_path}" if File.exist?(report_path)
+          end
         end
         puts "Check the Uploading tab for report uploading status." if Setting[:subscription_connection_enabled]
       end
--- a/lib/tasks/rh_cloud_inventory.rake
+++ b/lib/tasks/rh_cloud_inventory.rake
@@ -47,8 +47,8 @@ namespace :rh_cloud_inventory do
             false # don't upload; the user ran report:generate and not report:generate_upload
           )
           unless Setting[:subscription_connection_enabled]
-            report_path = File.join(base_folder, ForemanInventoryUpload.facts_archive_name(organization_id, filter))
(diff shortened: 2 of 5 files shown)


lib/tasks/rh_cloud_inventory.rake
```diff
@@ -46,6 +46,10 @@ namespace :rh_cloud_inventory do
             filter,
             false # don't upload; the user ran report:generate and not report:generate_upload
           )
+          unless Setting[:subscription_connection_enabled]
+            report_path = File.join(base_folder, ForemanInventoryUpload.facts_archive_name(organization_id, filter))
+            puts "Generated report at #{report_path}" if File.exist?(report_path)
+          end
```
jeremylenz, line 52: The value of the setting isn't a guarantee of where the file will end up. I think this can be improved even more if we just tell you where the file is, no matter what. We don't really need to make you switch tabs to see the file location.

How about something like
```rb
report_paths = ForemanInventoryUpload.report_file_paths(organization_id)                                                                                                                   
  if report_paths.any?                                                                                                                                                                       
    puts "Report saved to file #{report_paths.first}"                                                                                                                                         
  end
```

lib/foreman_inventory_upload.rb
```diff
@@ -32,7 +32,11 @@ def self.done_file_path(filename)
 
   def self.report_file_paths(organization_id)
     filename = facts_archive_name(organization_id)
-    Dir[ForemanInventoryUpload.uploads_file_path(filename), ForemanInventoryUpload.done_file_path(filename)]
+    Dir[
+      ForemanInventoryUpload.generated_reports_file_path(filename),
+      ForemanInventoryUpload.uploads_file_path(filename),
+      ForemanInventoryUpload.done_file_path(filename)
+    ]
```
jeremylenz, line 46: I am biased, but I recommend adding some comments here like I did in my PR: https://github.com/jeremylenz/foreman_rh_cloud/blob/33452b62093fd18e70c884454aa9b8af906ffb15/lib/foreman_inventory_upload.rb#L48-L54

jeremylenz (commented):
Thanks @chris1984! a couple small comments below

test/controllers/accounts_controller_test.rb
```diff
@@ -5,6 +5,49 @@ class AccountsControllerTest < ActionController::TestCase
...
+  test 'report_file_paths finds report in uploads folder' do
+    test_org = FactoryBot.create(:organization)
+    filename = ForemanInventoryUpload.facts_archive_name(test_org.id)
+
+    uploads_path = ForemanInventoryUpload.uploads_file_path(filename)
+    FileUtils.mkdir_p(File.dirname(uploads_path))
+    FileUtils.touch(uploads_path)
+
+    paths = ForemanInventoryUpload.report_file_paths(test_org.id)
+    assert_includes paths, uploads_path
```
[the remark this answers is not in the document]
chris1984: Not needed since this is the flow:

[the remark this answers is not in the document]
chris1984: # Report files start in generated
    # They are then MOVED (not copied) to uploads, then done.
    # When they are moved to the new folder, they overwrite any file with the same name.
    # If it's a generate-only, it will be in generated
    # Failed or incomplete uploads will be in uploads
    # Completed uploads will be in done
    # The ordering here ensures we get the correct file path every time.

chris1984:
Updated, screenshot from disconnected:
Screenshot_27-3-2026_15299_ip-10-0-168-198 rhos-01 prod psi rdu2 redhat com

Screenshot connected:
Screenshot_27-3-2026_153049_ip-10-0-168-198 rhos-01 prod psi rdu2 redhat com

lib/foreman_inventory_upload.rb
chris1984: Added comments

lib/tasks/rh_cloud_inventory.rake
chris1984: Thanks for the suggestion, added the location regardless of setting

jeremylenz:
Per offline discussion - Having the path show on the Generated tab (for connected) isn't useful if that's not where the file ends up. Let's remove it and go back to printing the message based on the setting.

chris1984:
> Per offline discussion - Having the path show on the Generated tab (for connected) isn't useful if that's not where the file ends up. Let's remove it and go back to printing the message based on the setting.

Put back to how it was before

jeremylenz (approved):
Code LGTM, thanks @chris1984!

Did not test.

chris1984:
@ColeHiggins2 can you test this on your 6.18.4 sat? Will hold off on merging until I get an ack from you as well

ColeHiggins2 (commented):
Kicking off Packit build with existing automation, Will post results

ColeHiggins2:
@chris1984 Tested on 6.18.4 and looks good, Will update automation now

ColeHiggins2 (approved):

PR #1196, opened 2026-04-13
Pass through cloud error responses as-is

nofaralfasi:
The cloud request forwarder was re-formatting error responses (non-200 HTTP responses) by wrapping them in a custom JSON structure with message, error, headers, and response fields. This caused double-escaping when the upstream service returned JSON errors, making it difficult for the frontend to properly parse and display error messages.

Changed both MachineTelemetriesController and UIRequestsController to pass through error response bodies directly with their original status code and content-type, matching the behavior of successful responses.

The cloud request forwarder was re-formatting error responses (non-200 HTTP responses) by wrapping them in a custom JSON structure:
```
{
  "message": "Cloud request failed",
  "error": "{\"errors\": [...]}",
  "headers": {},
  "response": "{\"errors\": [...]}"
}
```
This caused double-escaping when the upstream service returned JSON errors, making it difficult for the frontend to properly parse and display error messages.

This PR changes both MachineTelemetriesController and UIRequestsController to pass through error response bodies directly with their original status code and content-type, matching the behavior of successful responses.

- Backward compatibility: This is a breaking change for any code that expects the wrapped error format. However, the wrapped format was problematic and inconsistent with successful responses, so this change aligns error handling with the existing success path.
- Special error cases preserved: The specific handlers for RestClient::Unauthorized (401) and RestClient::NotModified (304) are preserved with their custom messages, as these provide meaningful context (e.g., "Authentication to the Insights Service failed") rather than exposing internal authentication details.
- Content-type preservation: The fix extracts and forwards the original Content-Type header from the error response, defaulting to application/json if not present.
- Both controllers updated: The same forwarding logic exists in both MachineTelemetriesController (for insights-client requests) and UIRequestsController (for UI requests), so both were updated consistently.

Manual testing via UI:

1. Log into Foreman and select a specific organization
2. Navigate to a URL that triggers a cloud error, e.g.: `https://<foreman-host>/insights_cloud/api/vulnerability/v1/systems/00000000-0000-0000-0000-000000000000`
3. Verify the response is the raw cloud error (not wrapped in a message/error/response structure)
4. Verify the response format:
**Before**: `{"message":"Cloud request failed","error":"{\"errors\":[...]}","headers":{},"response":"{\"errors\":[...]}"}`
image

**After**: `{"errors":[{"detail":"inventory_id must exist...","status":"404"}]}`
Screenshot From 2026-04-13 15-23-15

### DIFF
--- a/app/controllers/insights_cloud/api/machine_telemetries_controller.rb
+++ b/app/controllers/insights_cloud/api/machine_telemetries_controller.rb
@@ -32,14 +32,10 @@ module InsightsCloud::Api
       rescue RestClient::ExceptionWithResponse => e
         response_obj = e.response.presence || e.exception
         code = response_obj.try(:code) || response_obj.try(:http_code) || 500
-        message = 'Cloud request failed'
+        upstream_content_type = response_obj.try(:headers)&.[](:content_type)
+        content_type = upstream_content_type&.match?(/json/) ? upstream_content_type : 'application/json'
 
-        return render json: {
-          :message => message,
-          :error => response_obj.to_s,
-          :headers => {},
-          :response => response_obj,
-        }, status: code
+        return render body: response_obj.to_s, status: code, content_type: content_type
       rescue StandardError => e
         # Catch any other exceptions here, such as Errno::ECONNREFUSED
         logger.warn("Cloud request failed with exception: #{e}")
--- a/app/controllers/insights_cloud/ui_requests_controller.rb
+++ b/app/controllers/insights_cloud/ui_requests_controller.rb
@@ -36,14 +36,10 @@ module InsightsCloud
       rescue RestClient::ExceptionWithResponse => e
         response_obj = e.response.presence || e.exception
         code = response_obj.try(:code) || response_obj.try(:http_code) || 500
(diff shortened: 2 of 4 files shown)


app/controllers/insights_cloud/api/machine_telemetries_controller.rb
```diff
@@ -32,14 +32,8 @@ def forward_request
...
-        message = 'Cloud request failed'
-
-        return render json: {
-          :message => message,
-          :error => response_obj.to_s,
-          :headers => {},
-          :response => response_obj,
-        }, status: code
+
+        return render body: response_obj.to_s, status: code, content_type: 'application/json'
```
jeremylenz: I feel like hard-coding the content type as application/json isn't ideal. Could check it first:

```rb
content_type = response_obj.try(:headers)&.[](:content_type) || 'application/json'                                                                                                         
  return render body: response_obj.to_s, status: code, content_type: content_type
```

jeremylenz (changes requested):

jeremylenz (approved):
Thanks @nofaralfasi 

Works well. ACK 👍

PR #1201, opened 2026-04-27
SAT-44713 - Move dev dependencies to Gemfile

ofedoren:

Move theforeman-rubocop and rdoc dependencies to Gemfile instead of specifying them in gemspec.

With newer bundler and the way we setup the Foreman with plugins, there are resolution failures due to duplicated dependencies.

Run bundle install with any other plugin that has the same dependency with a different version.

Before: fails to resolve (or there are warning on older bundler version)
After: passes without any issue, rubocop still works locally and in CI

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,3 +1,6 @@
 source 'https://rubygems.org'
 
 gemspec
+
+gem 'rdoc'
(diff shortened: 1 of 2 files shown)


adamruzicka (approved):
Ack, unless rh cloud maintainers want to prune rdoc too

chris1984 (approved):

PR #1200, opened 2026-04-24
Replace Enzyme snapshot tests with RTL in ForemanInventoryUpload

jeremylenz:
#### What are the changes introduced in this pull request?

Replace all Enzyme-based snapshot tests in `webpack/ForemanInventoryUpload/` with React Testing Library (RTL) behavioral tests, as part of the React 18 upgrade (SAT-16054). Enzyme does not support React 18.

- 23 test files rewritten with RTL `render` + behavioral assertions
- 9 trivial test files deleted (selectors, action creators, reducers with no real logic)
- 31 `.snap` files removed
- Net reduction of ~960 lines

#### Considerations taken when implementing this change?

- Trivial tests (simple property-access selectors, pass-through action creators, single-field reducers) were deleted rather than blindly converted — they added no value beyond what the component-level tests already cover.
- Tests with real logic (conditional selectors, multi-action reducers, string-checking helpers) were rewritten as direct unit tests with explicit assertions.
- Integration tests were converted from Enzyme's `IntegrationTestHelper` to RTL with `Provider` + `redux-mock-store` + `fireEvent`.
- All RTL tests follow existing patterns in the codebase (e.g., using `.toBeTruthy()` instead of `toBeInTheDocument()` since jest-dom matchers aren't configured).
- Components with JSDOM-incompatible dependencies (C3.js charts, PF4 context consumers) are mocked at the module level.

#### What are the testing steps for this pull request?

- [ ] Run JS tests from foreman directory: `cd /path/to/foreman && npx jest --no-coverage --config webpack/jest.config.js --roots /path/to/foreman_rh_cloud/webpack --testPathPattern="ForemanInventoryUpload"`
- [ ] Verify no `.snap` files remain: `find webpack/ForemanInventoryUpload -name "*.snap" | wc -l` (expected: 0)
- [ ] Verify no Enzyme/snapshot helper imports remain: `grep -r "testComponentSnapshotsWithFixtures\|testActionSnapshotWithFixtures\|testReducerSnapshotWithFixtures\|testSelectorsSnapshotWithFixtures\|IntegrationTestHelper" webpack/ForemanInventoryUpload/ --include="*.js"` (expected: no results)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1203, opened 2026-04-27
Replace Enzyme snapshot tests with RTL in InsightsCloudSync

jeremylenz:

Replace all 9 Enzyme-based snapshot test files in `InsightsCloudSync/` with React Testing Library (RTL) behavioral tests, as part of the React 18 migration (SAT-38754). Enzyme does not support React 18.

- **Deleted 3 trivial tests** (InsightsSettingsReducer, InsightsSettingsSelectors, InsightsCloudSyncHelpers) — these tested library behavior (seamless-immutable merge, property access, string concatenation), not application logic
- **Rewrote 3 action tests** as direct unit tests with explicit dispatch/mock assertions (InsightsSettingsActions, InsightsCloudSyncActions, InsightsTableActions)
- **Rewrote 1 selector test** with direct value assertions covering URI decoding, conditional returns, and state aggregation (InsightsTableSelectors)
- **Rewrote 2 component tests** as RTL behavioral tests covering mount side-effects, conditional IOP/non-IOP rendering, re-fetch on prop change (InsightsTable, InsightsCloudSync)

Result: 6 test suites, 38 tests, 0 snapshots. All snapshot files and `__snapshots__/` directories removed from InsightsCloudSync.

- Followed the same patterns established in the ForemanInventoryUpload conversion (SAT-38753)
- Trivial tests were deleted rather than converted to avoid testing library internals
- Component tests mock Redux-connected children (e.g., Pagination) to isolate parent logic
- Avoided the mock-testing-mock anti-pattern — assertions verify real parent behavior, not mock output

- [ ] Run InsightsCloudSync JS tests: `cd /home/vagrant/foreman && npx jest --no-coverage --config webpack/jest.config.js --roots /home/vagrant/foreman_rh_cloud/webpack --testPathPattern="InsightsCloudSync"`
- [ ] Verify all 38 tests pass across 6 suites with 0 snapshots
- [ ] Verify no `.snap` files remain: `find webpack/InsightsCloudSync -name "*.snap" | wc -l` (expect 0)
- [ ] Verify no Enzyme helpers remain: `grep -r "testComponentSnapshotsWithFixtures\|testActionSnapshotWithFixtures\|testReducerSnapshotWithFixtures\|testSelectorsSnapshotWithFixtures" webpack/InsightsCloudSync/ --include="*.js"` (expect no results)


jeremylenz:
For the PageLayout mock comment: it was already correctly destructuring from a props object (({ children, header, toolbarButtons }) =>), so no change was needed there.

webpack/InsightsCloudSync/Components/InsightsTable/__tests__/InsightsTableActions.test.js
```diff
@@ -8,41 +7,134 @@ import {
...
   clearAllSelection,
 } from '../InsightsTableActions';
+import {
+  INSIGHTS_SET_SELECTED_IDS,
+  INSIGHTS_SET_SELECT_ALL_ALERT,
+  INSIGHTS_SET_SELECT_ALL,
+  INSIGHTS_HITS_API_KEY,
+  INSIGHTS_HITS_PATH,
+} from '../InsightsTableConstants';
 import { hits } from './fixtures';
```
nofaralfasi: `hits` is imported but never used in the test file.

webpack/InsightsCloudSync/Components/InsightsTable/__tests__/InsightsTableSelectors.test.js
```diff
@@ -1,6 +1,5 @@
-import { testSelectorsSnapshotWithFixtures } from '@theforeman/test';
 import { insightsStateWrapper } from '../../../../ForemanRhCloudTestHelpers';
-import { routerState, APIState, APIErrorState } from './fixtures';
+import { routerState, APIState, APIErrorState, hits } from './fixtures';
```
nofaralfasi: `hits` is imported but never used directly - it's only used indirectly via APIState which already includes it internally.

webpack/InsightsCloudSync/__tests__/InsightsCloudSyncActions.test.js
```diff
@@ -1,9 +1,42 @@
...
-  testActionSnapshotWithFixtures(fixtures));
+jest.mock('../../common/ForemanTasks', () => ({
+  setupTaskPolling: jest.fn(opts => ({ type: 'SETUP_TASK_POLLING', ...opts })),
+  taskRelatedToast: jest.fn((id, type, msg) => ({
+    type: 'TASK_TOAST',
+    id,
+    toastType: type,
+    msg,
+  })),
+}));
```
nofaralfasi: These mocks for `setupTaskPolling` and `taskRelatedToast` return detailed objects, but `handleSuccess` is never actually invoked in the tests - so these mocks aren't exercised. Could either be simplified to just `jest.mock('../../common/ForemanTasks')` or expanded to actually test the callback behavior.

nofaralfasi (approved):
Just a few minor comments, otherwise LGTM.

jeremylenz:
Rebased & comments addressed.

webpack/InsightsCloudSync/InsightsCloudSync.test.js
```diff
@@ -1,25 +1,103 @@
...
+
+  describe('non-IOP mode', () => {
+    it('renders page layout with child components', () => {
+      render(<RecommendationsPage {...defaultProps} />);
+
+      expect(screen.getByTestId('page-layout')).toBeTruthy();
+      expect(screen.getByTestId('insights-table')).toBeTruthy();
+      expect(screen.getByTestId('insights-settings')).toBeTruthy();
+      expect(screen.getByTestId('remediation-modal')).toBeTruthy();
+      expect(screen.getByTestId('toolbar-dropdown')).toBeTruthy();
```
nofaralfasi: These assertions test the mocks, not real behavior. The data-testid values come from the mocks defined above (lines 14-26), so these will always pass.

webpack/InsightsCloudSync/InsightsCloudSync.test.js
```diff
@@ -1,25 +1,103 @@
...
+jest.mock('./Components/RemediationModal', () => () => (
+  <div data-testid="remediation-modal" />
+));
+jest.mock('./Components/ToolbarDropdown', () => () => (
+  <div data-testid="toolbar-dropdown" />
+));
+jest.mock('./Components/InsightsTable/Pagination', () => () => null);
+jest.mock('./Components/InsightsSettings', () => () => (
+  <div data-testid="insights-settings" />
+));
```
nofaralfasi: Are all these mocks needed? Which of these components are tested elsewhere?
ToolbarDropdown - is this a simple presentational component that could render without mocking?
RemediationModal - same question
Pagination - same question
If they have no Redux dependencies, consider removing the mocks and rendering the real components.

webpack/InsightsCloudSync/InsightsCloudSync.test.js
```diff
@@ -1,25 +1,103 @@
...
+  describe('IOP mode', () => {
+    beforeEach(() => {
+      mockIsIop = true;
+    });
+
+    it('renders Scalprum component instead of page layout', () => {
+      render(<RecommendationsPage {...defaultProps} />);
+
+      expect(screen.getByTestId('scalprum-component')).toBeTruthy();
+      expect(screen.queryByTestId('page-layout')).toBeNull();
```
nofaralfasi: The IOP mode branching test is valuable (tests real conditional logic), but still relies on mock data-testid. Could simplify to just checking that one renders and the other doesn't, without needing data-testid on the mocks.

nofaralfasi (changes requested):
Following the pattern from PR #1200 review - consider whether this test file is needed at all. If `InsightsCloudSync` is primarily composition glue and all children are mocked, the test provides limited value. The IOP branching logic is the main thing worth testing here.

jeremylenz:
@nofaralfasi Updated:

Addressed in e3cf8aa:                                                                                                                                                                       
                       
  - Removed assertions on mock data-testid values — these were testing the mocks, not real behavior                                                                                           
  - Simplified mocks to () => null since we no longer assert on their output                                                                                                                  
  - Removed the ToolbarDropdown mock — it's not Redux-connected and renders fine without it (Pagination, RemediationModal, InsightsTable, and InsightsSettings all need mocks because they're 
  Redux-connected via connect() or hooks)                                                                                                                                                     
  - Rewrote the IOP branching test to assert on real CSS classes (.advisor vs .rh-cloud-insights) instead of mock data-testid                                                                 
  - Consolidated from 5 tests to 3 — the file now focuses on the IOP branching logic, which is the real value here

nofaralfasi (approved):
LGTM

PR #1215, opened 2026-06-12
Use ref parameter for katello dependency to support Ruby 3.3

sjha4:
use git ref instead of branch to bundle ruby. This ref should work with both master branch and PR merge refs which Katello CI workflow relies on to trigger rh_cloud tests.
AI has me convinced that: 
```
Bundler 2.5+ (shipped with Ruby 3.3) validates branch: parameter
  more strictly and rejects non-branch refs like SHAs or PR merge refs.
  The ref: parameter accepts any git reference including branches,
  making it compatible with both normal use and CI environments.
```
Green CIs here against master branch..And green CI on https://github.com/Katello/katello/pull/11778 after this is merged and the github action there is rerun with https://github.com/Katello/katello/pull/11778/changes#diff-7aa10e7d6cc9e32f26a92b7d4be84100a3247fa948f51cedff811996309b048eL26 .

### DIFF
--- a/gemfile.d/katello.rb
+++ b/gemfile.d/katello.rb
@@ -1,1 +1,1 @@
-gem 'katello', github: 'Katello/katello', branch: ENV.fetch('KATELLO_BRANCH', 'master')
+gem 'katello', github: 'Katello/katello', ref: ENV.fetch('KATELLO_REF', 'master')


sjha4:
I see the gemfile for ruby 3.3 run pointed correctly to katello master via git: https://github.com/theforeman/foreman_rh_cloud/actions/runs/27435006819/artifacts/7600281189 and so did ruby 3.0 tests. Both resolved to katello (5.0.0.pre.master)

parthaa (approved):
APJ

PR #1224, opened 2026-06-26
SAT-46497 - Disable insights_recommendations_count search in IoP mode

jeremylenz:
## Summary
- In IoP mode, recommendation counts are fetched live from the IoP Smart Proxy — they are not stored locally in `insights_facets.hits_count`. This made the `insights_recommendations_count` scoped search return incorrect results.
- Changed the scoped search from a simple `:relation` to an `ext_method` that raises `ScopedSearch::QueryNotSupported` with a clear message in IoP mode, following the same pattern as the `insights_uuid` fix.
- Updated existing test in `insights_facet_test.rb` to stub IoP mode, and added new tests for both IoP and non-IoP behavior.

## Test plan
- [ ] Verify `insights_recommendations_count = 1` search works correctly in non-IoP mode
- [ ] Verify `insights_recommendations_count = 1` search raises a clear error message in IoP mode
- [ ] Verify the Recommendations column still renders correctly on the hosts page in both modes
- [ ] Run `bundle exec rake test:foreman_rh_cloud`

### DIFF
--- a/app/models/concerns/rh_cloud_host.rb
+++ b/app/models/concerns/rh_cloud_host.rb
@@ -46,8 +46,6 @@ module RhCloudHost
     end
   end
 
-  NUMERIC_OPERATORS = ['=', '!=', '>', '>=', '<', '<='].freeze
-
   module ClassMethods
     def search_by_insights_recommendations_count(_key, operator, value)
       if ForemanRhCloud.with_iop_smart_proxy?
@@ -56,7 +54,7 @@ module RhCloudHost
         )
       end
 
-      unless RhCloudHost::NUMERIC_OPERATORS.include?(operator)
+      unless ScopedSearch::QueryBuilder::SQL_OPERATORS.value?(operator)
         raise ScopedSearch::QueryNotSupported.new(
           _('Unsupported operator for recommendations count search: %s') % operator
         )
--- a/app/models/concerns/rh_cloud_host.rb
+++ b/app/models/concerns/rh_cloud_host.rb
@@ -61,9 +61,19 @@ module RhCloudHost
       end
 
       facet_table = InsightsFacet.table_name
-      condition = sanitize_sql_for_conditions(
-        ["COALESCE(#{facet_table}.hits_count, 0) #{operator} ?", value_to_sql(operator, value)]
-      )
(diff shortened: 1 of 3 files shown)


jeremylenz:
Re the overall comment about using association-based joins: the ext_method interface for scoped_search expects raw SQL in the {joins:, conditions:} hash, and this is consistent with
  search_by_insights_uuid right below it. Changing just this method would be inconsistent.

app/models/concerns/rh_cloud_host.rb
```diff
@@ -46,6 +47,30 @@ def ensure_iop_insights_uuid
...
+      unless ScopedSearch::QueryBuilder::SQL_OPERATORS.value?(operator)
+        raise ScopedSearch::QueryNotSupported.new(
+          _('Unsupported operator for recommendations count search: %s') % operator
+        )
+      end
+
+      facet_table = InsightsFacet.table_name
+      condition = sanitize_sql_for_conditions(
+        ["COALESCE(#{facet_table}.hits_count, 0) #{operator} ?", value_to_sql(operator, value)]
+      )
```
nofaralfasi: One concern here: `search_by_insights_recommendations_count` accepts operators from `ScopedSearch::QueryBuilder::SQL_OPERATORS`, which includes `IN` / `NOT IN`, but currently always builds: `COALESCE(...hits_count, 0) #{operator} ?`
For `^` /` !^` searches (`IN` / `NOT IN`), this needs list placeholders `(IN (?, ?, ...))` or explicit rejection, otherwise behavior can be incorrect/fail for set operators.
Could we either:
handle `IN` / `NOT IN` the same way `as search_by_insights_uuid`, or
raise `ScopedSearch::QueryNotSupported` for those operators?
It would also be great to add a test for `insights_recommendations_count ^ 1,2` (or the expected rejection path) to lock behavior.

jeremylenz:
@nofaralfasi updated :+1:

nofaralfasi (approved):
LGTM.
I tested this in both IoP and non-IoP environments, and behavior works as expected in each mode.
Thanks for following up on the `^`/`!^` operator concern - the `IN`/`NOT IN` handling in `search_by_insights_recommendations_count` and the added `^` test address the main risk I saw in earlier revisions.

PR #1228, opened 2026-06-30
Fix ensure unfinishable Upload job is marked done

### LINKED ISSUE #1227: UploadReportDirectJob-Action no longer exits poll-loop, if no CandlePin Certs
#1127 modified the Upload-Action and dropped setting the Action to `done` in case it cannot finish successfully. It now simply returns which means it will retry after a growing amount of time.

https://github.com/theforeman/foreman_rh_cloud/blob/0247ca86b3b65fae2af6c0c61caf78bd815c9e92/lib/foreman_inventory_upload/async/upload_report_direct_job.rb#L59-L68

As I understand there is a daily task `Wait and Generate all reports job`, which will now stay in running/suspended until the timeout is hit. That is if one of the above exit conditions apply.

@jeremylenz is that expected behavior?

m-bucher:
Fixes #1227

Add code to mark aborted Jobs as `done` (broke with #1127).

Created unit-tests (might need some refinement).

Have a system not registered to Insights (no consumer certs) and start Upload-Job.
_Expected Result:_ Upload aborted, but Task finishes.

### DIFF
--- a/lib/foreman_inventory_upload/async/upload_report_direct_job.rb
+++ b/lib/foreman_inventory_upload/async/upload_report_direct_job.rb
@@ -59,11 +59,13 @@ module ForemanInventoryUpload
       def try_execute
         if content_disconnected?
           logger.info("Upload canceled: connection to Insights is not enabled. Report location: #{filename}")
+          done!
           return
         end
 
         unless organization.owner_details&.dig('upstreamConsumer', 'idCert')
           logger.info("Skipping organization '#{organization}', no candlepin certificate defined.")
+          done!
           return
         end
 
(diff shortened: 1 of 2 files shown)


m-bucher:
@jeremylenz ready-for-review

jeremylenz (approved):
Tested and works as expected.

In the log I see
```
[I|bac|25a4bddc] Skipping organization 'org2', no candlepin certificate defined.
```
and the task completes with success. Tested with both IoP and non-IoP setups, each with 2 orgs (1 manifest and 1 non-manifest org).

Thanks @m-bucher!

PR #1230, opened 2026-07-05
Fix inventory upload job hang on non-retryable errors

nofaralfasi:
When report upload returns non-retryable client errors (4xx except 408/429), mark UploadReportDirectJob as done instead of re-raising and polling forever. This prevents HostInventoryReportJob from staying stuck around 83% and keeps the inventory upload UI from remaining in a perpetual loading/disabled state. Also adds tests for non-retryable vs retryable client error behavior.

- Updated UploadReportDirectJob to treat non-retryable upload client errors as terminal.
- On HTTP 4xx (except 408 and 429), the job now logs, sets `output[:status] = 'upload_skipped_non_retryable'`, and finishes with `done!` instead of retrying.
- Added tests covering

- Preserved retry behavior for transient conditions (408 timeout and 429 rate limit).
- Scoped the change to upload error handling only; no changes to report generation flow.
- Added explicit task output status for observability.

- Run unit tests for upload job
- Verify new cases:
  - 422 => action completes with upload_skipped_non_retryable
  - 429 => action raises/retries as before.

### DIFF
--- a/lib/foreman_inventory_upload/async/upload_report_direct_job.rb
+++ b/lib/foreman_inventory_upload/async/upload_report_direct_job.rb
@@ -78,6 +78,18 @@ module ForemanInventoryUpload
 
         move_to_done_folder
         done!
+      rescue RestClient::ExceptionWithResponse => e
+        if non_retryable_upload_error?(e)
+          logger.warn(
+            "Upload skipped for organization '#{organization}' " \
+            "due to non-retryable response (#{e.http_code}): #{sanitized_http_body(e)}"
+          )
+          output[:status] = 'upload_skipped_non_retryable'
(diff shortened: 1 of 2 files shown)


jeremylenz (commented):
This seems the same as https://github.com/theforeman/foreman_rh_cloud/pull/1228 which was just merged?

nofaralfasi:
> This seems the same as #1228 which was just merged?

Good catch, you’re right. After re-testing and reviewing the logs, #1228 already fixes SAT-42246 (the stuck-at-83% case) by marking the early-abort upload paths as done, so the task no longer hangs. I’ll close #1230 as redundant to this fix.

PR #1205, opened 2026-05-01
SAT-25889 - Handle missing Foreman base host records gracefully

jeremylenz:
## Summary
- Handle cases where the Foreman/Satellite host itself doesn't have a host record in the database
- Add fallback to extract hostname from `Setting[:foreman_url]` when host record is missing
- Add nil guards throughout codebase to prevent crashes when foreman_host is nil
- Skip sync tasks gracefully when foreman_host is unavailable

## Changes
- Modified `ForemanRhCloud.foreman_host` to return nil instead of raising when host not found
- Added `ForemanRhCloud.foreman_url_hostname` to extract hostname from foreman_url setting as fallback
- Updated `marked_foreman_host` to use facet lookup instead of scoped search
- Added nil guards in:
  - `ForemanInventoryUpload::Generators::FactHelpers` (bios_uuid, host_ips, hostname_match)
  - `InsightsCloud::Async::InsightsGenerateNotifications` (skips when host missing)
  - `InventorySync::Async::InventorySelfHostSync` (skips when host missing)
- Added comprehensive tests for nil host scenarios

## Test plan
- [x] All existing tests pass
- [x] New tests verify behavior when foreman_host is nil
- [x] Metadata generation works without host record
- [x] Sync tasks skip gracefully instead of crashing

🤖 Generated with [Claude Code](https://claude.ai/claude-code)

PR #1231, opened 2026-07-13
Release 13.2.10

Odilhao:

This is a release PR for `13.2.10`, cherry-picking three fixes from `develop` onto `foreman_3_18`:

- **#1183 - Fix blank screen when switching to host without Insights tabs**: When using the breadcrumb host switcher to navigate from a registered host (with Recommendations/Vulnerabilities tabs) to an unregistered host, the UI now redirects to the Overview tab instead of showing a blank screen with an infinite loading spinner.
- **#1196 - Pass through cloud error responses as-is (SAT-47604)**: `MachineTelemetriesController` and `UIRequestsController` no longer re-wrap non-200 cloud error responses in a custom JSON structure. Error bodies are now passed through with their original status code and content-type, the same way successful responses are handled, avoiding double-escaping of upstream JSON errors.
- **#1205 - Handle missing Foreman base host records gracefully (SAT-25889 / SAT-47605)**: `foreman_host` now returns `nil` instead of raising `ActiveRecord::RecordNotFound` when the base Foreman host record doesn't exist, and `foreman_host_name` uses `Setting[:foreman_url]` instead of `SmartProxy.default_capsule.name`. Call sites (metadata generation, inventory sync, insights notifications, host helper methods) were updated with nil guards so inventory upload, inventory sync, and recommendations sync keep working in `foremanctl` setups where the base host record isn't created. This fixes the `"can't find record with friendly id: 'satellite.example.com'"` error.

- All three commits were cherry-picked with `-x` to preserve a link back to the original commit on `develop`.
- Verified the exact commit SHAs against the source PRs before picking, since #1196 and #1205 were squash-merged into `develop` as a single commit each — the SHAs cherry-picked here are the actual squash-merge commits, not the original per-commit SHAs from the source branches.
- #1183 and #1196 applied cleanly; #1205 auto-merged three files (`lib/foreman_inventory_upload.rb`, `lib/foreman_inventory_upload/generators/fact_helpers.rb`, `test/jobs/inventory_self_host_sync_test.rb`) with no manual conflict resolution needed.

1. **Blank screen fix (#1183)**
   - Navigate to a registered RHEL host and open the Recommendations or Vulnerabilities tab.
   - Use the breadcrumb switcher to switch to an unregistered host (no Insights facet).
   - Expected: redirected to the Overview tab instead of a blank/loading screen.

2. **Cloud error pass-through (#1196)**
   - Trigger a request through the cloud request forwarder (e.g. insights-client request or telemetry call) that results in a non-200 response from the cloud service.
   - Expected: the error response body, status code, and content-type are returned as-is to the client, without being wrapped/double-escaped.

3. **Missing base host record (#1205)**
   - In a `foremanctl`/IoP-style setup where the Foreman base host record is not created for a given hostname, run inventory upload, inventory sync, and recommendations sync.
   - Expected: these tasks complete without raising `ActiveRecord::RecordNotFound` / "can't find record with friendly id" errors, and skip host-dependent logic gracefully when the host is `nil`.

### DIFF
--- a/app/controllers/insights_cloud/api/machine_telemetries_controller.rb
+++ b/app/controllers/insights_cloud/api/machine_telemetries_controller.rb
@@ -32,14 +32,10 @@ module InsightsCloud::Api
       rescue RestClient::ExceptionWithResponse => e
         response_obj = e.response.presence || e.exception
         code = response_obj.try(:code) || response_obj.try(:http_code) || 500
-        message = 'Cloud request failed'
+        upstream_content_type = response_obj.try(:headers)&.[](:content_type)
+        content_type = upstream_content_type&.match?(/json/) ? upstream_content_type : 'application/json'
 
-        return render json: {
-          :message => message,
-          :error => response_obj.to_s,
-          :headers => {},
-          :response => response_obj,
-        }, status: code
+        return render body: response_obj.to_s, status: code, content_type: content_type
       rescue StandardError => e
         # Catch any other exceptions here, such as Errno::ECONNREFUSED
         logger.warn("Cloud request failed with exception: #{e}")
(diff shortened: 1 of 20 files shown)


chris1984 (approved):
LGTM tests are failing though

Odilhao:
You know that I don't have merge permissions, right?

PR #1232, opened 2026-07-14
fix(iop): add missing remed modal to pathway props

Fewwy:
Add missing remediations modal props so the pathway details systems table won't crash

Test it in iop environment using https://github.com/RedHatInsights/insights-advisor-frontend/tree/foreman-pathways-dev-v2 branch and iop-dev repository instructions

### DIFF
--- a/webpack/IopRecommendationDetails/IopRecommendationDetails.js
+++ b/webpack/IopRecommendationDetails/IopRecommendationDetails.js
@@ -25,6 +25,7 @@ const IopRecommendationDetails = props => {
           scope={scope}
           module={pathwayModule}
           pathwayId={pathwayMatch.params.slug}
+          IopRemediationModal={RemediationModal}
           {...props}
         />
       </div>


webpack/IopRecommendationDetails/IopRecommendationDetails.js
```diff
@@ -25,6 +25,7 @@ const IopRecommendationDetails = props => {
           scope={scope}
           module={pathwayModule}
           pathwayId={pathwayMatch.params.slug}
+          IopRemediationModal={RemediationModal}
```
jeremylenz, line 28: Does the Pathways component actually use the remediation modal?

Fewwy: We have it in c.r.c 

Image

jeremylenz: Right, I know c.r.c has its own remediation modal, I was just curious why we need to pass in the IoP one. Can you plan a remediation from Pathways and it works the same as advisor? I guess I'm not familiar with Pathways..

Fewwy: You should be able to do remediation for systems in the pathways. They are still the same systems just grouped together by a pathway.

jeremylenz (approved):
Lint errors are unrelated (caused by https://github.com/theforeman/foreman/pull/11008)

PR #1233, opened 2026-07-15
Fix no-magic-numbers lint errors

jeremylenz:
## Summary
- theforeman/foreman#11008 added the `no-magic-numbers` ESLint rule to Foreman core's shared plugin lint config (used by `npm run lint:plugins` in CI, via `script/lint/@theforeman/eslint-plugin-foreman/lint_generic_config.js`)
- This broke the `Lint foreman_rh_cloud` CI check with 8 new errors across 5 files
- Extracted the flagged numeric literals into named constants (polling/timeout intervals, seconds-per-minute conversions, percent-complete value) so the plugin lints cleanly again

## Test plan
- [x] Reproduced the CI failure by checking out Foreman `develop` and running `node script/npm_lint_plugins.js foreman_rh_cloud`
- [x] Confirmed the same command exits 0 after this change
- [x] `npm test` — all 191 existing tests still pass (3 unrelated pre-existing suite failures due to missing `foremanReact` mock modules in the local environment, unaffected by this change)


jeremylenz:
May or may not wait for https://github.com/theforeman/foreman/pull/11098 and https://github.com/theforeman/foreman/pull/10968 which will break more things.

chris1984 (approved):
Looks good to me, thanks for fixing this, I am guessing we should backport this to the 3_18 branch since that just failed with the same error with Odilions release PR.

jeremylenz:
> we should backport this to the 3_18 branch since that just failed with the same error with Odilions release PR.

I don't think this should affect that branch. The 3_18 branch needs to be updated to test with the correct Foreman branch - it's in the github workflows somewhere.

chris1984:
> > we should backport this to the 3_18 branch since that just failed with the same error with Odilions release PR.
> 
> I don't think this should affect that branch. The 3_18 branch needs to be updated to test with the correct Foreman branch - it's in the github workflows somewhere.

Sounds good, oversight on my side probably when I branched. Will make a follow up pr to correct that

PR #1235, opened 2026-07-16
SAT-32064 - update InsightsHostDetailsTab from pf3 to pf5

andreilakatos:

Updated InsightsHostDetailsTab from PF3 to PF5

Before:
Screenshot 2026-07-16 at 11 14 08
Screenshot 2026-07-16 at 11 14 33

After:
Screenshot 2026-07-16 at 11 14 20
Screenshot 2026-07-16 at 11 14 41

### DIFF
--- a/webpack/InsightsHostDetailsTab/components/ListItem/ListItem.js
+++ b/webpack/InsightsHostDetailsTab/components/ListItem/ListItem.js
@@ -1,33 +1,24 @@
-import React, { Fragment } from 'react';
-import { ListView, Icon } from 'patternfly-react';
+import React, { useState } from 'react';
+import {
+  AccordionContent,
+  AccordionItem,
+  AccordionToggle,
+  Flex,
+  FlexItem,
+  Truncate,
+} from '@patternfly/react-core';
+import { ExternalLinkAltIcon } from '@patternfly/react-icons';
 import PropTypes from 'prop-types';
 import { translate as __ } from 'foremanReact/common/I18n';
-
-const labelMapper = {
-  1: __('Low'),
-  2: __('Moderate'),
-  3: __('Important'),
-  4: __('Critical'),
-};
+import InsightsLabel from '../../../InsightsCloudSync/Components/InsightsTable/InsightsLabel';
 
 const ListItem = ({ title, totalRisk, resultsUrl, solutionUrl }) => {
-  const heading = (
-    <p className="ellipsis list-item-heading" title={title}>
-      {title}
-    </p>
-  );
-
-  const riskLabel = labelMapper[totalRisk];
-  const additionalInfo = [
-    <span key={`risk-info-${title}`} className={`risk-label risk-${totalRisk}`}>
-      <p>{riskLabel}</p>
-    </span>,
-  ];
+  const [isExpanded, setIsExpanded] = useState(false);
 
   const knowledgebaseLink = solutionUrl && (
     <p>
       <a href={solutionUrl} target="_blank" rel="noopener noreferrer">
-        {__('Knowledgebase article')} <Icon name="external-link" />
+        {__('Knowledgebase article')} <ExternalLinkAltIcon />
       </a>
     </p>
   );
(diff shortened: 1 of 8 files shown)


webpack/InsightsHostDetailsTab/components/ListItem/__tests__/ListItem.test.js
```diff
@@ -0,0 +1,100 @@
...
+      screen.queryByRole('link', {
+        name: /Read more about it in RH cloud insights/i,
+      })
+    ).not.toBeInTheDocument();
+  });
+
+  it('shows detail links after expanding the recommendation', async () => {
+    renderListItem();
+
+    userEvent.click(await screen.findByRole('button'));
```
nofaralfasi: `userEvent.click` should be awaited in async tests.

In this test file, click interactions are followed immediately by assertions. With React 18 and `@testing-library/user-event`, these interactions are async and not awaiting them can cause intermittent failures.

Please update these to `await userEvent.click(...)` (here and the other occurrences in this file) before asserting the post-click UI state.

webpack/InsightsHostDetailsTab/components/ListItem/__tests__/ListItem.test.js
```diff
@@ -0,0 +1,100 @@
...
+
+    expect(
```
nofaralfasi: Can we also assert the external-link behavior (`target="_blank"`), not just `href`?

In prior PF3→PF5 migrations we’ve had regressions around link rendering/behavior, so it would be good to verify these links still open in a new tab after migration.

For example, in this test:
- `Knowledgebase article` link: assert `target="_blank"` (and optionally `rel="noopener noreferrer"`)
- `Read more about it in RH cloud insights` link: assert `target="_blank"` (and optionally `rel="noopener noreferrer"`)

webpack/InsightsHostDetailsTab/components/ListItem/ListItem.js
```diff
@@ -1,58 +1,66 @@
...
+        <Flex
+          alignItems={{ default: 'alignItemsCenter' }}
+          justifyContent={{ default: 'justifyContentSpaceBetween' }}
+          flexWrap={{ default: 'nowrap' }}
+        >
+          <FlexItem>
+            <Truncate content={title} />
+          </FlexItem>
+          <FlexItem>
+            <InsightsLabel value={totalRisk} />
```
nofaralfasi, line 50: This may introduce a translation regression for risk labels.

Before migration, this component used localized labels via `__('Low'/'Moderate'/'Important'/'Critical')`.  
Now it renders `InsightsLabel`, which currently uses hardcoded English text.

Can we preserve localization by passing translated text explicitly (e.g. `text={__(...)}`), or update `InsightsLabel` itself to be i18n-aware?

andreilakatos:
Thanks @nofaralfasi, addressed the comments 🙏

nofaralfasi:
During validation, I found that the PF3 -> PF5 UI migration alone is not sufficient: the Legacy UI Insights tab still has a functional issue and may show “No recommendations were found for this host” even when recommendations exist.

Because of this underlying behavior, I’m seeing results that differ from what is currently shown/reported in this PR.

I’m attaching screenshots from my environment to show the discrepancy (after addressing the “No recommendations” issue):

Before applying this PR:
Screenshot From 2026-07-29 17-51-43

After applying this PR:
Screenshot From 2026-07-29 18-01-41

I suggest we first fix Legacy Insights tab behavior in a separate PR, and then continue validation of the PF3 -> PF5 changes here, so we can assess the migration on top of correct baseline functionality.

I’ll open a PR with the Legacy tab fix by EOT.
**Update**: the PR with the fix https://github.com/theforeman/foreman_rh_cloud/pull/1240.

andreilakatos:
> During validation, I found that the PF3 -> PF5 UI migration alone is not sufficient: the Legacy UI Insights tab still has a functional issue and may show “No recommendations were found for this host” even when recommendations exist.
> 
> Because of this underlying behavior, I’m seeing results that differ from what is currently shown/reported in this PR.
> 
> I’m attaching screenshots from my environment to show the discrepancy (after addressing the “No recommendations” issue):
> 
> Before applying this PR: Screenshot From 2026-07-29 17-51-43
> 
> After applying this PR: Screenshot From 2026-07-29 18-01-41
> 
> I suggest we first fix Legacy Insights tab behavior in a separate PR, and then continue validation of the PF3 -> PF5 changes here, so we can assess the migration on top of correct baseline functionality.
> 
> I’ll open a PR with the Legacy tab fix by EOT. **Update**: the PR with the fix #1240.

This could be fixed with a sanity check on my end. I'll do that

nofaralfasi (approved):
LGTM for the PF3 -> PF5 migration scope in this PR.\n\nI still want us to continue investigating the behavior tracked in #1240 separately, but that should not block this UI migration PR for now. Approving #1235 as-is.

nofaralfasi (approved):
LGTM for the PF3 -> PF5 migration scope in this PR.
I still want us to continue investigating the behavior tracked in #1240 separately, but that should not block this UI migration PR for now. Approving #1235 as-is.

PR #1237, opened 2026-07-23
fix(iop): separate route for pathway details

Fewwy:
Pathway details and recommendation details were rendered by a single component (IopRecommendationDetails) on one catch-all route, using useRouteMatch internally to switch between views. 
When <Link> inside the federated PathwayDetailsWrapped module navigated to a recommendation details URL, the route still matched the same catch-all, so the host router never re-mounted the component and useRouteMatch didn't re-evaluate — leaving the user stuck on the Pathway details page despite the URL changing. 

Splitting into separate routes with a dedicated IopPathwayDetails component ensures the router treats pathway-to-recommendation navigation as a proper route change, unmounting one view and mounting the other.

How to repeat the bug:
Run the dev environment, open pathway details page, click on Link to the Recommendation details in Recommendations table - you will see URL is changed but you are still on the same page and not navigated to the Recommendation details page


adonispuente (approved):
works as intended, lgtm!

jeremylenz (commented):
LGTM, please fix lint and rebase :)

jeremylenz (approved):

PR #1239, opened 2026-07-27
Fix prefer-pf-components lint error in HelpLabel to fix JS tests

chris1984:
Replace raw <button> with pf-v5-c-form CSS class with PatternFly <Button variant="plain"> component to satisfy the
@theforeman/rules/prefer-pf-components ESLint rule.

Tested `inventory_upload` page
Tested Recommendations page

pop over text icon looks the same and works correctly.

### DIFF
--- a/webpack/common/Switcher/HelpLabel.js
+++ b/webpack/common/Switcher/HelpLabel.js
@@ -1,18 +1,19 @@
 import React from 'react';
 import PropTypes from 'prop-types';
-import { Popover } from '@patternfly/react-core';
+import { Button, Popover } from '@patternfly/react-core';
 import { HelpIcon } from '@patternfly/react-icons';
 
 export const HelpLabel = ({ text, id, className }) => {
   if (!text) return null;
   return (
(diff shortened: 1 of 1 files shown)


nofaralfasi (approved):
LGTM. Tested this change on the `inventory_upload` and `Recommendations` pages, and it works as expected.
I also verified usage is limited to `SwitcherPF4`, which already passes `switcher-help-label`, so dropping `pf-v5-c-form__group-label-help` is safe here.
Optional follow-up: `onClick={e => e.preventDefault()}` may be unnecessary on a plain button unless there’s a specific reason to keep it.

chris1984:
> LGTM. Tested this change on the `inventory_upload` and `Recommendations` pages, and it works as expected. I also verified usage is limited to `SwitcherPF4`, which already passes `switcher-help-label`, so dropping `pf-v5-c-form__group-label-help` is safe here. Optional follow-up: `onClick={e => e.preventDefault()}` may be unnecessary on a plain button unless there’s a specific reason to keep it.

Removed the `onClick={e => e.preventDefault()}`

PR #1243, opened 2026-08-11
fix(compliance): add SCOPED_REQUESTS entries for test_results

romanblanco:
Without explicit entries, test_results endpoints fall through to the catch-all (`api/compliance/v2/.*`) which has no `tag_name`. Foreman won't append `?tags=` to forwarded requests, so test results are returned unscoped — any workspace can see any test result regardless of RBAC tag assignments.

Add list and show entries with `tag_name: :tags` before the catch-all, matching the existing pattern for report systems.


ColeHiggins2 (approved):
Looks good to me

nofaralfasi (approved):
LGTM.
Non-blocking suggestion: add `forward_request` tests for the two new compliance `test_results` endpoints to explicitly assert `:tags` is present in forwarded params (similar to existing vulnerability forwarding tests).

chris1984 (approved):

LadislavVasina1 (approved):

romanblanco:
@ColeHiggins2 can we get this merged?

chris1984:
> @ColeHiggins2 can we get this merged?

Will get this merged today

